Issue #42, opened 2024-02-02
First pass at a developer guide


ianthetechie:
Currently we have some documentation spread across the README, examples, and the ARCHITECTURE document. As things mature over the next few months, we should consolidate this into an mdBook and publish it on GitHub pages to make onboarding new devs easier.

PR #61, opened 2024-02-28
Feat/common hashable types

Archdoog:


apple/Tests/FerrostarCoreTests/FerrostarCoreModelTests.swift
ianthetechie, line 1: Guessing this will get filled out later?

apple/Sources/FerrostarCore/Extensions/CoreLocation Extensions.swift
```diff
@@ -0,0 +1,110 @@
...
+}
+
+extension CourseOverGround {
+    
+    /// Intialize a Course Over Ground object from the relevant Core Location types. These can be found on a
+    /// CLLocation object.
+    ///
+    /// - Parameters:
+    ///   - course: The direction the device is travelling, measured in degrees relative to true north.
+    ///   - courseAccuracy: The accuracy of the direction
```
ianthetechie, line 66: Should probably document that it will return nil if the input course is invalid.

apple/Sources/FerrostarCore/Extensions/CoreLocation Extensions.swift
```diff
@@ -0,0 +1,110 @@
...
+}
+
+extension Heading {
+    
+    /// Initialize a Heading if it can be represented as integers
+    ///
+    /// This will return nil for invalid headings.
+    ///
+    /// - Parameter clHeading: The CoreLocation heading provided by the location manager.
+    public init?(clHeading: CLHeading) {
```
ianthetechie, line 84: Same re: docs

apple/Sources/FerrostarCore/NavigationState.swift
```diff
@@ -19,22 +19,50 @@ public struct NavigationState {
...
-        self.heading = heading
-        self.courseOverGround = snappedLocation.course
-        self.fullRouteShape = fullRoute
+        self.snappedLocation = UserLocation(clLocation: snappedLocation)
+        if let heading {
+            self.heading = Heading(clHeading: heading)
+        }
+        self.courseOverGround = CourseOverGround(course: snappedLocation.course,
+                                                 courseAccuracy: snappedLocation.courseAccuracy)
+        self.fullRouteShape = fullRoute.map { GeographicCoordinate(lat: $0.latitude, lng: $0.longitude) }
```
ianthetechie, line 28: I think you may have added a convenience init in the extensions that cleans this up slightly.

apple/Sources/FerrostarMapLibreUI/MapLibre Extensions.swift
```diff
@@ -4,11 +4,11 @@ import MapLibre
 
 extension NavigationState {
     var routePolyline: MLNPolyline {
-        return MLNPolylineFeature(coordinates: fullRouteShape)
+        return MLNPolylineFeature(coordinates: fullRouteShape.map { CLLocationCoordinate2D(latitude: $0.lat, longitude: $0.lng) })
```
ianthetechie, line 7: I think the convenience init extensions might clean this up?

apple/Sources/FerrostarMapLibreUI/MapLibre Extensions.swift
```diff
@@ -4,11 +4,11 @@ import MapLibre
...
     }
 
     var remainingRoutePolyline: MLNPolyline {
         // FIXME
-        return MLNPolylineFeature(coordinates: fullRouteShape)
+        return MLNPolylineFeature(coordinates: fullRouteShape.map { CLLocationCoordinate2D(latitude: $0.lat, longitude: $0.lng) })
```
ianthetechie, line 12: I think the convenience init extensions might clean this up?

apple/Tests/FerrostarCoreTests/Models/CoreLocationModelTests.swift
```diff
@@ -0,0 +1,103 @@
...
+        
+        XCTAssertEqual(userLocation.coordinates.lat, -77.846323)
+        XCTAssertEqual(userLocation.coordinates.lng, 166.668235)
+        XCTAssertEqual(userLocation.horizontalAccuracy, 1.1)
+        XCTAssertEqual(userLocation.courseOverGround?.degrees, 45)
+        XCTAssertEqual(userLocation.courseOverGround?.accuracy, 3)
+        XCTAssertEqual(userLocation.timestamp, timestamp)
+    }
+    
+    // TODO: Decide methodology for invalid CLLocation data (negatives)
```
ianthetechie, line 65: Good question. Off the top of my head, I'd say the following are invalid:

* Lat > 90
* Lat < -90
* Lng > 180
* Lng < 180
* Negative heading
* Negative course
* Negative accuracy (on any model that includes accuracy)

ianthetechie (commented):
Great work! Mostly minor nits. Thanks for reorganizing and running swiftfmt or w/e as well on some sections 👍

common/README.md
```diff
@@ -13,3 +13,22 @@ and routing backends, which are described in detail in [ARCHITECTURE ](../ARCHIT
...
+```
+
+### iOS 
+
+Run the build iOS shell command to build the xcframework required to locally run the Swift Package and demo app.
+
+```sh
+cd common
+./build-ios.sh
+```
```
ianthetechie, line 34: Could you integrate this into `CONTRIBUTING.md` instead? That's the closest thing we have to a dev guide before #42; the README should evolve toward being a quickstart for end users (and a pointer to the fuller docs which we can organize better with MDBook with sections for developers wanting to hack on the SDK and not just use it).

common/ferrostar/src/models.rs
```diff
@@ -59,6 +59,42 @@ impl From<Rect> for BoundingBox {
...
 }
 
+/// The heading of the user/device.
+/// 
+/// Ferrostar prefers course over ground, but may use heading in some cases.
+#[derive(Clone, Copy, PartialEq, PartialOrd, Debug, uniffi::Record)]
+pub struct Heading {
+    /// The heading in degrees relative to true north.
+    pub true_heading: u16,
+    /// The maximum deviation in degrees between the reported heading and the true geomagnetic heading.
```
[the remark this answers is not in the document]
Archdoog: Heading is something we'd need to manually calculate from Sensor data sdk(s) on android as far as I can tell. So I think for now this holds true. We could revise it if and when we integrate an android "device" heading.

apple/Tests/FerrostarCoreTests/FerrostarCoreModelTests.swift
Archdoog: Deleted for now. It will be, but we can defer it for now.

common/ferrostar/src/models.rs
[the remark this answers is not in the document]
ianthetechie: After looking at it a bit, Android does expose it via the raw sensor APIs. I would suggest not making any guarantees in the doctrines about its accuracy though. It will be highly hardware specific at any rate.

apple/Tests/FerrostarCoreTests/Models/CoreLocationModelTests.swift
Archdoog: Let chat about this item. After having played around with CLLocation to see what it'll do I'm less certain we want to front load this in a convenience init. E.g. we might find a slightly invalid location is better for navigation than just discarding it in full. I'm also curious how bad a system provided location can get.

ianthetechie: I'm not quite following what you mean by a "slightly invalid location." A location is either valid or invalid. There isn't a middle ground.

Whether a CLLocation is useful at a given _accuracy_ is tunable by the user already at multiple levels of Ferrostar. However, any CLLocation with a negative accuracy is explicitly invalid and coordinates out of the range cannot be reliably interpreted so they should be discarded (we could enforce this in Rust; these will never be valid).

CLLocation reference on the accuracy field (we should discard all values with negative accuracy): https://developer.apple.com/documentation/corelocation/cllocation/1423599-horizontalaccuracy

apple/Sources/FerrostarCore/Extensions/CoreLocation Extensions.swift
```diff
@@ -0,0 +1,147 @@
...
+        // This behavior matches how CLLocation initializes with a coordinate (setting accuracy to 0 & date to now)
+        self.init(coordinates: GeographicCoordinate(cl: clCoordinateLocation2D),
+                  horizontalAccuracy: 0,
+                  courseOverGround: nil,
+                  timestamp: Date())
+    }
+    
+    /// Initialize a UserLocation from an Apple CoreLocation CLLocation
+    ///
+    /// Unlike CourseOverGround & Heading, this value will init with invalid values.
```
ianthetechie, line 135:
```suggestion
    /// Unlike CourseOverGround & Heading, this initializer will accept CLLocations that have invalid values.
```

Something like this... was a bit confusing on a first read ;)

ianthetechie (commented):
Just one minor thing :) This is great!

ianthetechie:
This is good enough that I'm just going to merge now and we'll go over the open discussion on Monday.

PR #63, opened 2024-03-03
Move from locations to waypoints; enforce code formatting; code cleanup

### LINKED ISSUE #4: Figure out a long-term solution for the Swift type re-export problem
In Swift it is not possible to simply mark a type for re-export. This is a fairly deep seated design decision, and there are several consequences. Downstream code depending on the core Swift package cannot construct or have an explicitly typed reference to any types from dependencies, including our own FFI(!). The types are "there" and you can store one in a non-explicitly-typed variable, pass this back to another function, and even inspect any of the properties, methods, etc. with full code completion in Xcode. But you cannot construct one yourself, nor can you store it in a typed field.

This is annoying for things like `NavigationStateUpdate`, which we currently basically re-export with our own type that (as an enum that is constructible only from the original enum) will generally cause compilation errors in the case of a divergence from the underlying type.

This is extremely problematic with types like `Route` in the context of enabling local route generation from user code, since we currently lack a good solution for constructing the `Route` in Swift code.

### LINKED ISSUE #43: Support custom route generation independent of HTTP
Ex: offline routing, having your own custom fallback layers, etc.

### LINKED ISSUE #62: Switch from coordinates to rich waypoints
Currently we use only coordinates for the waypoint data structure, and this hampers our ability to do any sort of "via" routes or customize leg splitting.

ianthetechie:
* Closes #4 
* Closes #43 
* Closes #62
* Adds swiftformat, ktfmt, and rustfmt checks (most of the rest of #2)

There are a few minor tasks outstanding (which should be marked as TODO), but this is ready for review.


CONTRIBUTING.md
```diff
@@ -47,7 +47,7 @@ For minor or straightforward bug fixes, feel free to proceed directly to a PR.
 2. Open the cargo workspace (`common/`) in your preferred editing environment. 
 
 The Rust project is a cargo workspace,
-and nothing beyond the above is requried should be needed to start hacking. 
+and nothing beyond the above is requried should be needed to start hacking.
```
ianthetechie, line 50: Add `cargo fmt` note

android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
```diff
@@ -18,254 +20,217 @@ import uniffi.ferrostar.RouteDeviation
...
-        stopNavigation()
-
-        val controller = NavigationController(
+      }
+
+  /**
+   * Starts navigation with the given parameters (erasing any previous state).
+   *
+   * Once you have a location fix and a desired route, invoke this method. It will automatically
+   * subscribe to location provider updates. Returns a view model which
```
ianthetechie, line 157: Finish this comment ;)

common/ferrostar/src/models.rs
```diff
@@ -43,6 +44,36 @@ impl From<GeographicCoordinate> for Point {
...
 
+/// A waypoint along a route.
+///
+/// Within the context of Ferrostar, a route request consists of exactly one [UserLocation]
+/// and at least one [Waypoint]. The route starts from the user's location (which may
+/// contain other useful information like their current course for the [crate::routing_adapters::RouteRequestGenerator]
+/// to use) and proceeds through one or more waypoints.
+///
+/// Waypoints are used during route calculation, are tracked throughout the lifecycle of a trip,
+/// and are used for recalculating when the sure deviates from the expected route.
```
ianthetechie, line 55:
```suggestion
/// and are used for recalculating when the user deviates from the expected route.
```

common/ferrostar/src/models.rs
```diff
@@ -43,6 +44,36 @@ impl From<GeographicCoordinate> for Point {
...
+///
+/// Note that support for features beyond basic geographic coordinates varies by routing engine.
```
ianthetechie, line 57:
```suggestion
/// Note that support for properties beyond basic geographic coordinates varies by routing engine.
```

ianthetechie (commented):
One more minor note to change the types in Location.swift

PR #73, opened 2024-03-26
Misc simulation fixes

ianthetechie:
* Fix camera jank due to saturating cast of negative headings
* Interpolation so that simulation can be smoother (ensuring a point exists every X meters; defaults to 10 in public-facing code)


.github/workflows/gradle-publish.yml
```diff
@@ -4,10 +4,6 @@
 name: Gradle Package
 
 on:
-  # TODO: Remove push trigger
-  push:
-    branches:
-      - main
```
ianthetechie, line 10: Cleanup from my initial pass at packaging; forgot that I had not pushed this directly to master.

.github/workflows/gradle-publish.yml
```diff
@@ -40,10 +36,6 @@ jobs:
       run: touch local.properties
       working-directory: android
 
-    - name: Execute Gradle build
-      run: ./gradlew build
-      working-directory: android
-
```
ianthetechie, line 46: This is not actually required; tests will build what they need ;)

common/ferrostar/src/simulation.rs
```diff
@@ -14,22 +15,20 @@ pub enum SimulationError {
...
-#[derive(uniffi::Record, Clone)]
+#[derive(uniffi::Record, Clone, PartialEq)]
 #[cfg_attr(test, derive(Serialize))]
 pub struct LocationSimulationState {
     pub current_location: UserLocation,
     remaining_locations: Vec<GeographicCoordinate>,
 }
 
-#[derive(uniffi::Enum)]
-pub enum SimulationAdvanceStyle {
```
ianthetechie, line 25: `SimulationAdvanceStyle` was dropped because I determined that it wasn't logical to specify this at every iteration but rather to decide upfront whether to interpolate.

common/ferrostar/src/simulation.rs
```diff
@@ -125,33 +158,108 @@ pub fn advance_location_simulation(
...
-    fn advance_to_next_location() {
-        let mut state = location_simulation_from_coordinates(vec![
-            GeographicCoordinate { lng: 0.0, lat: 0.0 },
-            GeographicCoordinate { lng: 1.0, lat: 1.0 },
-            GeographicCoordinate { lng: 2.0, lat: 2.0 },
-            GeographicCoordinate { lng: 3.0, lat: 3.0 },
-        ])
+    #[rstest]
+    #[case(None)]
+    #[case(Some(10.0))]
```
ianthetechie, line 167: This lets us parameterize tests with case variables. You can read this like "for each case, pass these parameters into the respective case annotated arguments."

common/ferrostar/src/simulation.rs
```diff
@@ -125,33 +158,108 @@ pub fn advance_location_simulation(
...
+            let new_state = advance_location_simulation(&state);
+            if new_state == state {
+                break;
+            }
+            state = new_state;
             states.push(state.clone());
         }
 
-        insta::assert_yaml_snapshot!(states);
+        insta::assert_yaml_snapshot!(format!("{:?}", resample_distance), states);
```
ianthetechie, line 200: This funky first argument gives the snapshot a deterministic name based on the args passed. Otherwise it's just a sequential value like 1, 2, etc., but the rstest crate ordering is not really deterministic in which order it executes cases ;)

ianthetechie (commented):
Some comments to speed up the review. Nothing really deep; most of the changes are actually snapshots which you don't really need to read as most of the key properties are asserted at various points.

Issue #25, opened 2023-11-19
[Android] Localized distance formatting


ianthetechie:
Android doesn't appear to have a built-in distance formatter analogous to `MKDistanceFormatter` on iOS.

Known options
* `DecimalFormat` - not deprecated, but Google basically says you shouldn't use this
* [`NumberFormatter`](https://developer.android.com/reference/android/icu/number/NumberFormatter) - Google says to use this for localized number formatting.

Unfortunately, `NumberFormatter` requires API level 30, which would seem to exclude a large number of users. It also only solves half the problem. We also need to determine which units we should be using, and this is not trivial (ex: UK prefers miles for distances), and the class doesn't appear to help with those.

Requirements off the top of my head:

- [x] Able to specify units somewhat granularly; we need to be able to have an imperial/metric split for the UK, for example, as they usually prefer miles for long distances but metric for everything else.
- [x] "Smart" rounding (ex: when below 100m, round to the nearest 10m; for large distances, probably round to the nearest mile or kilometer; switch from miles to fractional miles maybe, and then to feet, etc.).

Suggestions welcome on high-quality / de-facto standard Android libraries that can accomplish this.

Ali7040:
@ianthetechie Hi there, my name is Ali Haider. I am a frontend developer with skills in React, Angular, and TypeScript. I have contributed to an open-source project called OpenBeta and am currently contributing to an Angular project. I also have some experience with backend development using Node. I am interested in contributing to your project.

ianthetechie:
Thanks for your comment @Ali7040. Always happy to have more contributors :)

This issue is related to localized distance formatting on Android in Kotlin. Did you have any particular ideas related to this issue?

If you're looking to start a more general conversation about how you can contribute, I'd suggest having a look over the project README and ARCHITECTURE documents first and then either open up a [discussion](https://github.com/stadiamaps/ferrostar/discussions) (for open-ended questions like this) or else start discussing on a relevant issue thread.

ianthetechie:
Seems like this will be a mess with branches depending on APIs available and relying on developers to make a decision for their use case. We can evolve this over time but the first pass will probably be either be pretty "raw" or just using some tables from Wikipedia or something to guess at preferred units, combined with manual implementation of unit selection based on size and hard coded cutoffs where we switch from feet to miles, how we round, etc.

PR #76, opened 2024-04-01
Android distance formatter

ianthetechie:
This goes a long way toward #25 (just need to integrate it into banner views using the extensible design we worked out over the past week on iOS).

The main thing to discuss on the call is whether an API level change to 30 is acceptable.


ianthetechie:
Discussed on the call that we're actually going the other direction on API level support and dropping it to 25. That's now pretty well tested + documented so I'm gonna just merge this now. Not a lot to review. I'll get back to the revised banner implementation (in line with iOS) later this week.

PR #77, opened 2024-04-02
Feat/navigation view orientation

Archdoog:
- Adds portrait navigation view and outer orienting navigation view.
- Separates navigation view & navigation map view.
- Adds content inset system for navigation view.

### DIFF
--- a/apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
+++ b/apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
@@ -6,7 +6,6 @@ import MapLibreSwiftDSL
 import MapLibreSwiftUI
 import SwiftUI
 
-// TODO: Rename this NavigationView? Basically to detach the
 public struct NavigationMapView: View {
     @Environment(\.colorScheme) var colorScheme
 
--- a/apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
+++ b/apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
@@ -1,0 +1,87 @@
+import FerrostarCore
+import FerrostarSwiftUI
+import MapKit
+import MapLibre
+import MapLibreSwiftDSL
+import MapLibreSwiftUI
+import SwiftUI
+
+// TODO: Rename this NavigationView? Basically to detach the 
+public struct NavigationMapView: View {
+    @Environment(\.colorScheme) var colorScheme
+
+    let lightStyleURL: URL
+    let darkStyleURL: URL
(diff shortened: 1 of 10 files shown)


apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
```diff
@@ -6,15 +6,15 @@ import MapLibreSwiftDSL
 import MapLibreSwiftUI
 import SwiftUI
 
-// TODO: Rename this NavigationView? Basically to detach the 
+// TODO: Rename this NavigationView? Basically to detach the
```
ianthetechie, line 9: 😂

apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
```diff
@@ -24,13 +27,13 @@ public struct NavigationMapView: View {
         darkStyleURL: URL,
         navigationState: NavigationState?,
         camera: Binding<MapViewCamera>,
-        distanceFormatter: Formatter = MKDistanceFormatter()
+        @MapViewContentBuilder _ makeMapContent: () -> [StyleLayerDefinition] = { [] }
```
ianthetechie, line 30: What happened to the distance formatter? We definitely want to keep that user- configurable.

apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
```diff
@@ -60,55 +63,25 @@ public struct NavigationMapView: View {
...
-        lightStyleURL: URL(string: "https://demotiles.maplibre.org/style.json")!,
-        darkStyleURL: URL(string: "https://demotiles.maplibre.org/style.json")!,
-        navigationState: state,
-        camera: .constant(.center(state.snappedLocation.clLocation.coordinate, zoom: 12)),
-        distanceFormatter: formatter
-    )
-}
-
-#Preview("Navigation Map View (Metric)") {
+#Preview("Navigation Map View") {
```
ianthetechie, line 77: Similarly, I'd like to keep this as a demo of using multiple locales.

ianthetechie (commented):
Sorry GitHub only showed me a partial diff or something at first.

apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
Archdoog: This is only a "MapView" now, so there's nothing to format here. It's all pushed up one later to the "Navigation Views" where the instruction view is included.

ianthetechie (approved):

PR #80, opened 2024-04-05
Add BorderedPolyline composable

haysmike:
This adds a bordered polyline composable and uses it in the default `NavigationMapView` to create a nicer route polyline.

image


ianthetechie:
Nice! I'm pretty open to merging stuff like this quickly and iterating (I'm about to do something similar for banners on Android as well).

Looks like CI flagged a formatting issue. Just run ` ./gradlew ktfmtFormat` and push the changes.

ianthetechie (approved):

PR #83, opened 2024-04-07
Fix dark mode background color in iOS InstructionsView

CatMe0w:
In dark mode, the background of the InstructionsView remains white, making the text invisible.
screenshot

This adds `backgroundColor` to `InstructionRowTheme` and uses it in `InstructionsView` to render instruction cards in dark mode correctly.


ianthetechie:
Hey, thanks for the PR! This is great! I also noticed this but hadn't gotten around to fixing it yet 😂 @Archdoog and I will take a look at this tomorrow.

PR #95, opened 2024-04-18
Improved Camera Control on NavigationMapView

Archdoog:
- Adds concept to control camera between navigation snapped & user controlled.
- Corrects view issue with hidden pill.


apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
```diff
@@ -22,17 +22,55 @@ public struct NavigationMapView: View {
...
+    // MARK: Camera Settings
+
     @Binding private var camera: MapViewCamera
 
+    /// The snapped camera zoom. This is used to override the camera zoom whenever snapping is active.
+    @Binding private var snappedZoom: Double
+
+    /// Whether to snap the camera on the next navigation status update. When this is false,
+    /// the user can browse the map freely.
+    @Binding private var useSnappedCamera: Bool
```
ianthetechie, line 35: A few comments on this section.

1. Visibility: Unlike `State`, bindings are not private. The whole point is to be able to set them externally, so they can be public, even though the most common method of doing so is via a constructor parameter. (Aside: this also makes code cleaner in the common case since most views following this pattern can use auto-generated initializers; we only need a custom initializer because of how our builder works.)
2. No action needed on this PR, but I could see issues arising with two-way bindings and conflicting parts of the code updating the camera. Something to keep thinking about.

apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
```diff
@@ -22,17 +22,55 @@ public struct NavigationMapView: View {
...
+    ///   - snappingBreakawayVelocity: The drag gesture velocity used to disable snapping. This can be tuned to prevent
+    /// accidental drags.
+    ///   - content: Any additional MapLibre symbols to show on the map.
     public init(
         styleURL: URL,
         navigationState: NavigationState?,
         camera: Binding<MapViewCamera>,
+        snappedZoom: Binding<Double> = .constant(18),
+        useSnappedCamera: Binding<Bool> = .constant(true),
+        snappingBreakawayVelocity: CGFloat = 25,
```
ianthetechie, line 65: This view is the most fundamental of 3 layers of hierarchy of nav views, but most users will interact directly with `DynamicallyOrientingNavigationView`.

1. We need to bubble up all of these new bindings and tunables :)
2. I'd suggest *not* having defaults on these more fundamental views. This does two things:
  a. Prevents us from making mistakes / forgetting to pass things down through the higher level layers.
  b. Makes it clear to end users that they are explicitly opting for the more fundamental views.

So these defaults should (IMO) move all the way up to `DynamicallyOrientingNavigationView`, and then the compiler will tell you what you missed in the plumbing if these don't have defaults.

ianthetechie (commented):
Really nice set of improvements! Just a few comments on composability mostly.

apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift
Archdoog: Yep, it's intentionally a binding so that you can re-snap the camera (e.g. with a center on user button in a parent view/scene). In my navigation view, I'm basically using this value to hide/show my center on user button, then on tapping the button I set it back to true. Seems to work pretty well.

ianthetechie (approved):

PR #99, opened 2024-04-30
starting point for arrival state

Archdoog:

PR #105, opened 2024-05-07
Feat/trip progress

Archdoog:
Pushed directly to this repo so we can collaborate and wrap up the original: https://github.com/stadiamaps/ferrostar/pull/99


ianthetechie (approved):
Looks good after a few more minor changes + finishing out the type changes on Android. I think we can merge this now + you can get started on UI components next.

PR #107, opened 2024-05-08
A bit of linting

nyurik:
* unpublished crates should not have version, and should have `publish=false`
* add clippy lints to workspace
* worked through a number of clippy suggestions


ianthetechie:
Looks like we need to bump the MSRV (seems like 1.75 added support for version-less packages): https://github.com/rust-lang/cargo/pull/12786

ianthetechie (approved):

ianthetechie:
Just an FYI for next time: even changes to comments in the generated bindings result in UniFFI's pedantic integrity checksums failing ;) Normally one would simply not commit generated bindings, but regrettably I haven't yet figured out a way to avoid this given the... quirks of how SPM works. Android doesn't have this problem.

So, TL;DR, you'll need to run `build-ios.sh` in the `common` directory before committing and stage any changes in ferrostar.swift.

PR #110, opened 2024-05-24
feat: added basic arrival view

Archdoog:


apple/Tests/FerrostarSwiftUITests/Theme/ArrivalViewThemeTests.swift
```diff
@@ -0,0 +1,15 @@
+import XCTest
+@testable import FerrostarSwiftUI
+
+final class ArrivalViewThemeTests: XCTestCase {
```
ianthetechie, line 4: I doubt we need this test case TBH; it'll just be one more thing to change if we change our mind on the default constructor args 😂

apple/Tests/FerrostarSwiftUITests/Views/ArrivalViewTests.swift
```diff
@@ -0,0 +1,86 @@
...
+            ArrivalView(
+                progress: TripProgress(
+                    distanceToNextManeuver: 5420,
+                    distanceRemaining: 1_420_000,
+                    durationRemaining: 520_800
+                ),
+                distanceFormatter: germanDistanceFormatter,
+                estimatedArrivalFormatter: etaFormatter,
+                durationFormatter: longDurationFormatter,
+                fromDate: referenceDate
```
ianthetechie, line 82: I see that we only partially test localization here. Could we add a German ETA and duration formatter as well? That will help verify the changes I suggested in the defaults.

apple/Tests/FerrostarSwiftUITests/Theme/ArrivalViewThemeTests.swift
Archdoog: Agree this test is generally useless. The only two reasons I added it anyways are:

1. It's a good additional catch for misconfigurations that could lead to somewhat larger issues (e.g. missing text)
2. It matches what we did for the instructions view. 

We could also just snapshot test the full default ferrostar theme once we've got all the ui built.

ianthetechie: > It matches what we did for the instructions view.

😂 I guess feel free to delete that then; I must have missed it.

I'd say visual snapshot image tests of the UI are probably the way to go in issue detection here. Unless we have a specific edge case that we're trying to guard against. And that's probably solvable with other techniques that make illegal state unrepresentable.

Archdoog: It's gone! 😆

apple/Tests/FerrostarSwiftUITests/Views/ArrivalViewTests.swift
Archdoog: I revised our testing to include formatter tests as well as some locales on the views.

apple/Sources/FerrostarSwiftUI/Library/ArrivalFormatters.swift
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /// The formatter for duration on the arrival view.
+    public static var durationFormat: DateComponentsFormatter {
+        let formatter = DateComponentsFormatter()
+        formatter.allowedUnits = [.hour, .minute]
+        formatter.unitsStyle = .abbreviated
+        formatter.zeroFormattingBehavior = .dropAll
+        return formatter
+    }
```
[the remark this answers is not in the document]
Archdoog: While I very much like `Date.RelativeFormatStyle`, I think keeping it concise is important generally. This could be especially true when considering potential translations.

apple/Sources/FerrostarSwiftUI/Theme/ArrivalViewTheme.swift
```diff
@@ -0,0 +1,42 @@
...
+public struct DefaultArrivalViewTheme: ArrivalViewTheme {
+    public var style: ArrivalViewStyle = .full
+    public var measurementColor: Color = .primary
+    public var measurementFont: Font = .title2.bold()
+    public var secondaryColor: Color = .secondary
+    public var secondaryFont: Font = .subheadline
+    public var backgroundColor: Color = .init(.systemBackground)
+
+    public init() {
+        //
```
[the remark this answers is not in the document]
Archdoog: Fixed 👍

apple/Sources/FerrostarSwiftUI/Resources/Localizable.xcstrings
ianthetechie, line 1: If we are going to include translations, I'd prefer they be checked by a fluent speaker, so let's drop German for now (the thing I wanted to test wrt localization for this PR was just number and measurement formatting).

Archdoog: Removed 👍

ianthetechie (approved):

PR #111, opened 2024-05-30
Fix breaking out from navigation camera on iOS

Max-Leopold:
Two small changes to fix breaking out from the navigation camera on iOS:

1. Use a binding for the snapped instead of a constant to allow user to break from navigation camera.

It could be that this was done intentional to not allow breaking out from the navigation at all when using the `DemoNavigationView`. If so, feel free to ignore. The behavior looks quite buggy right now though because even if the user tries to move the map during navigation it will try to snap back to the user location.

2. Drag gesture velocity can be negative. So we should check the absolute value of the gesture velocity against the `breakwayVelocity`.

### DIFF
--- a/apple/DemoApp/Demo/DemoNavigationView.swift
+++ b/apple/DemoApp/Demo/DemoNavigationView.swift
@@ -32,6 +32,7 @@ struct DemoNavigationView: View {
     }
 
     @State private var camera: MapViewCamera = .center(initialLocation.coordinate, zoom: 14)
+    @State private var snappedCamera = true
 
     init() {
         let simulated = SimulatedLocationProvider(location: initialLocation)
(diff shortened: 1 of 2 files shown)


ianthetechie:
Hey @Max-Leopold; sorry it's taken a bit to get to this. Just getting back from some travel.

> It could be that this was done intentional to not allow breaking out from the navigation at all when using the DemoNavigationView. If so, feel free to ignore. The behavior looks quite buggy right now though because even if the user tries to move the map during navigation it will try to snap back to the user location.

Haha yeah, this was just a first hackish pass for a demo app. I'll try it out and give feedback but I'm guessing this is fine.

> Drag gesture velocity can be negative. So we should check the absolute value of the gesture velocity against the `breakwayVelocity`.

heh ;) good catch.

Unrelated: the CI error isn't your fault. There's a fix pending merge in another branch. I'll take care of that.

ianthetechie (approved):

PR #119, opened 2024-06-12
Feat/ios speed limit view

Archdoog:


apple/Sources/FerrostarMapLibreUI/Views/PortraitNavigationView.swift
```diff
@@ -7,63 +7,108 @@ import MapLibreSwiftUI
...
+        @ViewBuilder topTrailing: @escaping () -> TopTrailing = { InfiniteSpacer() },
+        @ViewBuilder midLeading: @escaping () -> MidLeading = { InfiniteSpacer() },
+        @ViewBuilder bottomTrailing: @escaping () -> BottomTrailing = { InfiniteSpacer() }
     ) {
+        self.theme = theme
         self.styleURL = styleURL
         self.navigationState = navigationState
-        self.distanceFormatter = distanceFormatter
-        userLayers = makeMapContent()
+        userLayers = mapContent()
```
Archdoog, line 47: I'd kinda like to make this consistent. @ianthetechie did you happen to see any particular discussion if the way I did the @ViewBuilders is less ideal that what we you had here with the non-escaping capture in init?

apple/Sources/FerrostarSwiftUI/Views/Controls/FerrostarButton.swift
```diff
@@ -0,0 +1,53 @@
+import SwiftUI
+
+public struct FerrostarButtonStyle: ButtonStyle {
```
ianthetechie, line 3: I think`StyledActionButton` might be a more descriptive type name here. There isn't really anything Ferrostar-specific; it's just got our `FerrostarButtonStyle` applied, a label, etc.

apple/Sources/FerrostarSwiftUI/Views/Controls/FerrostarButton.swift
```diff
@@ -0,0 +1,53 @@
...
+public struct FerrostarButton<Label: View>: View {
+    let action: () -> Void
+    @ViewBuilder var label: () -> Label
+
+    /// The basic Ferrostar SwiftUI button style.
+    ///
+    /// - Parameters:
+    ///   - action: The action the button performs on tap.
+    ///   - label: The label subview.
+    public init(action: @escaping () -> Void, label: @escaping () -> Label) {
```
ianthetechie, line 26: Carrying on the same theme from the other discussion, I think this should be non-escaping and maybe a ViewBuilder? I wish I could point to the profound reasoning why I am convinced of this, but I'm pretty sure actions should be escaping but not views.

apple/Sources/FerrostarSwiftUI/Views/GridViews/InnerGridView.swift
```diff
@@ -0,0 +1,179 @@
...
+    public init(
+        @ViewBuilder topLeading: @escaping () -> TopLeading = { InfiniteSpacer() },
+        @ViewBuilder topCenter: @escaping () -> TopCenter = { InfiniteSpacer() },
+        @ViewBuilder topTrailing: @escaping () -> TopTrailing = { InfiniteSpacer() },
+        @ViewBuilder midLeading: @escaping () -> MidLeading = { InfiniteSpacer() },
+        @ViewBuilder midCenter: @escaping () -> MidCenter = { InfiniteSpacer() },
+        @ViewBuilder midTrailing: @escaping () -> MidTrailing = { InfiniteSpacer() },
+        @ViewBuilder bottomLeading: @escaping () -> BottomLeading = { InfiniteSpacer() },
+        @ViewBuilder bottomCenter: @escaping () -> BottomCenter = { InfiniteSpacer() },
+        @ViewBuilder bottomTrailing: @escaping () -> BottomTrailing = { InfiniteSpacer() }
```
ianthetechie, line 53: Same discussion re: escaping.

apple/Sources/FerrostarSwiftUI/Views/SpeedLimit/SpeedLimitView.swift
```diff
@@ -0,0 +1,47 @@
...
+                valueFormatter: valueFormatter,
+                unitFormatter: unitFormatter
+            )
+        default:
+            ROWSpeedLimitView(
+                speedLimit: speedLimit,
+                valueFormatter: valueFormatter,
+                unitFormatter: unitFormatter
+            )
+        }
```
ianthetechie, line 34: I am not sure this is the best way to go about deciding which speed limit to show. For one, [the US isn't the only country to use mph for speed limits](https://www.rhinocarhire.com/Drive-Smart-Blog/Which-Countries-use-MPH-or-KPH.aspx).

Second, I'm pretty convinced that locale is not the correct way to set speed display preference.

I think we can punt on the specifics of this for the moment and:

1. Have it be a pair of preferences for the moment (maybe bundled into the theme / display units / formatter stuff??). It's really a pair of preferences: do you want the US/MUTCD style or Vienna Convention style that's used *almost* everywhere else. If I were designing this, I'd make it vary on geography, but that's way out of scope for now ;)
2. Move the logic at the use site (get rid of this view wrapper until we have a better idea of how users want to determine it automatically).

apple/Sources/FerrostarSwiftUI/Views/SpeedLimit/USSpeedLimitView.swift
```diff
@@ -0,0 +1,73 @@
+import SwiftUI
+
+public struct USSpeedLimitView: View {
```
ianthetechie, line 3: I struggled to find the most correct name for this, but it's used outside the US too, so I think we can just call it "US-style" for now.

```suggestion
/// A US-style speed limit sign using a rectangular white box and a black border.
public struct USStyleSpeedLimitView: View {
```

apple/Sources/FerrostarSwiftUI/Views/SpeedLimit/ROWSpeedLimit.swift
```diff
@@ -0,0 +1,60 @@
+import SwiftUI
+
+public struct ROWSpeedLimitView: View {
```
ianthetechie, line 3: Aside: some countries change various parts of this; we can expose those as customizations later on if desired, but don't need to for this PR. Just trying to be a bit less US-centric than Mapbox while also getting the terminology more correct 😂 

```suggestion
/// The common speed circular limit sign with a white background and red border.
/// Also known as the Vienna Convention Sign C14.
public struct ViennaConventionStyleSpeedLimitView: View {
```

ianthetechie:
I also noticed that several of the tests are (now?) nondeterministic.

1. Dark mode probably needs to be explicitly set for snapshot tests.
2. Arrival view tests almost all fail on my machine; snapshots expect feet; my laptop uses meters.
3. Default formatter tests make the same error since they don't explicitly set certain locale details in `testDistanceFormatter` and `testEstimatedArrivalFormatter`.

Archdoog:
1. Add environment settings like dark mode/light mode to tests. Tests need explicit formatters (time zones, etc), environment dark mode & environment locale (units).
2. Add a readme discussion that the simulator needs to exactly match.

apple/Sources/FerrostarSwiftUI/Views/SpeedLimit/USSpeedLimitView.swift
```diff
@@ -0,0 +1,74 @@
+import SwiftUI
+
+/// A US-style speed limit sign using a rectangular white box and a black border.
+public struct USStyleSpeedLimitView: View {
```
CatMe0w, line 4: Instead of "US" and "US-style", how about calling it "Imperial"?
```suggestion
public struct ImperialSpeedLimitView: View {
```

ianthetechie: `Imperial` describes a system of measurement units. But the same style of sign is used in Canada with metric units :P I tried to come up with some better ideas than "US-style" as a classifier, and I _think_ that MUTCD might be correct, but to be honest I had a hard time verifying this.

If anyone can verify that MUTCD is the right term, we can go with that (perhaps `MUTCDStyleSpeedLimitView`). Otherwise, "US-style" is a reasonably accurate descriptor.

apple/Sources/FerrostarMapLibreUI/Views/DynamicallyOrientingNavigationView.swift
```diff
@@ -33,18 +46,30 @@ public struct DynamicallyOrientingNavigationView: View {
...
+        @ViewBuilder topTrailing: @escaping () -> TopTrailing = { InfiniteSpacer() },
+        @ViewBuilder midLeading: @escaping () -> MidLeading = { InfiniteSpacer() },
+        @ViewBuilder bottomTrailing: @escaping () -> BottomTrailing = { InfiniteSpacer() }
     ) {
+        self.theme = theme
         self.styleURL = styleURL
         self.navigationState = navigationState
-        self.distanceFormatter = distanceFormatter
-        userLayers = makeMapContent()
+        userLayers = mapContent()
```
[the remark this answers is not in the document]
Archdoog: Reverted

apple/Sources/FerrostarSwiftUI/Resources/Localizable.xcstrings
```diff
@@ -36,6 +36,29 @@
...
         }
       }
+    },
+    "No Location Available" : {
+
+    },
+    "Rerouting..." : {
+
+    },
+    "Something Useful" : {
```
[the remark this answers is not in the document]
Archdoog: Xcode apparently auto generates localizations from **any** `Text("Something")` view even those in previews 🤣. https://stackoverflow.com/questions/77648648/xcode-15-1-localizable-strings-used-in-previews-are-automatically-added-to-xc

apple/Tests/FerrostarSwiftUITests/Views/__Snapshots__/NavigatingInnerGridViewTests/testUSView.1.png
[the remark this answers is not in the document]
Archdoog: Certainly is. This is just the output from the MeasurementFormatter's speed unit on iOS.

apple/Tests/FerrostarSwiftUITests/Views/__Snapshots__/NavigatingInnerGridViewTests/testROWView.1.png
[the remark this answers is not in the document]
Archdoog: Same as mi/hr. This is what the formatter outputs for short units.

apple/Sources/FerrostarSwiftUI/Views/Controls/FerrostarButton.swift
Archdoog: I'm fine with `StyledActionButton`, but my intention was to prepend Ferrostar so it's easy to follow if someone want's to use the same button style for their own navigation actions(s) outside of the ferrostar scope. Similar to the ViewModifiers, prepending with Ferrostar makes it easy to understand in a final implementation.

apple/Sources/FerrostarSwiftUI/Views/InfiniteSpacer.swift
```diff
@@ -0,0 +1,13 @@
+import SwiftUI
+
+public struct InfiniteSpacer: View {
```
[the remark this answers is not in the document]
Archdoog: Removed it. I was using it to "push" the columns out, but it's easy enough with a normal spacer now.

apple/Sources/UniFFI/ferrostar.swift
ianthetechie, line 1: Did any Rust code change in this PR? If not, I think you might be slightly behind and should merge master into your branch + re-run to ensure this doesn't overwrite things. (master has a rather large PR that, among other things, upgrades UniFFI.)

apple/Sources/FerrostarSwiftUI/Views/Controls/FerrostarButton.swift
Archdoog: How about `NavigationUIButton`

apple/Sources/FerrostarSwiftUI/Views/SpeedLimit/USSpeedLimitView.swift
```diff
@@ -54,20 +54,21 @@ public struct USStyleSpeedLimitView: View {
             .background(Color.white)
             .frame(width: 52, height: 76)
             .cornerRadius(4)
+            .colorScheme(.light)
```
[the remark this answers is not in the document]
Archdoog: This fixes the speed limit sign colors to light mode. Since it's something we don't want to invert.

apple/Sources/FerrostarSwiftUI/Views/SpeedLimit/SpeedLimitView.swift
```diff
@@ -1,38 +1,61 @@
...
+
+    private func useUSStyle() -> Bool {
+        switch toggleStyleBy {
+        case .usStyle:
+            SpeedLimitFixedToUSStyle().useUSStyle()
+        case .viennaConvention:
+            SpeedLimitFixedToViennaConventionStyle().useUSStyle()
+        }
+    }
 }
```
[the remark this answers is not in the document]
Archdoog: I think this should be addressed in: https://github.com/stadiamaps/ferrostar/issues/128. Basically injecting the signage provider throughout the stack, removing any manual toggles/enums.

apple/Sources/FerrostarSwiftUI/Views/SpeedLimit/SpeedLimitView.swift
```diff
@@ -1,38 +1,61 @@
...
     @Environment(\.locale) private var locale
 
+    var toggleStyleBy: ToggleStyleBy
     var speedLimit: Measurement<UnitSpeed>
     var valueFormatter: NumberFormatter
     var unitFormatter: MeasurementFormatter
 
     public init(
         speedLimit: Measurement<UnitSpeed>,
+        toggleStyleBy: ToggleStyleBy = .viennaConvention, // Change the default once we have a better solution.
```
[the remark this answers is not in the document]
Archdoog: Captured here: https://github.com/stadiamaps/ferrostar/issues/128

apple/Sources/FerrostarSwiftUI/Views/SpeedLimit/SpeedLimitView.swift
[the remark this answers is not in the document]
ianthetechie: Fair enough; still a bit hard to follow for the moment, but agree that can be addressed later.

apple/Sources/FerrostarSwiftUI/Views/SpeedLimit/USSpeedLimitView.swift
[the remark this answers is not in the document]
ianthetechie: Leaving a comment for posterity because this was EXTREMELY interesting... Apparently named colors (ex: `.green`) may be aware of light/dark mode, and resolve to different hex values! That's both unexpected but also kinda cool.

Issue #26, opened 2023-11-28
[core] no_std support


ianthetechie:
At some point we'll want to explore support for no_std platforms. Our initial focus is on major mobile operating systems, which have support for the full standard library, but certain embedded platforms we may target in the future may not.

This issue tracks known steps we'll need to take to support no_std:

- [ ] non-UniFFI interface: UniFFI requires the standard library, particularly making use of things like `std::sync::Arc` in constructors, and types like `std::time::SystemTime` for mapping timestamps.
- [ ] We use `std::collections::HashMap` a few places; could probably replace with `alloc::collections::BTreeMap`; we can also use `Vec` from `alloc`
- [ ] `NavigationController` currently uses a `Mutex` for internal state mutation. This is arguably an antipattern and breaks the clean "functional core / imperative shell" pattern. We should probably rearchitect this to return a new controller state (which must be saved by the caller) before v1.0.

ianthetechie:
Another thing to note is that, for the longest time, `core::error::Error` was not stable. That changed today with the release of Rust 1.81 :) So that removes one more constraint on things.

PR #120, opened 2024-06-13
Initial pass at supporting wasm32-unknown-unknown in a web browser

ianthetechie:
First pass at `wasm32-unknown-unknown`, specifically for web browsers.

I'll leave some comments with the key pieces, but broadly speaking, this should have no functional changes (only some error APIs change).

This incidentally does like 95% of the work for #26.


.github/workflows/wasm-js.yml
```diff
@@ -0,0 +1,33 @@
...
+        working-directory: common
+
+  build:
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
+      - name: Cargo Build
+        run: cargo build --verbose --lib --target wasm32-unknown-unknown --no-default-features --features wasm_js
+        working-directory: common
+      # TODO: Figure out a strategy for running unit tests; probably requires a web browser for time and randomness...
```
ianthetechie, line 33: I don't think this is a big deal initially; we can look at this later once we have wasm_bindgen + JS bindings working in a browser for some trivial case.

common/ferrostar/src/models.rs
```diff
@@ -1,20 +1,33 @@
...
-#[derive(Debug, thiserror::Error, uniffi::Error)]
+#[derive(Debug)]
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Error))]
 pub enum ModelError {
-    #[error("Failed to generate a polyline from route coordinates: {error}.")]
+    #[cfg_attr(
+        feature = "std",
+        error("Failed to generate a polyline from route coordinates: {error}.")
+    )]
```
ianthetechie, line 24: For `no_std` platforms, we cannot implement `Error` because the type is from `std`. However, in a few months (October), `Error` is [moving from `std` to `core`](https://github.com/orgs/time-rs/projects/1/views/1?pane=issue&itemId=15671948) :D

.github/workflows/wasm-js.yml
CatMe0w: Is integrating wasm-pack a good idea? wasm-pack can produce an npm package with a JS wrapper to the Rust core, and also has built-in headless browser unit test support built-in.

common/ferrostar/Cargo.toml
```diff
@@ -15,23 +15,33 @@ rust-version.workspace = true
...
-serde_json = "1.0.96"
+serde_json = { version = "1.0.117", default-features = false }
 thiserror = "1.0.40"
-uniffi.workspace = true
+uniffi = { workspace = true, optional = true }
 uuid = { version = "1.8.0", features = ["v4"] }
+getrandom = { version = "0.2.15", optional = true }
+wasm-bindgen = { version = "0.2.92", optional = true }
+web-time = { version = "1.1.0", optional = true }
+wee_alloc = { version = "0.4.5", optional = true }
```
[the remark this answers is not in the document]
ianthetechie: Ooh good catch! I didn't realize it was unmaintained or had security issues.

And yes, I think we can probably use the default allocator. I added this while on a plane without WiFi when I mistakenly thought that wasm32-unknown-unknown was a no_std-only target. Turns out std is supported but a few things like time will panic and randomness needs JS.

.github/workflows/wasm-js.yml
ianthetechie: Quite possibly. I'll take a look at your POC as well as wasm_pack docs before our Monday meeting and try to make an informed decision on this.

I also found a few test harness options that look really promising but many are not that stable yet so I'm fine with trying something first and changing later if necessary.

common/ferrostar/Cargo.toml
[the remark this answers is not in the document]
CatMe0w: And another pain of wasm targets with std is artifact may be big, which can drastically degrade the user experiences. So we may still keep the wasm target with no_std. And for the same purpose, I think we may want to enable LTO/strip symbols in release builds in Cargo configs.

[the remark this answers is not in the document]
ianthetechie: Yeah, we can compare sizes once we actually care; bit of a premature optimization otherwise to remove std. All of the work is done for no_std wasm32-unknown-unknown except for an allocator.

LTO makes sense, but I'm not sold on stripping symbols. That negatively affects the ability of developers to debug issues in or profile production apps accurately, where you always want to ship a release build. Perhaps we distribute two variants though for the oddly space conscious mobile dev. Embedded apps will not use a pre compiled lib and will set their own build flags to strip if they want to on the final application.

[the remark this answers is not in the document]
ianthetechie: Resolved in latest commits.

.github/workflows/wasm-js.yml
ianthetechie: Do you have any comments on this which need to be addressed _before_ this PR is merged @CatMe0w? I think your POC repo will necessarily have to build on the stuff in this branch, so I'd say let's get this merged unless you have any other blocking feedback.

CatMe0w: My current work is built on this branch right now and I think this PR in done and can be merged. Actually I am also waiting for this to be merged as I thought you might still have things to do on this branch. My bad, I should have said that earlier!

ianthetechie: No worries :)

PR #122, opened 2024-06-20
Fix docs for Github Personal Access Token

haysmike:
Looks like the build is looking for `gpr.token` instead of `gpr.key` now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,7 +140,7 @@ cargo install cargo-ndk
    
    ```
    gpr.user=username
-   gpr.key=key
+   gpr.token=token
    ```
 5. Open the Gradle workspace ('android/') in Android Studio.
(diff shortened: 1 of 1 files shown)


Archdoog:
Something broke swift format vs the uniffi generated `ferrostar.swift` file and I had to re-generate it by:

1. Running `build-ios.sh` from within the `common/` folder.
2. Running `swiftformat .` from the root folder. 

This is fixed in #119, but we could probably get this PR here merged first since it's an easier approve if we wanted to fix that here first.

ianthetechie:
Yeah, honestly I wish there was a good way to set up CI workflows so that it's just commit on your branch if you forgot to swiftformat :P There's no good solution... even pre-commit hooks won't work here since you may have an old version of swiftformat.

ianthetechie:
I just quick fixed it to merge both this and #123.

Issue #126, opened 2024-06-30
Crash on arrival


ianthetechie:
This manifests on iOS today with the arrival view on iOS, but I'm pretty sure the root cause is the core. Here's a sample backtrace.

```
Last Exception Backtrace:
0   CoreFoundation                	       0x19c428f20 __exceptionPreprocess + 164
1   libobjc.A.dylib               	       0x1942da018 objc_exception_throw + 59
2   Foundation                    	       0x19b92f868 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 187
3   Foundation                    	       0x19b3af308 -[NSDateComponentsFormatter stringFromTimeInterval:] + 255
4   iOS Demo                      	       0x100424270 closure #1 in ArrivalView.body.getter + 648
5   SwiftUI                       	       0x1a0305324 0x1a03052bd + 103
6   iOS Demo                      	       0x100422cac ArrivalView.body.getter + 616
7   iOS Demo                      	       0x100429280 protocol witness for View.body.getter in conformance ArrivalView + 12
```

TL;DR, there's some value being passed to the `DateComponentsFormatter` which is causing a crash. I tried several "obvious" edge cases like zero, negative zero (the value in question is a `Double`), and -1. Then I remembered that there are plenty more, because doubles... like `NaN`, infinity, and subnormals.

I believe that something in `algorithms.rs` (the `calculate_trip_progress` function) is responsible. I have observed similar issues with doing these sorts of calcs before; sometime even traceable back to the geo rust calculations, which have on occasion not been fully implemented or sound against edge cases (any sort of IEEE-754 division is inherently unsafe.)

Archdoog:
@ianthetechie Closed by #127?

ianthetechie:
Ahh yes, that should have fixed this.

PR #127, opened 2024-06-30
Fix invalid progress calculation

ianthetechie:
This is probably the root cause of #126.

### DIFF
--- a/common/build-ios.sh
+++ b/common/build-ios.sh
@@ -63,11 +63,12 @@ build_xcframework() {
   fi
 }
 
-cargo build --lib --release --target x86_64-apple-ios
-cargo build --lib --release --target aarch64-apple-ios-sim
-cargo build --lib --release --target aarch64-apple-ios
-
 basename=ferrostar
+
+cargo build -p $basename --lib --release --target x86_64-apple-ios
+cargo build -p $basename --lib --release --target aarch64-apple-ios-sim
+cargo build -p $basename --lib --release --target aarch64-apple-ios
+
 generate_ffi $basename
 create_fat_simulator_lib $basename
 build_xcframework $basename
--- a/common/ferrostar/Cargo.toml
+++ b/common/ferrostar/Cargo.toml
@@ -44,6 +44,3 @@ rstest = "0.21.0"
 
 [lib]
 crate-type = ["cdylib", "staticlib", "lib"]
-
-[profile.release]
-lto = "thin"
--- a/common/ferrostar/src/algorithms.rs
+++ b/common/ferrostar/src/algorithms.rs
@@ -273,19 +273,20 @@ fn distance_to_end_of_step(snapped_location: &Point, current_step_linestring: &L
 
 /// Computes the arrival state for a snapped location along the route.
 /// This includes distances and durations.
+///
+/// NOTE: `remaining_steps` includes the current step.
 pub fn calculate_trip_progress(
     snapped_location: &Point,
-    current_step: &RouteStep,
     current_step_linestring: &LineString,
     remaining_steps: &[RouteStep],
 ) -> TripProgress {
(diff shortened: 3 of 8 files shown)


common/ferrostar/src/algorithms.rs
```diff
@@ -273,28 +273,33 @@ fn distance_to_end_of_step(snapped_location: &Point, current_step_linestring: &L
...
+///
+/// NOTE: `remaining_steps` includes the current step.
 pub fn calculate_trip_progress(
     snapped_location: &Point,
-    current_step: &RouteStep,
     current_step_linestring: &LineString,
     remaining_steps: &[RouteStep],
 ) -> TripProgress {
-    if remaining_steps.is_empty() {
+    let Some(current_step) = remaining_steps.first() else {
```
ianthetechie, line 284: Passing current_step explicitly here opened the door for potentially very confusing errors caused by this ambiguity. I've just made it impossible to make that mistake now.

common/ferrostar/Cargo.toml
```diff
@@ -43,7 +43,4 @@ insta = { version = "1.33.0", features = ["yaml"] }
 rstest = "0.21.0"
 
 [lib]
-crate-type = ["cdylib", "staticlib", "lib"]
-
-[profile.release]
-lto = "thin"
\ No newline at end of file
+crate-type = ["cdylib", "staticlib", "lib"]
```
ianthetechie, line 46: Linter catch: this didn't have any effect in a workspace sub-project ;)

common/build-ios.sh
```diff
@@ -63,11 +63,12 @@ build_xcframework() {
...
 
-cargo build --lib --release --target x86_64-apple-ios
-cargo build --lib --release --target aarch64-apple-ios-sim
-cargo build --lib --release --target aarch64-apple-ios
-
 basename=ferrostar
+
+cargo build -p $basename --lib --release --target x86_64-apple-ios
+cargo build -p $basename --lib --release --target aarch64-apple-ios-sim
+cargo build -p $basename --lib --release --target aarch64-apple-ios
```
ianthetechie, line 70: Minor improvement: we don't need to build `uniffi-bindgen` for any other architectures.

common/ferrostar/src/algorithms.rs
```diff
@@ -262,39 +263,43 @@ fn distance_along(point: &Point, linestring: &LineString) -> Option<f64> {
...
 
     // This could be improved with live traffic data along the route.
     // TODO: Figure out the best way to enable this use case
-    let pct_remaining_current_step =
-        distance_to_next_maneuver / current_step_linestring.haversine_length();
+    let pct_remaining_current_step = if current_step.distance > 0f64 {
+        distance_to_next_maneuver / current_step.distance
+    } else {
+        0f64
+    };
```
ianthetechie, line 302: This was the core issue. If the snapped point is exactly at the end of the step, this results in division by zero.

common/ferrostar/src/algorithms.rs
```diff
@@ -476,6 +481,37 @@ proptest! {
...
+        prop_assert_eq!(progress.duration_remaining, 0f64);
+    }
+
+    #[test]
+    fn test_end_of_trip_progress_valhalla_arrival(
+        x1: f64, y1: f64,
+    ) {
+        // This may look wrong, but it's actually how Valhalla (and presumably others)
+        // represent a point geometry for the arrival step.
+        let current_route_step = gen_dummy_route_step(x1, y1, x1, y1);
```
ianthetechie, line 506: And this is why I noticed the issue; the last maneuver's geometry is the same point twice, so its length is zero; thus, the division by zero above, which ended up crashing one of the iOS formatters in the arrival view with a `NaN`.

Archdoog (approved):

PR #129, opened 2024-07-02
Initial proof-of-concept web component library + demo app

CatMe0w:
This is the initial implementation of Ferrostar web component library.

These lists are incomplete and may be expanded in the future.

### TODOs:
- [ ] TypeScript types
- [ ] `RouteAdapter` with combinations other than Valhalla and OSRM
- [ ] `RouteDeviationDetector`
- [x] `locationProvider` from simulated locations
- [x] `locationProvider` from Web Geolocation API
- [ ] `spokenInstruction`

### Known issues:
- [x] Non-WASM core builds need passing `--all-features`


CatMe0w:
The first JS code has been pushed. Currently, it can only fetch and parse data from the Valhalla server and render static routes on the map.

ianthetechie:
Thanks! I'd say none of these TODOs need to happen for this PR. Many small PRs with some subset of functionality are better than a huge "perfect" one after a month of work :)

I'll try this out and give any feedback on the code over the next few days!

CatMe0w:
Here we are, a working live demo, just like the Android and iOS demo apps.
However, this is still primitive and the technical details need to be discussed further.

CatMe0w:
We have a real location demo now. ;)

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -3,9 +3,11 @@ use crate::models::{RouteStep, SpokenInstruction, UserLocation, VisualInstructio
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
 use geo::LineString;
+#[cfg(feature = "wasm-bindgen")]
+use serde::{Deserialize, Serialize};
 
 /// A subset of state values that are used to show the user their current progress along the trip and it's components.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
```
ianthetechie, line 10: I'm guessing this is at least one cause of the known limitation you mentioned with builds not working. You've unconditionally derived these traits, but the import is (correctly) conditional. You can use `cfg_attr` here like you do below :)

(btw props for getting this far with features! This is a pretty advanced part of cargo that many rust devs never really need to learn :D)

common/ferrostar/src/simulation.rs
```diff
@@ -2,6 +2,11 @@ use crate::algorithms::trunc_float;
 use crate::models::{CourseOverGround, GeographicCoordinate, Route, UserLocation};
 use geo::{coord, DensifyHaversine, GeodesicBearing, LineString, Point};
 use polyline::decode_polyline;
+use serde::{Deserialize, Serialize};
```
ianthetechie, line 5: I think this _probably_ needs to be feature gated (unless I missed a reason we need it for other targets).

web/src/ferrostar-core.ts
```diff
@@ -0,0 +1,141 @@
...
+  @property({ attribute: false })
+  httpClient?: Function = fetch;
+
+  // FIXME: type
+  @property({ type: Object })
+  locationProvider!: any;
+
+  // FIXME: type
+  @property({ type: Object })
+  costingOptions!: any;
```
ianthetechie, line 27: Just to make suer I understand, the FIXME is because these should eventually be a stronger type, right?

I appreciate your commitment to perfection, but this can probably be a TODO for now (and if my understanding is correct, we should be able to get this merged as-is) :D

ianthetechie (commented):
I'm adding a partial review right now even though I've only had a chance to properly review about half of the code (most of the bindgen-specific and some of the web code will require me doing a bit more research of my own).

Overall looking good! Just a few comments on how we can improve things. I'll try to review the rest over the coming days. In the meantime, this PR is starting to get a bit large, so maybe you could add any additional new _feature_ development in a branch off of this; that will let us focus first on QAing what's here already and then taking enhancements in the next pass (which will be smaller and easier to review).

Thanks, and have a great weekend!

ianthetechie:
If you could also fix up all the merge conflicts (on this branch), that'd also help us review this; GitHub does strange things to the diff with conflicts present in my experience.

common/ferrostar/src/models.rs
```diff
@@ -79,18 +77,16 @@ impl From<GeographicCoordinate> for Point {
...
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
-#[cfg_attr(test, derive(Serialize))]
 pub struct Waypoint {
     pub coordinate: GeographicCoordinate,
     pub kind: WaypointKind,
 }
 
 /// Describes characteristics of the waypoint for the routing backend.
-#[derive(Clone, Copy, PartialEq, PartialOrd, Debug)]
+#[derive(Clone, Copy, PartialEq, PartialOrd, Debug, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
CatMe0w: The reason I left these derives unconditioned is Rust doesn't allow duplicating traits, like:
```
#[cfg_attr(test, derive(Serialize))]
#[cfg_attr(feature = "wasm-bindgen", derive(Deserialize, Serialize))]
```
This will cause an error and i didn't want to make compiler conditions too huge, so I decided to let the compiler remove the unused codes.

However, I've found another approach to keep correct conditions and won't make a very long `cfg` block. Let's see if this will work.

CatMe0w:
I accidentally merged everything into my fork. Force pushed to rollback that.

CatMe0w:
CI passes. Looks like I can't resolve conflicts from my side without merging the current `main` branch into my fork, but these conflicts are trivial, mainly consisting of comment updates after this PR was created. I believe it is ready for review and merge.

ianthetechie:
Thanks; I'll take care of the merge conflicts and finish up the review within the next 24h!

ianthetechie:
I've resolved the merge conflicts but hit a few issues when trying to reproduce your build instructions locally.

I've updated the CI to reflect your instructions, and it fails in the same way my local builds do (good news since it's reproducible haha).

ianthetechie:
Heeeeeeey! Not bad! It looks like the route simulation works! The issue of sorts appears to be with the tsconfig. Thoughts?

https://github.com/stadiamaps/ferrostar/assets/2140440/1aae8543-e230-4bb8-b406-c5f61e7ce339

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -250,3 +253,48 @@ impl NavigationController {
...
+    pub fn new(route: JsValue, config: JsValue) -> Result<JsNavigationController, JsValue> {
+        let route: Route = serde_wasm_bindgen::from_value(route)?;
+        let config: NavigationControllerConfig = serde_wasm_bindgen::from_value(config)?;
+
+        Ok(JsNavigationController(NavigationController::new(
+            route, config,
+        )))
+    }
+
+    pub fn get_initial_state(&self, location: JsValue) -> Result<JsValue, JsValue> {
```
ianthetechie, line 275: I'm curious if there is a better way to avoid this newtype dance and return type erasure with `JsValue` on both sides. Let me know if you have any thoughts + maybe we can discuss on the call later.

common/ferrostar/src/routing_adapters/mod.rs
```diff
@@ -174,3 +179,57 @@ impl RouteAdapter {
         self.response_parser.parse_response(response)
     }
 }
+
+/// JavaScript wrapper for `RouteAdapter`.
```
ianthetechie, line 183: I have the same question here as on the navigation controller.

web/index.html
```diff
@@ -0,0 +1,111 @@
...
+          const waypoints = [{ coordinate: { lat: parseFloat(lat), lng: parseFloat(lon) }, kind: "Break" }];
+
+          const locationProvider = new BrowserLocationProvider();
+          locationProvider.requestPermission();
+          locationProvider.start();
+
+          // This approach is not ideal, any better way to wait for the locationProvider to acquire the first location?
+          while (!locationProvider.lastLocation) {
+            await new Promise((resolve) => setTimeout(resolve, 100));
+          }
```
ianthetechie, line 71: Good question; I'll need to review the MDN docs on this.

web/index.html
```diff
@@ -0,0 +1,111 @@
...
+
+          const routes = await core.getRoutes(locationProvider.lastLocation, waypoints);
+          const route = routes[0];
+
+          // FIXME: we shouldn't need to set the locationProvider again
```
ianthetechie, line 76: Where is the first time it's being set? It looks like this is the first time to me.

web/index.html
```diff
@@ -0,0 +1,111 @@
...
+
+          const routes = await core.getRoutes(initialSimulatedLocation, simulatedWaypoints);
+          const route = routes[0];
+
+          const locationProvider = new SimulatedLocationProvider();
+          locationProvider.lastLocation = initialSimulatedLocation;
+          locationProvider.warpFactor = 2;
+          locationProvider.setSimulatedRoute(route);
+
+          // FIXME: we shouldn't need to set the locationProvider again
```
ianthetechie, line 104: Same question as above.

web/src/ferrostar-core.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  constructor() {
+    super();
+
+    // A workaround for avoiding "Illegal invocation"
+    if (this.httpClient === fetch) {
+      this.httpClient = this.httpClient.bind(window);
+    }
+
+    // A workaround for loading the marker icon images in Vite
```
ianthetechie, line 52: Any links to known issues? I'm fine not having this figured out for the current PR; just want to know what's outstanding for cleanup.

ianthetechie (commented):
Really coming along! I have a bunch of minor questions now that I've gotten deepen into it. Let's discuss on the call later.

web/tsconfig.json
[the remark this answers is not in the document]
CatMe0w: These are all default configs, generated directly by Vite.

web/src/location.ts
```diff
@@ -0,0 +1,81 @@
...
+export class BrowserLocationProvider {
+  private geolocationWatchId: number | null = null;
+  lastLocation: any = null;
+  lastHeading = null;
+  warpFactor = 1;
+
+  updateCallback: () => void = () => {};
+
+  requestPermission() {
+    navigator.permissions.query({ name: "geolocation" });
```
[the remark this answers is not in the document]
CatMe0w: This will pop up a permission request window on the browser. However, I just found that `navigator.geolocation.watchPosition` can trigger the prompt as well. I will look into if we can remove this function.

common/ferrostar/src/navigation_controller/mod.rs
CatMe0w: This was my biggest concern and a significant blocker during my early work. Initially, I tried to reuse the existing `uniffi::export` blocks, but `Arc`s don't make sense in JavaScript. Implementing `WasmDescribe` trait seems like a possible approach, but the workload could be insane. And according to wasm-bindgen's documentation, using `JsValue` is the recommended way to export structs to JavaScript.

web/index.html
CatMe0w: We have already set it in the HTML tag. Lit should be able to trace the changes on `locationProvider`...

web/src/ferrostar-core.ts
CatMe0w: This issue: https://stackoverflow.com/questions/44720448/fetch-typeerror-failed-to-execute-fetch-on-window-illegal-invocation
Maybe a quirk of Web Component or Lit.

web/src/location.ts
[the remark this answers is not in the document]
ianthetechie: Yeah, as far as I can tell, calls like `watchPosition` will result in browser prompts when needed. I'd say we can remove this method, unless the point is to update some state which is observed elsewhere (ex: to show a "location unavailable" indicator on the page, but that's not an immediate requirement.)

common/ferrostar/src/navigation_controller/mod.rs
ianthetechie: Gotcha; thanks! I'll do a bit of thinking on whether there's a better long-term solution. This feels like a case where we should be able to use a macro to reduce the boilerplate, but it's good enough as-is to merge. @Archdoog and I will experiment with other approaches.

web/index.html
ianthetechie: Whoops; I misread this initially. This isn't specific to the geolocation API :D 

For the `BrowserLocationProvider` we can basically just wrap this function which calls back once it gets the location (or errors): https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition.

For the `SimulatedLocationProvider` we'd have to mimic the interface ourselves but it shouldn't be too hard.. I think adding a list of waiting callbacks or something that requested the location (slightly different from `updateCallback` which is more of a long-term subscription).

web/src/location.ts
```diff
@@ -0,0 +1,81 @@
...
+      }
+
+      this.simulationState = updatedState;
+      this.lastLocation = updatedState.current_location;
+
+      // Since Lit cannot detect changes inside objects, here we use a callback to trigger a re-render
+      // This is a minimal approach if we don't want to use a state management library like MobX, but might not be the ideal solution
+      if (this.updateCallback) {
+        this.updateCallback();
+      }
```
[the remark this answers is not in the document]
ianthetechie: Discussed on the call; can revisit in a later PR.

ianthetechie:
Ok, tests should compile and run now! Note that there aren't actually any wasm-bindgen tests yet, but this will at least enable us to write those in the future.

PR #138, opened 2024-07-07
Fix half-implemented onTapExit changes

ianthetechie:
During the review of the PR for this, we decided to make `onTapExit` optional so that if no handler was present, we wouldn't display a useless button. It looks like this only got applied to the very lowest level of the view hierarchy though, so it doesn't behave as expected in apps.


ianthetechie:
Note: I dropped `@escaping` because I don't think we need it for all cases. If we hit annoyances in implementation, we can add it back.

ianthetechie:
Second note: I'm tempted to revert a subtle change to the argument order. I assume that customizing map layers will be a pretty common thing, and it's very convenient to be able to do this with a trailing closure. The last PR subtly broke this assumption (and all my debug layers) but didn't actually cause compiler errors 😂

Any thoughts on keeping the map content builder last?

Archdoog:
> Note: I dropped `@escaping` because I don't think we need it for all cases. If we hit annoyances in implementation, we can add it back.

An optional closure is `@escaping` by default 😉. A Swift subtlety. You can pretty much trust the compiler on whether it's needed or not these days.

Archdoog (approved):

PR #140, opened 2024-07-10
Correct a bad package name in contributing docs

CatMe0w:
It appears that the Rust core package name was changed some time ago. Since CI does not use this command for testing, we were not aware of this issue until now.


ianthetechie:
Oh heh 😂 Good catch!

PR #149, opened 2024-07-14
Merge Priority! Android: Upgraded to Versions Catalog (libs.versions.toml)

### LINKED ISSUE #130: Android - Migrate to version catalog
Since we've got a multi-module project & the newer android studio version catalog has all the features for highlighting updates, etc. We should go ahead and upgrade sooner than later: https://developer.android.com/build/migrate-to-catalogs. This will let us easily bump shared versions across the demo app and modules.

Archdoog:
- Upgraded our gradle architecture to the version catalog

This likely updates a handful of packages, but seems to build, test, etc.

Closes #130

### DIFF
--- a/android/core/build.gradle
+++ b/android/core/build.gradle
@@ -64,7 +64,6 @@ dependencies {
     // ValhallaCoreTest.kt
     androidTestImplementation libs.okhttp.mock
     androidTestImplementation libs.kotlinx.coroutines.test
-
     androidTestImplementation libs.androidx.test.junit
     androidTestImplementation libs.androidx.test.espresso
 }
--- a/android/core/build.gradle
+++ b/android/core/build.gradle
@@ -56,7 +56,7 @@ dependencies {
     implementation(libs.okhttp.core)
     implementation libs.moshi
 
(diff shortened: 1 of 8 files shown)


android/core/build.gradle
```diff
@@ -41,30 +41,33 @@ android {
...
+    implementation libs.kotlinx.coroutines
+    implementation libs.kotlinx.datetime
+    implementation libs.androidx.lifecycle.viewmodel.ktx
+
+    implementation(platform(libs.okhttp.bom))
+    implementation(libs.okhttp.core)
+    implementation libs.moshi
+
+    // TODO: Fix this. Maybe related?: https://github.com/gradle/gradle/issues/21267
+    //noinspection UseTomlInstead
```
ianthetechie, line 60: Ooh interesting. I'm just going to suggest making the comment a bit clearer so it's obvious what's broken.

```suggestion
    // TODO: Migrate version to TOML (doesn't work). Likely related issue: https://github.com/gradle/gradle/issues/21267
    //noinspection UseTomlInstead
```

ianthetechie (approved):
Nice! I've been meaning to do this but never got around to it.

We don't have a more sophisticated Android demo outside the project yet, so I'll assume that if it passes CI then it's good.

PR #154, opened 2024-07-15
Fix duplicated markers in web

CatMe0w:
The title says it all. ╰(*°▽°*)╯

### DIFF
--- a/web/src/ferrostar-core.ts
+++ b/web/src/ferrostar-core.ts
@@ -56,7 +56,7 @@ class FerrostarCore extends LitElement {
   }
 
   updated(changedProperties: any) {
-    if (changedProperties.has('locationProvider') && this.locationProvider) {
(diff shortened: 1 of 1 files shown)


web/src/ferrostar-core.ts
```diff
@@ -137,6 +139,14 @@ class FerrostarCore extends LitElement {
     this.currentLocationMapMarker!.setLatLng(this.locationProvider.lastLocation.coordinates);
   }
 
+  private resetMap() {
+    this.map!.eachLayer((layer) => {
+      if (layer instanceof L.Marker || layer instanceof L.Polyline) {
+        this.map!.removeLayer(layer);
+      }
+    });
+  }
```
ianthetechie, line 148: This isn't great, but I don't know the best way to do it in Leaflet so I'll merge anyways ;) We can revisit after switching to MapLibre GL JS.

PR #160, opened 2024-07-18
Feat/ios/dynamic navigation view

### LINKED ISSUE #39: First pass at snapped/following camera calculation
We want the core to calculate most parameters so that all rendering frontends can implement the standard "following" camera consistently.

This information should be returned as part of `TripState`.

Components needed:
- [X] Snapped user location
- [x] ~~Zoom level~~ Distance (can be simplistic at first, but may get more dynamic over time based on things like distance to the next maneuver and speed)
- [X] Pitch (Configurable parameter set at init)
- [X] ~~Smoothed~~ heading (this will determine the orientation of the user's location arrow)
- [X] ~~Smoothed~~ Course over ground (the user's direction of actual travel, independent of compass readings, which can be quite bad in some situations)
- [x] Current deviation from the route line; useful for frontends to decide if they should show the real location or snapped one.

Archdoog:
- Finishes the DynamicallyOrientingNavigationView (for now)
- **Breaking changes on args**! Though it seems necessary to sync up with Android & simplify.
- Removes hidden padding from InstructionView.

Closes #39 (not directly, but I think it safely wraps up the discussion)


apple/Sources/FerrostarMapLibreUI/Views/DynamicallyOrientingNavigationView.swift
```diff
@@ -69,28 +64,44 @@ public struct DynamicallyOrientingNavigationView<
...
+                )
+            default:
+                PortraitNavigationView(
+                    styleURL: styleURL,
+                    camera: $camera,
+                    navigationCamera: navigationCamera,
+                    navigationState: navigationState,
+                    onTapExit: onTapExit,
+                    makeMapContent: { userLayers },
+                    topCenter: { topCenter },
```
ianthetechie, line 94: Something about the way we are doing this transition isn't quite right. I think part of the issue is that we end up recreating the underlying map view. This isn't necessarily bad if we can bootstrap the new map view to the correct state, but right now there are a t least two major issuses:

1. Location updates in MapLibre lag (somewhat by design); this isn't normally an issue, but it prevents quickly re-tracking the user (this is the sort of thing that makes me think Meta is right with obsessing over low-level camera control haha).
2. The initial camera isn't preserved at all (we rely on MapLibre's location tracking, but that's not going to work given the first issue). The map is reset at null island and then tries to fly to the user. We needed to fix this anyways with the plain portrait view, but switching orientations really highlights the problem.

apple/Sources/FerrostarMapLibreUI/Extensions/MapViewCamera.swift
```diff
@@ -0,0 +1,23 @@
...
+    /// The default camera configured for navigation.
+    ///
+    /// - Parameters:
+    ///   - zoom: The zoom value (default is 18.0)
+    ///   - pitch: The pitch (default is 45.0)
+    /// - Returns: The configured MapViewCamera
+    static func navigation(zoom: Double = 18.0, pitch: Double = 45.0) -> MapViewCamera {
+        MapViewCamera.trackUserLocationWithCourse(zoom: zoom,
+                                                  pitch: pitch)
+    }
```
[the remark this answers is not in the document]
ianthetechie: We can figure out a better name + solution later once we have more concrete examples.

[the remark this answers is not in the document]
Archdoog: Agreed: https://github.com/stadiamaps/ferrostar/issues/165

apple/DemoApp/Demo/DemoNavigationView.swift
```diff
@@ -66,20 +66,19 @@ struct DemoNavigationView: View {
...
-                useSnappedCamera: $snappedCamera
-            ) {
-                let source = ShapeSource(identifier: "userLocation") {
-                    // Demonstrate how to add a dynamic overlay;
-                    // also incidentally shows the extent of puck lag
-                    if let userLocation = locationProvider.lastLocation {
-                        MLNPointFeature(coordinate: userLocation.clLocation.coordinate)
+                navigationState: ferrostarCore.state,
+                onTapExit: { stopNavigation() },
+                makeMapContent: {
```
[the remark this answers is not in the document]
Archdoog: I've moved all of this to the `innerGrid` modifier function. While it's not a perfect solution, it's much nicer. Happy to discuss some of the challenges on that function if you're curious.

apple/Sources/FerrostarMapLibreUI/Views/DynamicallyOrientingNavigationView.swift
Archdoog: This is much improved with a singular navigation map view. It required a decent bit of rework, but in testing looks pretty sharp now

Archdoog:
I checked the failed snapshot test diff & it was just a bit of the text rendering/alias on the speed limit sign. Probably need to add a percentage diff allowance here, but we can deal with that later.

apple/Sources/FerrostarMapLibreUI/Views/DynamicallyOrientingNavigationView.swift
```diff
@@ -24,78 +19,103 @@ public struct DynamicallyOrientingNavigationView<
...
     private let userLayers: [StyleLayerDefinition]
 
-    var topCenter: TopCenter
-    var topTrailing: TopTrailing
-    var midLeading: MidLeading
-    var bottomTrailing: BottomTrailing
+    public var topCenter: (() -> AnyView)?
+    public var topTrailing: (() -> AnyView)?
+    public var midLeading: (() -> AnyView)?
+    public var bottomTrailing: (() -> AnyView)?
```
ianthetechie, line 25: Do we need to add these to the constructor? Or I guess there's an extension so maybe the idea is that you don't need to override(?) these very often so it's easier to just leave it to extension methods?

apple/Sources/FerrostarSwiftUI/Views/GridViews/NavigatingInnerGridView.swift
```diff
@@ -22,10 +17,10 @@ public struct NavigatingInnerGridView<
...
     // MARK: Customizable Containers
 
-    var topCenter: TopCenter
-    var topTrailing: TopTrailing
-    var midLeading: MidLeading
-    var bottomTrailing: BottomTrailing
+    public var topCenter: (() -> AnyView)?
+    public var topTrailing: (() -> AnyView)?
+    public var midLeading: (() -> AnyView)?
+    public var bottomTrailing: (() -> AnyView)?
```
ianthetechie, line 23: Same question as above; different class ;)

ianthetechie (approved):
Just some minor comments; this looks fantastic in testing though, so I'll go ahead and merge!

apple/Sources/FerrostarMapLibreUI/Views/DynamicallyOrientingNavigationView.swift
Archdoog: Nope. To keep the constructors sane, I think these only get modified exclusively through the extension view modifier style.

PR #161, opened 2024-07-18
Add concurrency groups to all PR related CI

### LINKED ISSUE #2: Set up CI test and lint actions
This will programmatically enforce standards and make code reviews a lot easier.

Rust:
- [ ] Basic build and test
- [ ] Formatting checks
- [ ] clippy lints
- [ ] cargo-semver?

Apple:
- [ ] Basic build and test
- [ ] TBD formatting check (swiftfmt?)
- [ ] swiftlint

Android:
- [ ] Basic build and test (don't forget to run normal tests *and* android tests!)
- [ ] TBD formatting checks
- [ ] TBD lints

### LINKED ISSUE #150: CI: Github action concurrency label to cancel stale workflows.
We want to add concurrency labels on all CI tasks so that Github can cancel stale tasks when a new push comes in. To do this, we need to add:

```yml
concurrency:
      group: ${{ github.workflow }}-${{ github.ref }}-build
      cancel-in-progress: true
```

To each task at the same level as `runs-on`. Important: we must make sure the group is per task per ref. Meaning it construct a group string like `android-mybranchesref-build`, `android-mybranchesref-test`, etc. That way the task only cancels its match for that branch/PR.

Some tasks like releases should not get this.

Archdoog:
Closes #150

We also might consider #2 closed at this point, but not sure.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -10,6 +10,9 @@ jobs:
   build:
 
     runs-on: macos-13
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}-android-build
+      cancel-in-progress: true
     permissions:
       contents: read
       packages: read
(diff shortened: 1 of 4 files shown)


ianthetechie:
Good call! Only reason I'm still leaving #2 open is because I think we probably want lints eventually, but it's just never been enough of a priority, and there are some false positives right now (ex: speed stuff that's there but never actually used) due to work in progress.

PR #163, opened 2024-07-23
Prevent auto lock screen while navigating on iOS

CatMe0w:
We certainly wouldn't want the phone to automatically lock the screen during navigation.

But I'm not so sure this should be added directly to the core, instead of the demo app. However, core itself comes with such behavior that may be unexpected, so I'll just add this to the demo app, leaving the choice to downstream developers and users.

### DIFF
--- a/apple/DemoApp/Demo/DemoNavigationView.swift
+++ b/apple/DemoApp/Demo/DemoNavigationView.swift
@@ -196,11 +196,14 @@ struct DemoNavigationView: View {
             route: route,
             config: config
         )
+
+        preventAutoLock()
     }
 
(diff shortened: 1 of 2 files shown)


ianthetechie (commented):
This is great; thanks! And I agree this is something that we should leave to app developers, not put in the core.

Can you also add a short paragraph (doesn't even need to be a full example; just reference + link to the Apple API docs) in the User Guide (iOS Tutorial chapter probably)? Just edit the markdown files under the `guide` folder; it's an mdbook project.

ianthetechie (approved):

Issue #155, opened 2024-07-16
Use MapLibre GL JS instead of Leaflet


CatMe0w:
Leaflet lacks built-in support for rotation and pitch, making it unsuitable for navigation purposes. Additionally, placing a marker that follows the user's orientation is challenging. Although https://github.com/bbecquet/Leaflet.RotatedMarker is mentioned in Leaflet's official documentation, it appears impractical and even ridiculous for navigation:
![image](https://github.com/user-attachments/assets/2f207134-6fc4-4e26-899a-e50bb2384c2b)

Switching to MapLibre GL JS could be a optional long term goal, after stabilizing UI. This would also help reduce the need for workarounds related to Leaflet quirks in modern web toolchains.

ianthetechie:
Closed by #166.

PR #166, opened 2024-07-29
Transition to MapLibre GL JS

### LINKED ISSUE #152: Camera view moving for web components
In iOS and Android implementations, there are 3D maps with views that can be rotated to follow the user's direction, but only static 2D maps with a fixed perspective are available in web-based Leaflet. (While it is possible to add third-party plugins to enable rotation, it doesn't make sense for city maps with text.)

Is it necessary to have the view zoom in and follow the user's position automatically?

CatMe0w:
This closes #152 and #155.

https://github.com/user-attachments/assets/15a04174-2aba-49ad-9b1a-a59c0b30d215

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -9,9 +9,15 @@
     <script type="module" src="/src/location.ts"></script>
   </head>
   <body>
+    <!--
+	This should not require any API key if run via `npm run dev`.
+	Learn how to get a free prod API key here: https://docs.stadiamaps.com/authentication/.
+	See https://stadiamaps.github.io/ferrostar/vendors.html for more info on basemap vendors.
+      -->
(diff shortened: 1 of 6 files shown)


CatMe0w:
wait what?!

web/index.html
ianthetechie, line 1: Thanks for the extra comments on here; these are great 💯

ianthetechie (approved):
Looking great! One minor thing: I'm going to switch the demo to the Stadia Maps tiles since the MapLibre demo tiles aren't really usable 😂 We don't require a key for localhost development anyways, and the GitHub Pages domain here is whitelisted so we can host a demo (on my TODO list).

PR #167, opened 2024-07-30
Synced android and ios navigation state, refactored NavigationViewModel

Archdoog:
- Updates ferrostar rust server to 0.6.0
- Includes breaking changes to FerrostarCore init, NavigationViewModel and others.
- Refactors iOS NavigationState to rely on TripState like Android.
- Adds `NavigationControllerConfig` set on `FerrostarCore` init.
- Adds `Idle` to `TripState` to allow presetting value for android dependency injection. This allows us to properly construct the StateFlow at app launch.
- Revises startNavigation to override config optionally.
- Refactors `NavigationViewModel` for better dependency injection.

### DIFF
--- a/apple/Sources/FerrostarCore/FerrostarCore.swift
+++ b/apple/Sources/FerrostarCore/FerrostarCore.swift
@@ -85,24 +85,25 @@ public protocol FerrostarCoreDelegate: AnyObject {
     private let routeProvider: RouteProvider
     private let locationProvider: LocationProviding
     private var navigationController: NavigationControllerProtocol?
-    private var tripState: TripState?
     private var routeRequestInFlight = false
     private var lastAutomaticRecalculation: Date? = nil
     private var lastLocation: UserLocation? = nil
     private var recalculationTask: Task<Void, Never>?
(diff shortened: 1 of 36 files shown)


apple/Sources/FerrostarCore/FerrostarCore.swift
```diff
@@ -359,8 +360,9 @@ extension FerrostarCore: LocationManagingDelegate {
         update(newState: newState, location: location)
     }
 
-    public func locationManager(_: LocationProviding, didUpdateHeading newHeading: Heading) {
-        state?.heading = newHeading
+    public func locationManager(_: LocationProviding, didUpdateHeading _: Heading) {
+        // TODO: Make use of heading in TripState?
```
Archdoog, line 364: I suspect we probably want this to be attached to the UserLocation -> Course somehow? But we should discuss & consider a new ticket. Don't think it was actually doing anything.

Archdoog: https://github.com/stadiamaps/ferrostar/issues/169

PR #176, opened 2024-08-06
Added Osrm to route builder for custom route adapters

Archdoog:
- Adds method to createRouteFromOsrm to simplify import for custom route adapters that use OSRM (e.g. `valhalla-mobile`)
- Revises `OsrmResponseParser` to use new route creation method. _Note:_ this architecture bundles the via_waypoints and waypoints into each Route init. While this is technically extra work, I figured it was the lesser cost of duplicating and creating extra function entry points for pre-built waypoints.
- Renames ReponseParserError to OsrmParserError. I'd be interested in better naming here, but given some of the error naming TODOs, figured we might want to strategize more globally where we're going with an error system. Especially how tidily rust handles error translation.


common/ferrostar/src/routing_adapters/error.rs
```diff
@@ -53,7 +53,7 @@ impl From<serde_json::Error> for RoutingRequestGenerationError {
 #[derive(Debug)]
 #[cfg_attr(feature = "std", derive(thiserror::Error))]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Error))]
-pub enum RoutingResponseParseError {
+pub enum OsrmParsingError {
```
ianthetechie, line 56: Is something about this specific to OSRM? Perhaps I'm too optimistic, but I think we can have a single error enum that covers what devs care about for most backends.

common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -35,21 +37,35 @@ impl OsrmResponseParser {
...
 
 impl RouteResponseParser for OsrmResponseParser {
-    fn parse_response(&self, response: Vec<u8>) -> Result<Vec<Route>, RoutingResponseParseError> {
+    fn parse_response(&self, response: Vec<u8>) -> Result<Vec<Route>, OsrmParsingError> {
         let res: RouteResponse = serde_json::from_slice(&response)?;
-        let via_waypoint_indices: HashSet<_> = res
-            .routes
+
+        // This isn't the most functional in style, but it's a bit difficult to construct a pipeline
+        // today. Stabilization of try_collect may help.
```
ianthetechie, line 44: I think there's a way to do this but I can look into that on my own time.

The thing I failed to notice though is that we just silently discard all errors 😂 I think that's on me, and I think it's probably wrong; we should bail on errors.

common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -65,50 +81,46 @@ impl RouteResponseParser for OsrmResponseParser {
...
+            Err(OsrmParsingError::ParseError {
+                error: "Bounding box could not be calculated".to_string(),
+            })
+        }
     }
 }
 
 impl RouteStep {
-    fn from_osrm(
+    pub fn from_osrm(
```
ianthetechie, line 118: Does this need to be public?

android/core/src/main/java/com/stadiamaps/ferrostar/core/extensions/RouteExtensions.kt
```diff
@@ -0,0 +1,23 @@
...
+ * automatically map & combine via and break waypoints from the route and waypoint data objects.
+ *
+ * @param route The encoded JSON data for the OSRM route.
+ * @param waypoints The encoded JSON data for the OSRM waypoints.
+ * @param polylinePrecision The polyline precision.
+ * @return The navigation [Route]
+ */
+fun Route.Companion.fromOsrm(
+    route: ByteArray,
+    waypoints: ByteArray,
```
ianthetechie, line 19: Is a caller of this function (and the Swift version and the underlying Rust) likely to have waypoints as raw bytes? Or as something higher level? I honestly don't know since I'm not a likely user of it.

common/ferrostar/src/routing_adapters/osrm/mod.rs
ianthetechie: So TIL you don't actually need `try_collect` to do what you want here! And, serendipitously, I actually needed this exact same pattern on some code I was working on over the weekend. The solution is described in [this comment](https://github.com/rust-lang/rust/issues/94047#issuecomment-1041432644) rather tersely with some code. What isn't obvious is that you can actually write your return type as `Result<Vec<_>, _>` and `collect` will do the right thing.

common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -65,50 +76,43 @@ impl RouteResponseParser for OsrmResponseParser {
...
                 }
-
-                routes.push(Route {
-                    geometry,
-                    bbox: bbox.into(),
-                    distance: route.distance,
-                    waypoints: waypoints.clone(),
-                    steps,
-                });
             }
```
ianthetechie, line 95: Can also be a functional combinator.

android/core/src/main/java/com/stadiamaps/ferrostar/core/extensions/RouteExtensions.kt
ianthetechie: Discussed on the call.

PR #180, opened 2024-08-12
Add integrated search box, and other minor improvments

CatMe0w:
## Add integrated search box
image

## Move stop button to `ferrostar-core`
image

It looks more like iOS implementation now.

## Minor changes
- Implemented a proper way for serializing/deserializing Rust's `SystemTime` from JS.
- Removed animation when starting navigation to avoid initiating too many map tile requests.
- Other internal bug fixes.

### DIFF
--- a/web/src/ferrostar-core.ts
+++ b/web/src/ferrostar-core.ts
@@ -177,7 +177,10 @@ export class FerrostarCore extends LitElement {
       },
     });
 
-    this.map?.setCenter(route.geometry[0]);
+    this.map?.easeTo({
+      center: route.geometry[0],
+      bearing: startingLocation.courseOverGround?.degrees || 0,
+    });
 
     this.currentLocationMapMarker = new maplibregl.Marker().setLngLat(route.geometry[0]).addTo(this.map!);
   }
--- a/web/src/ferrostar-core.ts
+++ b/web/src/ferrostar-core.ts
@@ -41,6 +41,7 @@ export class FerrostarCore extends LitElement {
 
   routeAdapter: RouteAdapter | null = null;
   map: maplibregl.Map | null = null;
+  searchBox: MapLibreSearchControl | null = null;
   navigationController: NavigationController | null = null;
   currentLocationMapMarker: maplibregl.Marker | null = null;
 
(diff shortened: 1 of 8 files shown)


web/src/ferrostar-core.ts
```diff
@@ -113,6 +158,8 @@ export class FerrostarCore extends LitElement {
 
   // TODO: type
   async startNavigation(route: any, config: any) {
+    if (this.useIntegratedSearchBox) this.map?.removeControl(this.searchBox!);
```
ianthetechie, line 161: Good idea making this configurable.

web/src/location.ts
```diff
@@ -61,10 +61,7 @@ export class BrowserLocationProvider {
           courseOverGround: {
             degrees: Math.floor(position.coords.heading || 0),
           },
-          timestamp: {
-            secs_since_epoch: Math.floor(position.timestamp / 1000),
-            nanos_since_epoch: 0,
-          },
+          timestamp: Date.now(),
```
ianthetechie, line 64: This seems like a regression 🤔

web/src/ferrostar-core.ts
```diff
@@ -122,11 +169,7 @@ export class FerrostarCore extends LitElement {
           coordinates: route.geometry[0],
           horizontalAccuracy: 0.0,
           courseOverGround: null,
-          // TODO: find a better way to create the timestamp?
-          timestamp: {
-            secs_since_epoch: Math.floor(Date.now() / 1000),
-            nanos_since_epoch: 0,
-          },
+          timestamp: Date.now(),
```
ianthetechie, line 172: Why do we always use the current timestamp?

web/src/ferrostar-core.ts
```diff
@@ -88,6 +124,15 @@ export class FerrostarCore extends LitElement {
...
       zoom: 18,
     });
+
+    if (this.useIntegratedSearchBox) {
+      this.searchBox = new MapLibreSearchControl({
+        onResultSelected: (feature) => {
+          this.startNavigationFromSearch(feature.geometry.coordinates);
+        },
+      });
+      this.map.addControl(this.searchBox, "bottom-left");
```
ianthetechie, line 134: I'd suggest this going in the top left instead (convention; that's where most such UIs put this).

ianthetechie (commented):
One more general comment: I'm not sure what happened to the CSS for the map control. It normally looks great without any extra work.

Expected:

image

Current (also shows attribution even without results; that's unexpected):

image

CatMe0w:
image
CSS fixed, and I moved all the text and buttons to the top since leaving the simulate navigation button on the bottom while the search box is on the top might be not intuitive. Is that ok?

PR #181, opened 2024-08-12
Initial draft of adding TypeScript types

CatMe0w:
This draft PR is based on the other PR I have now.

This draft is incomplete and and further discussion is needed.

In the latest commit of this PR/branch, the crate `ts-rs` is used to generate TypeScript type declarations from serde macros. Upstream git repo is used instead of the crates.io releases due to some features not yet available in the releases. (see https://github.com/Aleph-Alpha/ts-rs/pull/316)

The generated type declarations are not yet well integrated with wasm-pack artifacts, and function types still use `any`. Additionally, to generate the type declarations, you need to run `cargo test` before running `wasm-pack`.


ianthetechie:
[tsify](https://github.com/madonoharu/tsify) might be another option to look into.

CatMe0w:
Force pushed to use the latest main branch HEAD as a base. The originial edits of this PR remain unchanged.

ianthetechie:
Thanks! I'l ltry to get back to this one within the next week :)

ianthetechie:
I'm marking this ready for review, as I think I've got it to the point where it seems to 1) work, and 2) is producing verifiably useful improvements!

There are definitely still some more TODOs in terms of type mapping to TS, but as web isn't a core focus area for any of the core devs, I think we can merge this sooner to start getting the benefits and improve TS defs gradually over time (I can open an issue if you guys agree).

ianthetechie (approved):

PR #184, opened 2024-08-13
Clear the TTS queue when stopping navigation

### LINKED ISSUE #117: Clear the text-to-speech queue when starting a new navigation session
When recalculating in a somewhat rapid loop (ex: driving the wrong direction but routing keeps telling you to start going the opposite way), the text-to-speech queue can pile up as the instructions may take longer to read than the recalculation throttle delay. Once you're going the right direction again, you may still keep getting the same voice prompt.

I believe the issue occurs because the TTS queue isn't cleared. Normally this behavior is correct, but it really should drop the queue and stop playback of the current utterance whenever `startNavigaton` is invoked.

I noticed this bug on Android. It might affect iOS as well, so worth a check.

ianthetechie:
This closes #117 since `stopNavigation` is called as part of the rerouting process.

### DIFF
--- a/android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
+++ b/android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
@@ -297,6 +297,7 @@ class FerrostarCore(
     _navigationController = null
     _state.value = NavigationState()
     _queuedUtteranceIds.clear()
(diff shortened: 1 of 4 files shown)


ianthetechie:
I can't reproduce the strange behavior I noticed previously; maybe it was a fluke. I think we can probably just merge this after a review and do a point release so that people can test it and report issues from real-world use.

Archdoog (approved):

PR #189, opened 2024-08-23
Publish web SDK

### LINKED ISSUE #193: Documentation (tutorials etc.) for web users
Refer to the other tutorials in the mbdook guide for examples of the correct style.

CatMe0w:
Finally, here it is.

Note that the guide is still very preliminary and needs polishing!

Tested and working, although CI and unit tests are missing. Automated testing still needs more work because WebAssembly and web components is not compatible enough with Vitest/Jest, and the introduction of Vitest browser mode introduces more complexity so I want to avoid using browser mode as much as possible.

Closes #193


ianthetechie (commented):
Mostly looks good; I just have a few nitpicks about documentation and found a few omissions going through it.

I'll push my changes and merge though so we can get this out ASAP.

guide/src/web-getting-started.md
```diff
@@ -0,0 +1,162 @@
...
+```html
+<ferrostar-core
+  id="core"
+  valhallaEndpointUrl="https://valhalla1.openstreetmap.de/route"
+  styleUrl="https://tiles.stadiamaps.com/styles/outdoors.json?api_key=YOUR_API_KEY"
+  profile="bicycle"
+></ferrostar-core>
+```
+
+NOTE: `<ferrostar-core>` requires setting CSS manually or it will be invisible.
```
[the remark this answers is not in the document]
CatMe0w: I think this is an intended behavior, but it's fine to create an issue to track this.

web/package.json
```diff
@@ -3,10 +3,20 @@
...
+  },
+  "files": [
+    "dist"
+  ],
   "scripts": {
     "dev": "vite",
     "build": "tsc && vite build",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "test": "vitest",
```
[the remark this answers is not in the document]
CatMe0w: Oops. This line should not be included in this PR, but adding Vitest into dependencies is also okay to resolve this.

guide/src/web-getting-started.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Configuration explained
+
+`<ferrostar-core>` provides a few properties to configure.
+Here are the most important ones:
+
+- `valhallaEndpointUrl`: The Valhalla routing endpoint to use. You can use any reasonably up-to-date Valhalla server, including your own. See [vendors](./vendor.md#routing) for a list of known compatible vendors.
+- `httpClient`: You can set your own fetch-compatible HTTP client to make requests to the routing API (ex: Valhalla).
+- `costingOptions`: You can set the costing options for the route provider (ex: Valhalla JSON options).
+- `useIntegratedSearchBox`: Ferrostar web includes a search box powered by Stadia Maps, but you can disable this and replace with your own.
```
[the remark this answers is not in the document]
CatMe0w: Done.

guide/src/web-getting-started.md
```diff
@@ -60,100 +78,28 @@ ferrostar-core {
...
 
-### Configure the Ferrostar Core
-
-(TODO)
-Here's an example:
-
-```javascript
-const config = {
```
CatMe0w: this `config` object is required for `core.startNavigation`, but I'm not sure how to document it correctly. Also, should we include this as a default value for `core.startNavigation` in our web component code?

guide/src/web-getting-started.md
```diff
@@ -0,0 +1,112 @@
...
+`<ferrostar-core>` provides a few properties to configure.
+Here are the most important ones:
+
+- `valhallaEndpointUrl`: The Valhalla routing endpoint to use. You can use any reasonably up-to-date Valhalla server, including your own. See [vendors](./vendor.md#routing) for a list of known compatible vendors.
+- `httpClient`: You can set your own fetch-compatible HTTP client to make requests to the routing API (ex: Valhalla).
+- `costingOptions`: You can set the costing options for the route provider (ex: Valhalla JSON options).
+- `useIntegratedSearchBox`: Ferrostar web includes a search box powered by Stadia Maps, but you can disable this and replace with your own.
+- `useVoiceGuidance`: Enable or disable voice guidance.
+
+NOTE: `useIntegratedSearchBox` and `useVoiceGuidance` are disabled by default. Set them to any value to enable them.
```
CatMe0w, line 95: This is a [limitation](https://lit.dev/docs/components/properties/#boolean-attributes) of Lit.

ianthetechie: Makes sense; thanks for linking that + clarifying in the docs!

guide/src/web-getting-started.md
ianthetechie: > This config object is required for core.startNavigation

Actually we made a change very recently which moves the config requirement to the `FerrostarCore` constructor on other platforms ([Android example](https://github.com/stadiamaps/ferrostar/blob/2e6dfbd7e95a65e3fc18406e21f8d9c402be7077/android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt#L62)). I think doing the same on the web would be reasonable too. Most of the time you want to just set a config once. This makes it an _optional_ override when calling `startNavigation`.

> Also, should we include this as a default value for core.startNavigation in our web component code?

Yeah, the above addresses this, but it should be an optional override with a default set on the `FerrostarCore` component. Somewhat related: the `startNavigationFromSearch` internal function currently just uses hard-coded values; storing the config at the component level can fix this.

Re: documentation, I think it's good enough for now as the most common path is clear. We can improve customization docs as we refine the design.

PR #190, opened 2024-08-24
Add voice guidance for web

CatMe0w:


web/src/ferrostar-core.ts
```diff
@@ -159,11 +166,14 @@ export class FerrostarCore extends LitElement {
 
   // TODO: type
   async startNavigation(route: any, config: any) {
+    // Remove the search box when navigation starts
```
ianthetechie, line 169: Nice touch!

ianthetechie (approved):
This is fantastic! Really nice to see this in before the end of GSoC 😍 

And thanks for the added comments as well. Those make reading much smoother.

PR #197, opened 2024-08-25
Fix web demo app building

CatMe0w:
This may fix the HTTP 404 error in the [hosted demo](https://stadiamaps.github.io/ferrostar/web-demo).

See also: https://github.com/vitejs/vite/discussions/8222


ianthetechie (approved):
Looks about right locally, so all we can do is try :D

PR #200, opened 2024-08-28
feat: keep screen alive convenience for android

Archdoog:
- Adds keep screen alive disposable effect (and adds it to demo app)
- Revises hide system ui module path (file name mismatch).

### DIFF
--- a/android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/runtime/KeepScreenOnDisposableEffect.kt
+++ b/android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/runtime/KeepScreenOnDisposableEffect.kt
@@ -1,0 +1,40 @@
+package com.stadiamaps.ferrostar.composeui.runtime
+
+import android.app.Activity
+import android.view.Window
+import android.view.WindowManager
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.ui.platform.LocalContext
+import androidx.core.view.WindowCompat
+import androidx.core.view.WindowInsetsControllerCompat
+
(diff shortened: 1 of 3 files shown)


android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/runtime/KeepScreenOnDisposableEffect.kt
```diff
@@ -0,0 +1,40 @@
...
+
+/** Get the WindowInsetsController for the provided window. */
+@Composable
+fun windowInsetsController(window: Window): WindowInsetsControllerCompat {
+  return WindowCompat.getInsetsController(window, window.decorView)
+}
+
+/**
+ * A Composable that keeps the screen on while the hosting Composable is in the view hierarchy.
+ * OnDispose, the flag is cleared and the screen will return to its normal behavior.
```
ianthetechie, line 27:
```suggestion
 * On dispose, the flag is cleared and the screen will return to its normal behavior.
```

android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoNavigationScene.kt
```diff
@@ -39,6 +40,10 @@ fun DemoNavigationScene(
     locationProvider: SimulatedLocationProvider = AppModule.locationProvider,
     core: FerrostarCore = AppModule.ferrostarCore
 ) {
+  // Keeps the screen at consistent brightness on while this Composable is in the view hierarchy.
```
ianthetechie, line 43:
```suggestion
  // Keeps the screen on at a consistent brightness while this Composable is in the view hierarchy.
```

PR #214, opened 2024-08-31
Added nearest coordinate index for polyline

Archdoog:
- Adds index of polyline coordinate calculator in preparation for current attributes and snap bearing/course.

### DIFF
--- a/common/ferrostar/src/algorithms.rs
+++ b/common/ferrostar/src/algorithms.rs
@@ -5,10 +5,8 @@ use crate::{
     navigation_controller::models::TripProgress,
 };
 use geo::{
-    Closest, ClosestPoint, EuclideanDistance, HaversineDistance, HaversineLength, LineLocatePoint,
-    LineString, Point,
+    Closest, ClosestPoint, Coord, EuclideanDistance, HaversineDistance, HaversineLength, LineLocatePoint, LineString, Point
 };
-
 use crate::navigation_controller::models::{
     StepAdvanceMode, StepAdvanceStatus,
     StepAdvanceStatus::{Advanced, EndOfRoute},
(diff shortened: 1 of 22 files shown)


common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -100,6 +105,12 @@ impl NavigationController {
                         step: current_step,
                         linestring,
                     } => {
+                        let current_geometry_index = index_of_closest_origin_point(
+                            *snapped_user_location,
+                            &self.route.get_linestring(),
```
ianthetechie, line 110: I think this code is wrong? Why do we look at the entire route linestring here? This should always be over the current step IIUC.

common/ferrostar/src/algorithms.rs
```diff
@@ -27,6 +26,36 @@ use std::time::SystemTime;
...
+    let point = Point::from(location.coordinates);
+    let skip_index = skip_to_index as usize;
+
+    line.lines()
+        .enumerate()
+        .skip(skip_index)
+        .min_by(|(_, line1), (_, line2)| {
+            let dist1 = line1.euclidean_distance(&point);
+            let dist2 = line2.euclidean_distance(&point);
+            dist1.partial_cmp(&dist2).unwrap()
```
ianthetechie, line 49: I think we can get rid of this `unwrap` with a `total_cmp`.

common/ferrostar/src/algorithms.rs
```diff
@@ -27,6 +26,36 @@ use std::time::SystemTime;
 #[cfg(all(test, feature = "web-time"))]
 use web_time::SystemTime;
 
+/// Get the index of the closest point in the line.
```
ianthetechie, line 29: This docstring needs a bit more detail ;) In particular, it's not clear what `skip_to_index` does and why you'd use it.

It must also document edge cases pending resolution.

ianthetechie (commented):
Ran out of time this morning but here's a partial review.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -108,7 +108,7 @@ impl NavigationController {
                         let current_geometry_index = index_of_closest_origin_point(
                             *snapped_user_location,
                             &self.route.get_linestring(),
-                            *current_geometry_index,
+                            current_geometry_index.unwrap_or(0),
```
ianthetechie, line 111: Do we always want to fall back to zero? Given the fact that we can "skip" forward and it's impossible to go backward, is it possible that we could lock to the final index, skipping the entire array?

I am still not 100% sure if this is the behavior we want (you've probably thought it through better than me, so maybe you can explain if I'm missing something?), but if it is, we need to consider the ramifications and probably have the function either return a Result with a descriptive Error type or an enum describing the different states.

If we don't do this, then I _think_ the result should always be Some, so we should log what happened (at a minimum) so we can detect bugs.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -205,7 +205,7 @@ impl NavigationController {
                 let current_geometry_index = index_of_closest_origin_point(
                     snapped_user_location,
                     &self.route.get_linestring(),
-                    *current_geometry_index,
+                    current_geometry_index.unwrap_or(0),
```
ianthetechie, line 208: Same as previous comment re: `unwrap_or`.

common/ferrostar/src/navigation_controller/mod.rs
ianthetechie: (To clarify why I think it's wrong, we usually look only at the current step both for efficiency and to limit weird step jumps like you or someone else mentioned is possible in Mapbox code. Other code calling the `index_of_closest_origin_point` function also limits to the current step.)

ianthetechie:
Discussed on Slack and decided to go with checks over the current route step for now, with the current index localized to the step. We will currently allow "backward" progress within a step, but we can always revisit this decision (ex: if it gets too jittery around boundaries).

PR #219, opened 2024-09-05
feat: starting point for course snapping

Archdoog:
- Adds course snapping functionality for user location.


android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
```diff
@@ -70,11 +70,14 @@ class DefaultNavigationViewModel(
           .map { coreState ->
             lastLocation =
                 when (coreState.tripState) {
-                  is TripState.Navigating -> coreState.tripState.snappedUserLocation
+                  is TripState.Navigating -> {
+                    Log.d("NavigationViewModel", "Course is ${lastLocation?.courseOverGround?.degrees} at idx ${coreState.tripState.currentStepGeometryIndex}")
```
Archdoog, line 74: Remove this before merge. It's useful for debugging.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -139,6 +144,11 @@ impl NavigationController {
                             *snapped_user_location,
                             &self.route.get_linestring(),
                         );
+                        let snapped_user_location_with_course = apply_snapped_course(
```
Archdoog, line 147: I was using this in the trip state constructed just below it. But reverted that to see if the memory heap stall went away.

ianthetechie: I believe the memory heap issue was due to the incorrect bounds checking logic I flagged above.

ianthetechie (commented):
I've pushed some fixes which should address the worst issues. I think this should un-block you from continuing :)

One unit test will still fail, but that is to be expected, since we're unconditionally snapping the course (which should be configurable).

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -120,4 +133,5 @@ pub enum StepAdvanceMode {
 pub struct NavigationControllerConfig {
     pub step_advance: StepAdvanceMode,
     pub route_deviation_tracking: RouteDeviationTracking,
+    pub snap_course: SnapCourseTo,
```
ianthetechie, line 136:
```suggestion
pub struct NavigationControllerConfig {
    /// Configures when navigation advances to the next step in the route.
    pub step_advance: StepAdvanceMode,
    /// Configures when the user is deemed to be off course.
    ///
    /// NOTE: This is distinct from the action that is taken.
    /// It is only the determination that the user has deviated from the expected route.
    pub route_deviation_tracking: RouteDeviationTracking,
    /// Configures how the heading component of the snapped location is reported in [`TripState`].
    pub snapped_location_course filtering: CourseFiltering,
```

ianthetechie (commented):
This is super close now! I think all of my remaining feedback is around how to name things ;)

We can also discuss during tomorrow's meeting how this relates to #250. I saw your comments about putting the raw location in `TripState` and while I don't think that's the right move, it's possible that maybe I'm missing some other organization that works better. Both this and the other PR are concerned with filtering/processing location, and so maybe there is something general that we should be organizing both under (paving the way for things like Kalman filtering with a more formal structure than, for example, wrapping it in a location provider). It's tricky because some of these are very specific to a current trip whereas others (ex: raw location, Kalman filtering, other techniques for removing GPS noise, etc.) are more independent.

ianthetechie (approved):

ianthetechie:
Note to self before final merge: verify that this doesn't break anything on the web components; those are not covered as well by automated tests.

ianthetechie:
This is perfect!

I just took a screen recording of driving while the navigation mode was bicycle (to force some off route). It tracked the heading perfectly (even when going the exact wrong direction at the start).

https://github.com/user-attachments/assets/600bf117-594b-4063-8b27-776ed6dc0c6d

PR #241, opened 2024-09-12
Publicly mutable `activityType` for changing modes.

michaelkirk:
If the application routes with one mode, then the user changes modes
later, we can now update the activityType without having to create and
start updating a new CoreLocationManager.


michaelkirk:
Note that https://github.com/stadiamaps/ferrostar/blob/main/CONTRIBUTING.md  mentions a CHANGELOG, but it doesn't look like it exists yet.

ianthetechie:
Yikes; my bad on the CHANGELOG! That was intended, but never happened 😅 Added #243

ianthetechie (approved):

PR #249, opened 2024-09-14
Web cleanup

### LINKED ISSUE #206: Allow "bring your own map" or ensure it's easy to customize everything about ours
This will be a hard requirement to using in almost any non demo app.

### LINKED ISSUE #207: Add MapLibre geolocation control to web
This should probably default to following user location for most applications.

ianthetechie:
This is another round of web improvements.

* Adds hooks for configuring the map on first load.
* Automatically adds a MapLibre `GeolocateControl`, which is the UI that most people will want/expect. By default, it will automatically trigger location updates on load, but this can be easily disabled.
* Removes the stand-in marker that tracked the user location *except during simulation* (where there is no alternative; the `GeolocateControl` doesn't have a swappable location engine as on mobile).
* Removes the Stadia Maps autocomplete search control, but retains it in the demo app. This is extremely useful, and adding it directly was the quickest path to getting something actually usable+testable over the summer, but now we can remove the tight linkage thanks to other improvements in this PR. The dev dependency is upgraded to the latest version, which has a lot of visual improvements (see https://github.com/stadiamaps/maplibre-search-box/pull/4)
* Adds hooks for supplying custom CSS (needed due to the way Lit's scoped shadow DOM works in order to properly style things like MapLibre plugins).

Closes #206. Closes #207.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,7 @@ import PackageDescription
 
 let binaryTarget: Target
 let maplibreSwiftUIDSLPackage: Package.Dependency
-let useLocalFramework = true
+let useLocalFramework = false
 let useLocalMapLibreSwiftUIDSL = false
 
 if useLocalFramework {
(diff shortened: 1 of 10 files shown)


Archdoog (approved):
Approved during ferrostar call 8/16/24

Issue #209, opened 2024-08-31
Make display of snapped location configurable


ianthetechie:
Currently all platforms show the user's snapped location only. We should allow showing the raw location as this will be a practical requirement for non automotive navigation.

- [ ] iOS
- [ ] Android
- [ ] Web

ianthetechie:
Here's some background on how things work today and what needs to happen to make snapping configurable.

# Android

The location is read from the `snappedLocation` property of `NavigationUiState`. This happens inside the `NavigationMapView` composable here: https://github.com/stadiamaps/ferrostar/blob/22f408f82bec2127dff89959a25fecb13a44d10f/android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/NavigationMapView.kt#L52.

The `NavigationUiState` is created from the `TripState` generated by the core (Rust). The location comes from here (for the purposes of this issue; you don't need to look at where the location provider comes from): https://github.com/stadiamaps/ferrostar/blob/22f408f82bec2127dff89959a25fecb13a44d10f/android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt#L71.

# iOS

On iOS, things are slightly less thought out (actually a bit surprising! Normally it's reversed!). We don't store any location in the [`NavigationState`](https://github.com/stadiamaps/ferrostar/blob/main/apple/Sources/FerrostarCore/NavigationState.swift) at the top level.

# Ideas for implementation

I think that it would be a mistake to put this in `TripState`, because we want to support scenarios where the user is not actively "navigating" somewhere. Instead, I think that it belongs in the models observed by the UI.

On Android, this is already in the right place (the `snappedLocation` property of `NavigationUiState`). This already uses optional fields. For iOS, I think the right place to put this is in a new _optional_ property on `NavigationState` (maybe we should unify the naming here? Thoughts @Archdoog?). In both cases the new/renamed property should be named `userLocation`. The Android implementation (linked above) will need to change a bit since it's currently switching to whichever is available in a way that the user can't control.

Then, we can extend the composable UI (SwiftUI / Jetpack Compose) components with a boolean property controlling whether to use the snapped location or not, when available (there are a few layers of views; the top level is the dynamically orienting variety, which is what most users will interact with).

cu8code:
@ianthetechie thanks for the clarification. As I was looking into the android code, I had some additional questions. in `uiState.value.snappedLocation` were is the snapping occurring.

[...]

PR #250, opened 2024-09-14
Added Android Support for Make display of snapped location configurab…

cu8code:
#209 

This PR adds the support for making the snapped location configurable by adding an additional variable `location: UserLocation`  that exposes the unsnapped location of the user!

### DIFF
--- a/android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
+++ b/android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
@@ -19,6 +19,7 @@ import uniffi.ferrostar.UserLocation
 import uniffi.ferrostar.VisualInstruction
 
 data class NavigationUiState(
+    val location: UserLocation?,
     val snappedLocation: UserLocation?,
     val heading: Float?,
     val routeGeometry: List<GeographicCoordinate>,
@@ -33,10 +34,12 @@ data class NavigationUiState(
     fun fromFerrostar(
         coreState: NavigationState,
         isMuted: Boolean?,
-        userLocation: UserLocation?
+        location: UserLocation?,
+        snappedLocation: UserLocation?
     ): NavigationUiState =
         NavigationUiState(
-            snappedLocation = userLocation,
+            snappedLocation = snappedLocation,
+            location = location,
             // TODO: Heading/course over ground
             heading = null,
             routeGeometry = coreState.routeGeometry,
@@ -60,22 +63,22 @@ interface NavigationViewModel {
 class DefaultNavigationViewModel(
     private val ferrostarCore: FerrostarCore,
     private val spokenInstructionObserver: SpokenInstructionObserver? = null,
-    locationProvider: LocationProvider
+    private val locationProvider: LocationProvider
 ) : ViewModel(), NavigationViewModel {
 
-  private var lastLocation: UserLocation? = locationProvider.lastLocation
+  private var snappedLocation: UserLocation? = locationProvider.lastLocation
 
   override val uiState =
       ferrostarCore.state
           .map { coreState ->
-            lastLocation =
+              val location = locationProvider.lastLocation
+            snappedLocation =
                 when (coreState.tripState) {
                   is TripState.Navigating -> coreState.tripState.snappedUserLocation
                   is TripState.Complete,
                   TripState.Idle -> locationProvider.lastLocation
                 }
-
-            uiState(coreState, spokenInstructionObserver?.isMuted, lastLocation)
+              uiState(coreState, spokenInstructionObserver?.isMuted, location, snappedLocation)
             // This awkward dance is required because Kotlin doesn't have a way to map over
             // StateFlows
             // without converting to a generic Flow in the process.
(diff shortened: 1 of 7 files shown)


android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
```diff
@@ -60,22 +63,22 @@ interface NavigationViewModel {
 class DefaultNavigationViewModel(
     private val ferrostarCore: FerrostarCore,
     private val spokenInstructionObserver: SpokenInstructionObserver? = null,
-    locationProvider: LocationProvider
+    private val locationProvider: LocationProvider
 ) : ViewModel(), NavigationViewModel {
 
-  private var lastLocation: UserLocation? = locationProvider.lastLocation
+  private var snappedLocation: UserLocation? = locationProvider.lastLocation
```
ianthetechie, line 69: I don't think this is a valid default value. The location here is straight from the provider (unsnapped).

android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
```diff
@@ -60,22 +63,22 @@ interface NavigationViewModel {
...
 
   override val uiState =
       ferrostarCore.state
           .map { coreState ->
-            lastLocation =
+              val location = locationProvider.lastLocation
+            snappedLocation =
                 when (coreState.tripState) {
```
ianthetechie, line 76: I think I didn't specify very well how I envision this being used (my bad!).

I _think_ it would be useful to have a Boolean argument to the default navigation view model which controls whether the snapped location is used (when available).

It would still need to fall back to the unsnapped location when not navigating. Having a booked though to prefer snapped location would allow for as much flexibility as I can think of at the moment. Maybe @Archdoog also has an opinion on the most useful interface.

android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
```diff
@@ -85,7 +88,7 @@ class DefaultNavigationViewModel(
               started = SharingStarted.WhileSubscribed(),
               initialValue =
                   uiState(
-                      ferrostarCore.state.value, spokenInstructionObserver?.isMuted, lastLocation))
+                      ferrostarCore.state.value, spokenInstructionObserver?.isMuted, locationProvider.lastLocation, locationProvider.lastLocation))
```
ianthetechie, line 91: Do we want to use the last location for both cases here? I think should be the unsnapped value and the other could be snapped.

ianthetechie (changes requested):
Thanks for the PR! Few comments.... I may have missed some context though as I'm reviewing on mobile but wanted to give some initial feedback.

Also you can fix the build issues by running kftmt. There are details in CONTRIBUTING.md

android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
Archdoog: Thinking through the goals here, maybe we should:

1. Add a `rawLocation` to the NavigationState in common. This would generally simplify the stack if a dev wants to use it for some custom behavior (whether UI or other). This is easy to accomplish since we already have single location provider under the hood passing through the `NavigationController`.
2. We could enhance this by providing an encapsulated `fun userLocation(userLocationMode: UserLocationMode) -> UserLocation`. Where `UserLocationMode` is an enum of raw location, snapped location, etc. Then the view model could just publish whatever the location is for the desired mode. This behavior/pattern could also be more useful globally for other platforms - not just android.

Side note, this might be a good point to brainstorm feature bloat on `NavigationState`. I think we're consistently going to have the desire to pack more information into that object (or multiple objects that distill navigation info), so we might want to standardize our process a bit. E.g. separate objects or namespaces like `navigationState.location`, `navigationState.instructions`, etc? 

Thoughts?

ianthetechie: I feel like adding this to the `NavigationState` is a bad idea for several reasons (and said as much in this comment: https://github.com/stadiamaps/ferrostar/issues/209#issuecomment-2342570373).

* As you said, it is really bloating the state
* It doesn't even have anything to do with navigation. Snapped location is very much specific to navigation, but the user's "raw" location is not in any way tied to `NavigationState`. We want to use this outside the context of a trip; it's not hard to get the data; we just want to expose it in the observed _UI state_.

ianthetechie: @Archdoog and I discussed this a bit in the core devs meeting today and will play around with it as-is first before making a decision on what changes need to be made (if any; we realized that it might actually be great as-is but that we just need to actually _use_ it first). No action needed on your part right now @cu8code; we'll probably either 1) merge as-is (after fixing lints or w/e), or 2) have a better idea after using it for a bit.

Sorry to hold this up, but I'll revisit this later in the week :)

ianthetechie: Sorry for the delay on this. Yes, it looks like your existing approach is all we need at the view model level :) I've gone ahead and added the conveniences I was looking for at a higher level (the composable views), which is where they belong anyways.

ianthetechie (approved):

PR #254, opened 2024-09-16
Support HTTP GET request generation

### LINKED ISSUE #220: Support HttpGet action in RouteRequest enum
This will enable support for GraphHopper and other routing APIs that expect GET requests.

ianthetechie:
Closes #220.

cc @hactar for visibility (and a review if you're so inclined).


hactar (approved):
Thanks! Changes look fine as far as I can tell with my still limited knowledge.

I also have a question: are any of the changes in this PR auto generated from rust or any of the shell scripts? If yes, it would be great if these files could be marked some way (comment in the first lines of the file - ideally including what command generated the file from which original file?) - as a beginner contributor I might be tempted to manually change something in these files, only to have my changes removed by the file's regeneration.

apple/Sources/UniFFI/ferrostar.swift
```diff
@@ -3092,6 +3092,7 @@ public func FfiConverterTypeRouteDeviationTracking_lower(_ value: RouteDeviation
 
 public enum RouteRequest {
     case httpPost(url: String, headers: [String: String], body: Data)
+    case httpGet(url: String, headers: [String: String])
```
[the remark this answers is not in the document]
ianthetechie: Correct; this is what lets us do it in a consistent manner across platforms. This isn't quite as important when you're bringing your own code (in a `CustomRouteProvider`, which can do literally anything), and I guess it does make things slightly less ergonomic when implementing your own HTTP-based route provider _in platform code_. I think the tradeoff is worth it though, since it means we can add any number of reasonably common providers (ex: GraphHopper) directly in the core. In those cases, we'll expose convenience initializers like we do for Valhalla already and we won't have to write logic to construct HTTP requests and execute them in a bespoke manner for each platform.

ianthetechie:
> are any of the changes in this PR auto generated from rust or any of the shell scripts? 

Yeah, that's a really good question! The only auto-generated file that you might be tempted to change (AFAIK) is `ferrostar.swift`. Regrettably, we _have_ to commit this to source control to get easy SPM integration. We cover how to do iOS binding generation here: https://github.com/stadiamaps/ferrostar/blob/main/CONTRIBUTING.md#ios. I can't think of any other files offhand which are auto-generated but committed to source control, easily accessible (not stuffed in some ignored, obviously marked build folder), etc.

There are a few protections in place around this. First, the UniFFI authors go to pretty great lengths to ensure that incompatible edits cause a fatal error the first time anything from the FFI is used (checksums and such to ensure the file matches the module being imported, and aborting early as any operations with bad bindings are unsound). Second, if you happen to commit such a file and open a PR, CI regenerates bindings automatically anyways (it has to run the build script to build a local XCFramework to test that your code works!), and any working copy changes are automatically committed. I think this might result in a CI failure for forks; for the core devs, we usually use branches on the main repo, so these would get committed to the branch and the PR updated. (We *have* to do this as part of releases too, since iOS needs to build a binary framework that's hosted as part of the release, and that artifact has a checksum that's timestamp- and machine-dependent, so it has to happen on CI and commit the checksum as part of the release process for SPM).

Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 
 let binaryTarget: Target
 let maplibreSwiftUIDSLPackage: Package.Dependency
-let useLocalFramework = false
+let useLocalFramework = true
```
ianthetechie, line 8: That shouldn't be committed ;)

apple/Sources/FerrostarCore/Extensions/RouteRequestExtensions.swift
```diff
@@ -0,0 +1,37 @@
...
+            case let .httpGet(url: requestUrl, headers: headers):
+                guard let url = URL(string: requestUrl) else {
+                    throw FerrostarCoreError.invalidRequestUrl
+                }
+
+                urlRequest = URLRequest(url: url)
+                urlRequest.httpMethod = "GET"
+                for (header, value) in headers {
+                    urlRequest.setValue(value, forHTTPHeaderField: header)
+                }
```
ianthetechie, line 29: See if it's possible to clean this up a bit.

android/core/src/main/java/com/stadiamaps/ferrostar/core/Extensions.kt
```diff
@@ -1,6 +1,28 @@
...
+            .url(url)
+            .post(body.toRequestBody())
+            .apply { headers.map { (name, value) -> header(name, value) } }
+            .build()
+      }
+      is RouteRequest.HttpGet -> {
+        Request.Builder()
+            .url(url)
+            .get()
+            .apply { headers.map { (name, value) -> header(name, value) } }
```
ianthetechie, line 25: This might be something we can clean up.

Archdoog (approved):
Approved during ferrostar call 8/16/24 (pending existing comments)

.github/workflows/ios.yml
```diff
@@ -37,10 +37,13 @@ jobs:
...
       working-directory: common
 
+    - name: Ensure that Package.swift is NOT configured for local development at this stage (easy to forget)
+      run: sed -i '' 's/let useLocalFramework = true/let useLocalFramework = false/' Package.swift
+
     - name: Commit changed files (it is easy to forget about ferrostar.swift)
       uses: stefanzweifel/git-auto-commit-action@v5
       with:
-        file_pattern: 'apple/Sources/UniFFI/*'
+        file_pattern: 'apple/Sources/UniFFI/* Package.swift'
```
ianthetechie, line 46: cc @Archdoog I fixed this for good :P CI will take care of it.

PR #257, opened 2024-09-17
feat: google play services extensions

### LINKED ISSUE #115: Add a FusedLocationProviderClient wrapper
We want to include a `LocationProvider` backed by `FusedLocationProviderClient`, as this is what most commercial apps will want to use. Some care needs to be taken with this to ensure that it is not an issue for open-source projects wanting to publish on F-Droid, so this will probably end up being an _optional_ separate package. We can still use the same gradle workspace, but it needs to be published separately.

Archdoog:
Closes #115


ianthetechie (approved):
Thanks!

PR #258, opened 2024-09-19
[ios] non-notched layout

michaelkirk:
This branch includes several commits. I can break them into separate PR's if you'd prefer, but they are intended to be both well factored (one issue addressed per commit) and related (layout is a bit whack-a-mole, so merging them independently might make some things worse before everything is better).

<table>
  <tr>
    <th style="width: 50%;">Before</th>
    <th style="width: 50%;">After</th>
  </tr>
  <tr>
    <td><strong>sanity check: the notched device layout was fine before.</strong></td>
    <td><strong>regression check: ArrivalView is intentionally shorter, but otherwise the layout looks the same.</strong></td>
  </tr>
  <tr>
    <td>Screenshot</td>
    <td>Screenshot</td>
  </tr>
  <tr>
    <td>Screenshot</td>
    <td>Screenshot</td>
  </tr>
  <tr>
    <td><strong>But non-notched device layout was missing some padding.</strong></td>
    <td><strong>Now non-notched device layout looks reasonable! 🎉</strong></td>
  </tr>
  <tr>
    <td>Screenshot</td>
    <td>Screenshot</td>
  </tr>
  <tr>
    <td>Screenshot</td>
    <td>Screenshot</td>
  </tr>
</table>


apple/Sources/FerrostarMapLibreUI/Views/DynamicallyOrientingNavigationView.swift
```diff
@@ -73,6 +74,22 @@ public struct DynamicallyOrientingNavigationView: View, CustomizableNavigatingIn
...
+//                        // MapView
+//                        Rectangle().fill(.red.opacity(0.5)).ignoresSafeArea()
+//
+//                        // MapView's safeArea
+//                        Rectangle().fill(.yellow.opacity(0.5))
+//
+//                        // Overlay safeArea
+//                        Rectangle().fill(.blue.opacity(0.5)).padding(overlaySafeAreaInsets)
+//                    }
+//                )
```
michaelkirk, line 92: I can remove this if you like. It might help you review to see what it looks like though - see the code/comments for the color key:

Screenshot 2024-09-18 at 20 09 00
Screenshot 2024-09-18 at 20 09 03
Screenshot 2024-09-18 at 20 09 18
Screenshot 2024-09-18 at 20 09 22

apple/Sources/FerrostarMapLibreUI/Extensions/EdgeInsets.swift
```diff
@@ -0,0 +1,14 @@
+import SwiftUICore
+
+extension EdgeInsets {
+    /// Given a view's insets, returns the insets for the child view, such that
+    /// the child's insets, accumulated with the parents insets, will be at least `minimumInset`
+    func collapsedChildInsets(minimumInset: CGFloat) -> EdgeInsets {
```
michaelkirk, line 6: I tried really hard to think of a better name and this is the best I came up with. Suggestions welcome!

apple/Sources/FerrostarMapLibreUI/Extensions/View.swift
```diff
@@ -0,0 +1,13 @@
+import SwiftUICore
+
+extension View {
+    /// Do something reasonable-ish for clients that don't yet support
+    /// safeAreaPadding - in this case, fall back to regular padding.
+    func safeAreaPaddingPolyfill(_ insets: EdgeInsets) -> AnyView {
```
michaelkirk, line 6: I'm not sure if there's existing patterns for polyfill type methods, but since it gets used in a few places, it seemed worth extracting. Otherwise, doing an `#availability` check inline in a view body is annoying because you have to do an assignment.

like:

```swift
var body: some View {
  HStack {
    VStack {
      Text("Foo")
    }.safeAreaPaddingPolyfill(16)
  }
}
```

Has to become something like:
```swift
var body: some View {
  HStack {
    let foo = VStack {
      Text("Foo")
    }
    if #available(iOS 17.0, *) {
      foo.safeAreaPadding(insets)
    } else {
      foo.padding(insets)
    }
  }
}
```

apple/Sources/FerrostarMapLibreUI/Models/NavigationMapViewContentInsetMode.swift
```diff
@@ -12,7 +12,7 @@ public enum NavigationMapViewContentInsetMode {
     /// where the user location should appear toward the bottom of the map.
     ///
     /// This is used to accommodate a top InstructionView
-    case portrait(within: GeometryProxy, verticalPct: CGFloat = 0.75)
+    case portrait(within: GeometryProxy, verticalPct: CGFloat = 0.75, minVertical: CGFloat = 210)
```
michaelkirk, line 15: 75% is too much for a tiny baby screen like my iPhone SE. Here's what it looks like with 7f49ceb reverted:

Screenshot 2024-09-18 at 20 48 59

Note how the puck is almost underneath the ArrivalView. This is exacerbated by the bottom padding that this PR adds to the ArrivalView.

apple/Sources/FerrostarMapLibreUI/Views/Overylays/LandscapeNavigationOverlayView.swift
```diff
@@ -56,7 +56,6 @@ struct LandscapeNavigationOverlayView: View, CustomizableNavigatingInnerGridView
                         distanceFormatter: formatterCollection.distanceFormatter,
                         distanceToNextManeuver: progress.distanceToNextManeuver
                     )
-                    .padding(.top, 16)
```
michaelkirk, line 59: This padding is now the responsibility of the parent.

apple/Sources/FerrostarMapLibreUI/Extensions/View.swift
```diff
@@ -0,0 +1,13 @@
...
+        if #available(iOS 17.0, *) {
+            AnyView(self.safeAreaPadding(insets))
+        } else {
+            AnyView(padding(insets))
+        }
```
michaelkirk, line 11: BTW, as an alternative, we could just use `padding` on all platforms, rather than `safeAreaPadding`, but `safeAreaPadding` is cool when available because you can do things like this:

Screenshot 2024-09-18 at 20 58 37

See how the InstructionView spills out past the safeArea? It's not possible with `padding` (egregious hacks like hardcoded negative padding maybe not withstanding).

Currently we don't have a use case for this, but since the grid is intended to be configurable, I think this gives users more flexibility if they want some kind of edge-to-edge thing in their overlays.

apple/DemoApp/Demo/DemoNavigationView.swift
```diff
@@ -128,6 +128,8 @@ struct DemoNavigationView: View {
                                     }
                                 } label: {
                                     Text("Start Nav")
+                                        .lineLimit(1)
+                                        .minimumScaleFactor(0.5)
```
michaelkirk, line 132: This is unrelated, but I noticed on small devices with slightly scaled up text:

Screenshot 2024-09-18 at 21 18 25

ianthetechie:
Wow 😍 It might take me till early next week to get to this, but this is awesome! Thanks for all the comparison screenshots and thorough comments!

apple/Sources/FerrostarMapLibreUI/Views/LandscapeNavigationView.swift
```diff
@@ -8,9 +8,11 @@ import SwiftUI
 
 /// A landscape orientation navigation view that includes the InstructionsView and ArrivalView on the
 /// leading half of the screen.
-public struct LandscapeNavigationView: View {
```
michaelkirk, line 11: I assume `CustomizableNavigatingInnerGridView` was omitted by accident here - it's present in DynamicallyOrienting and Portrait NavigationView's.

apple/Tests/FerrostarSwiftUITests/Views/__Snapshots__/ArrivalViewTests/testArrivalViewDefaultTheme.1.png
michaelkirk, line 1: This is my first time doing snapshot tests in swift. Pretty cool! 

I feel pretty good about unilaterally decreasing the hight of the arrival view as it exists in the NavigationView with a close button, but I could see where the single line versions (like this one) might benefit from the additional padding.

I have to ask though - is this view actually used in this way anywhere?

apple/Sources/FerrostarMapLibreUI/Extensions/EdgeInsets.swift
michaelkirk: I thought harder and reworked https://github.com/stadiamaps/ferrostar/pull/258/commits/961343f03bbdfc62cf7879f2fc85e67d6f460503 to be more self contained.

apple/Sources/FerrostarMapLibreUI/Models/NavigationMapViewContentInsetMode.swift
Archdoog: **Puck Position**

On android we actually changed this value to 60% for that reason. We also adjusted the zoom based on travel activity. Both of these factors helped quite a lot with the general viewport. See: [NavigationCamera.kt](https://github.com/stadiamaps/ferrostar/blob/5a90b977cd572595f45ac8a1325f7579486aeca5/android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/runtime/NavigationCamera.kt). 

**Padding**

My preference when it comes to padding is to shoot for 16 as this is roughly Apple's iOS default value provided (e.g. if we just used `.padding(.vertical)` it would be 16 or close to it). With modern screen sizes and DPI's I tend to lean toward this default as it encourages slightly larger surfaces.

That said, there will always be different opinions and priorities with styling. Per that, I'd be all for making an "outer" padding part of the theme configuration. Thoughts @ianthetechie?

Archdoog, line 15: Aside from the larger discussion on this topic. If we keep this as is, I think we should:

1. Add a param explanation for `minVertical` versus the percentage. Explaining why you'd adjust this value and how the percentage interacts with it. E.g. at the call site, it could be easy to assume percentage is within the height minus minVertical or vise versa.
2. Consider if we want a default value here. It seems likely we've got variable inputs here with the size of the arrival and instruction view?

Just want to make sure this is easy to reason with for someone implementing a slightly custom UI, etc.

Archdoog (commented):
This is looking great! Thanks for cleaning up some of the loose ends here 😄

Have some discussion points on the padding and puck positioning since it's a general problem for all platforms.

apple/Sources/FerrostarMapLibreUI/Extensions/View.swift
ianthetechie: Looks good to me :) There isn't an existing pattern, partly because we explicitly started with a fairly recent support target (I actually wanted to start with 17 so we could get the `@Observable` macro, but @Archdoog correctly pointed out that this would unnecessarily limit adoption haha).

apple/Sources/FerrostarMapLibreUI/Extensions/View.swift
```diff
@@ -0,0 +1,70 @@
...
+    ///    |   | 8 | 8 |        child view (self)           | 16 | 0 |   |
+    ///    |   |   |   |                                    |    |   |   |
+    ///    |   |   |   +------------------------------------+    |   |   |
+    ///    |   |   |    insets added by this method (Bottom: 0)  |   |   |
+    ///    |   |   +---------------------------------------------+   |   |
+    ///    |   |    `parentGeometry.safeAreaInsets` (Bottom: 20)     |   |
+    ///    |   +-----------------------------------------------------+   |
+    ///    |                                                             |
+    ///    +-------------------------------------------------------------+
+    /// ```
```
ianthetechie, line 26: Boss-level ASCII art :D

apple/Sources/FerrostarMapLibreUI/Extensions/View.swift
```diff
@@ -0,0 +1,70 @@
...
+    func complementSafeAreaInsets(parentGeometry: GeometryProxy, minimumInset: CGFloat = 16) -> some View {
```
ianthetechie, line 27: Perhaps I'm over-engineering this, but my first question after reading the docs was which edge insets the number applied to. I think we can resolve that ambiguity AND make this more flexible by making it accept edge insets rather than a single number.

apple/Sources/FerrostarMapLibreUI/Models/NavigationMapViewContentInsetMode.swift
ianthetechie: Yeah, we had a lot of tweaking to get the puck positioning to an acceptable point on Android. We've done a lot more layout work there (ironic since both of us are very much iOS heavy haha), but the fragmentation is significantly more obvious over there.

This looks tolerable to me on a wide range of phones in testing. Tablets, the puck is maybe a bit high, but it's fine for now. We can always adjust later.

Making it part of the theme makes sense. Just to clarify, I think you're referring to the `NavigationUITheme`, right? In fact, moving it there would theoretically let us make the default implementation dynamic based on device type (either a size heuristic or idiom tablet checks). 

Assuming we're all agreed on that direction, it just struck me REALLY funny that this is a class and not a struct 🤔 Any reason not to make `NavigationUITheme` a struct @Archdoog?

michaelkirk: I've added some documentation and a rename to hopefully clarify this new parameter. This will conflict with a typo also fixed in #260.

Let me know if there's more you'd like me to change. I readily admit that I did this without consulting the Android codebase, so I appreciate the context. Let me know if you'd like me to do anything in particular to converge the implementations.

ianthetechie (approved):
Looks good now! I just went ahead and pushed the final few changes over the line, and ensured additionally that the parameter is exposed upstream now (no magic that users can't customize).

I'll merge assuming CI checks pass.

PR #71, opened 2024-03-25
Feat/visual instruction view

Archdoog:

PR #259, opened 2024-09-20
[ios] InstructionView secondary row uses theme bg color

michaelkirk:
I noticed while working on the InstructionView that the `secondaryStyle.backgroundColor` is ignored, and instead is hardcoded as a transformation of the `primaryStyle.backgroundColor`. 

I think using the `secondaryStyle.backgroundColor` is less surprising, so I've made that change.

Further, while working on that, I noticed that all the render tests were using a test style (`TestingInstructionRowTheme`) rather than the default styles, which has a couple shortcomings:

1. it was using the same style for primary and secondary, giving an unrealistic look at how people will use these views
2. darkMode testing wasn't possible
3. Having a visually diff of the rendered style across the many permutations of view state with every PR "for free" seems super useful

So, I've change these tests to use the default styles rather than the test style. However, I'm assuming there was a good reason for introducing the Test style in the first place, and so if it's still desirable for some reason, we can revisit that part of the change. For context, it looks like it was added as part of the original InstructionView feature in https://github.com/stadiamaps/ferrostar/pull/71


Archdoog (approved):
Look good. Thanks!

ianthetechie (approved):
LGTM; thanks!

Re: the question about using a test theme, I actually don't recall the exact reason for that. Though I do know a lot of _similar_ changes have been made (ex: baking formatters rather than accepting defaults like most end users will) to make the tests reproducible. I can't detect any changes here that break that (and my machine is notoriously configured to clash with a lot of system defaults that will pass CI haha).

PR #261, opened 2024-09-23
Publish reframed Annotations on Current Step

### LINKED ISSUE #89: Telemetry/CurrentAttributes State from Core
In order to drive various UI features from the state, we need to provide a general purpose state object that can contain values like:

1. The current speed limit at the user's location on the route - if relevant/available `Option<SpeedLimit>`
2. The current filtered speed (e.g. once the core is providing filtered/improved user/vehicle speed) `Option<Speed>`
3. Etc. - Please add other use cases for this.

Archdoog:
- Removes hard coded annotation type in favor of BTreeMap for annotations. This gives us the flexibility of any json attribute an API wants to return plus keeps the json keys sorted for consistency and testability (since we're passing this as a json string through the FFI)
- Reshapes the original annotations format of many arrays of a single value into one array with an object of key value pairs (making it easy to pass to a client platform).

Closes #89 

Two follow up tickets are:

1. Parsing the data on the client side is here: https://github.com/stadiamaps/ferrostar/issues/271. 
2. Convert MaxSpeed to something reasonable: https://github.com/stadiamaps/ferrostar/issues/272


android/core/src/main/java/com/stadiamaps/ferrostar/core/extensions/TripStateExtensions.kt
```diff
@@ -41,3 +45,26 @@ fun TripState.deviation() =
...
+/**
+ * Get the route annotations at the user's current location.
+ *
+ * This function will deserialize the annotation bytes provided by the trip
+ * state.
+ *
+ * @param type The json data class that annotations should be deserialized to.
+ * @return The annotation data or null if the trip is not navigating.
+ */
+inline fun <reified T> TripState.annotation(type: Class<T>): T? =
```
Archdoog, line 58: Conceptually, we want a developer to be able to specify their json type for annotations that includes whatever their API returns. This on get method shows that, but we may want to change to a pre-parsed system. Where the annotation object's `type: Class<T>` and json adapter is added as a param to any RouteAdapter. The route adapter is then responsible for pre-parsing the data. Not sure how easy this would be to get out here as the typed value though?

common/ferrostar/src/algorithms.rs
```diff
@@ -487,6 +487,17 @@ pub(crate) fn get_linestring(geometry: &[GeographicCoordinate]) -> LineString {
         .collect()
 }
 
+/// Get a slice of an array that matches the contents of another slice.
+pub fn slice_indexes<T>(array: Vec<T>, slice: Vec<T>) -> Option<(usize, usize)>
```
Archdoog, line 491: Can delete this. We're not using it.

common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -83,20 +87,52 @@ impl Route {
...
+
+                    return leg.steps.iter().map(move |step| {
+                        let step_geometry =
+                            get_coordinates_from_geometry(&step.geometry, polyline_precision)?;
+
+                        // Slice the annotations for the current step. The annotations array represents segments between coordinates 
+                        // and is thus one element shorter than the geometry array. To account for this, we need to subtract 2 from 
+                        // the length of the geometry array on each step (one to account for the 0-based index, and one to account 
+                        // for the length difference).
+                        let step_index_len = step_geometry.len() - 2 as usize;
```
Archdoog, line 119: This could well be the buffer overflow. Need to actually assert `- 2` is correct.

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
```diff
@@ -0,0 +1,83 @@
...
+pub fn convert_max_speed_to_mps(max_speed: MaxSpeed) -> Option<f64> {
+    match max_speed {
+        MaxSpeed::Known { speed, unit } => match unit {
+            MaxSpeedUnits::KilometersPerHour => return Some(speed * 0.27778),
+            MaxSpeedUnits::MilesPerHour => return Some(speed * 0.44704),
+        },
+        #[allow(unused)]
+        MaxSpeed::Unknown { unknown } => return None,
+    }
+}
```
ianthetechie, line 49: I think this can be exposed as an implementation method on the `MaxSpeed` enum itself.

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
```diff
@@ -0,0 +1,83 @@
...
+        println!("Step Indexes: (start_index = {start_index}, end_index={end_index}, annotation.len: {annot_len})");
+
+        let slice = &annotations[start_index..end_index];
+        Some(slice.to_vec())
+    } else {
+        None
+    };
+}
+
+/// Converts all single value annotation vectors into a single vector of values.
```
ianthetechie, line 22:
```suggestion
/// Converts the the OSRM-style annotation object consisting of separate arrays
/// to a single vector of parsed objects (one for each coordinate pair).
```

common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -83,20 +87,52 @@ impl Route {
...
+                        let end_index = start_index + step_index_len;
+
+                        let annotation_slice =
+                            get_annotation_slice(start_index, end_index, annotations.clone());
+
+                        start_index = end_index;
```
ianthetechie, line 125: Do steps overlap the start/end coord? If so, this may need to be `end_index - 1`.

common/ferrostar/src/routing_adapters/osrm/models.rs
```diff
@@ -87,14 +87,40 @@ pub struct Annotation {
...
-    #[allow(dead_code)]
-    max_speed: Vec<MaxSpeed>,
+    pub max_speed: Vec<MaxSpeed>,
+}
+
+/// A flattened OSRM flavored annotation object.
+/// This represents the value for each annotation type and is typically surfaced for each
+/// geometry segment.
+#[derive(Serialize, Debug, Clone, PartialEq)]
+pub struct AnnotationValue {
```
ianthetechie, line 97: This is probably a bit too Valhalla-specific. Here is what "actual" OSRM provides: https://project-osrm.org/docs/v5.5.1/api/#annotation-object.

I think these should _all_ be optional fields, and our fields should be a superset of all the known APIs. (I'd also like to have support for arbitrary keys/values too, but that can perhaps come later). For me, that means we need to additionally look at the API docs for OSRM itself, Mapbox, and probably GraphHopper. Though I can't find any docs on what annotations GraphHopper supports. Maybe @hactar can share some details.

common/ferrostar/src/routing_adapters/osrm/models.rs
```diff
@@ -87,14 +87,40 @@ pub struct Annotation {
...
     /// and Valhalla.
     #[serde(default, rename = "maxspeed")]
-    #[allow(dead_code)]
-    max_speed: Vec<MaxSpeed>,
+    pub max_speed: Vec<MaxSpeed>,
+}
+
+/// A flattened OSRM flavored annotation object.
+/// This represents the value for each annotation type and is typically surfaced for each
+/// geometry segment.
```
ianthetechie, line 95:
```suggestion
/// An extended OSRM set of annotations.
/// This captures attributes between two points of travel in the route geometry.
```

common/ferrostar/src/routing_adapters/osrm/models.rs
```diff
@@ -87,14 +87,40 @@ pub struct Annotation {
...
+#[derive(Serialize, Debug, Clone, PartialEq)]
+pub struct AnnotationValue {
+    pub distance: f64,
+    pub duration: f64,
+    /// The speed limit for this geometry segment in meters per second.
+    /// If the speed limit is unknown, this will be `None`.
+    pub max_speed_mps: Option<f64>,
+    /// The expected travel speed.
+    pub speed_mps: f64,
```
ianthetechie, line 104: Minor edits. I think we can make our APIs clearer than the original ;) Also we can probably use `f32` in more places. It's often more efficient to compute on, and definitely more efficient to store.

I'm also tempted to to something like `type MetersPerSecond = f32` or even a newtype (though those might introduce annoyance serializing over the FFI).

```suggestion
    pub distance: f32,
    pub duration: f32,
    /// The speed limit for this geometry segment in meters per second.
    /// If the speed limit is unknown, this will be `None`.
    pub speed_limit: Option<f32>,
    /// The estimated travel speed, in meters/second.
    pub estimated_speed: f32,
```

ianthetechie (commented):
Here's a first pass over it. Honestly I think this is fine methodology-wise. We can debug the indexing issues tomorrow, but I think the approach is probably fine so you can proceed with refinements and clean-up.

apple/Sources/FerrostarCore/NavigationState.swift
```diff
@@ -20,4 +20,29 @@ public struct NavigationState: Hashable {
         self.routeGeometry = routeGeometry
         self.isCalculatingNewRoute = isCalculatingNewRoute
     }
+    
+    public var currentProgress: TripProgress? {
```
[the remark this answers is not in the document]
Archdoog: Yeah! Adding annotations to the `if let case` finally blew up SwiftUI's ability to type parse 😆. So it was time.

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
```diff
@@ -0,0 +1,84 @@
...
+use itertools::izip;
+
+/// Get's he slice of annotations
+pub(crate) fn get_annotation_slice(
+    start_index: usize,
+    end_index: usize,
+    annotations: Option<Vec<AnnotationValue>>,
+) -> Option<Vec<AnnotationValue>> {
+    return if let Some(annotations) = &annotations {
+        let annot_len = annotations.len();
```
michaelkirk, line 11: nit: You can use `?` unwrapping to propagate optionals, just like you can to propagate results:

```
let annotations = annotations?;
```

Admittedly I haven't dug into this code much, but I'm also wondering if rather than checking for None in this method, it might be cleaner to move the `None` check up a level, and avoid calling this method altogether in that case.

ianthetechie: Yeah, it's probably a map pattern ;) I made some comments about that earlier.

android/core/src/main/java/com/stadiamaps/ferrostar/core/extensions/TripStateExtensions.kt
Archdoog: This is deferred to https://github.com/stadiamaps/ferrostar/issues/271

common/ferrostar/src/models.rs
```diff
@@ -252,12 +252,6 @@ pub struct Route {
     pub steps: Vec<RouteStep>,
 }
 
-impl Route {
-    pub(crate) fn get_linestring(&self) -> LineString {
```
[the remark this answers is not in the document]
Archdoog: It wasn't. Probably something we could standardize somewhere, but figured I'd clean it out for now.

common/ferrostar/src/routing_adapters/algorithms.rs
```diff
@@ -0,0 +1,26 @@
...
+    geometry: &String,
+    polyline_precision: u32,
+) -> Result<Vec<GeographicCoordinate>, ParsingError> {
+    let linestring = decode_polyline(geometry, polyline_precision).map_err(|error| {
+        ParsingError::ParseError {
+            error: error.to_string(),
+        }
+    })?;
+
+    // TODO: Trait for this common pattern?
```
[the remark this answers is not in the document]
Archdoog: This was an old comment from you that moved. So probably low priority for now.

common/ferrostar/src/routing_adapters/osrm/mod.rs
Archdoog: This was a bit tricky to get right because there are some exceptions. But it's good now and the test `parse_valhalla_asserting_annotation_lengths` helps understand + the snapshot tests. In summary:

1. Annotations should never repeat.
2. Each step has one less annotation than coordinate.
3. The last step never has annotations as it's two of the route's last coordinate (duplicate).

common/ferrostar/src/models.rs
```diff
@@ -337,6 +333,15 @@ impl RouteStep {
             distance_to_end_of_step - instruction.trigger_distance_before_maneuver <= 5.0
         })
     }
+
+    /// Get the annotation data at a specific point along the step.
+    ///
+    /// `at_coordinate_index` is the index of the coordinate in the step geometry.
+    pub fn get_current_annotation_json(&self, at_coordinate_index: u64) -> Option<String> {
```
ianthetechie, line 340: Probably something along these lines is a bit clearer. Return type subject to pending other discussion.

```suggestion
    pub fn get_annotation_at_current_index(&self, at_coordinate_index: u64) -> Option<Vec<u8>> {
```

common/ferrostar/src/models.rs
```diff
@@ -454,6 +459,14 @@ pub struct VisualInstruction {
     pub trigger_distance_before_maneuver: f64,
 }
 
+/// A flat annotations string value map that can be used to store arbitrary
```
ianthetechie, line 462: May also need revision if we change types.

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -69,6 +69,9 @@ pub enum TripState {
         ///
         /// Note it is the responsibility of the platform layer to ensure that utterances are not synthesized multiple times. This property simply reports the current spoken instruction.
         spoken_instruction: Option<SpokenInstruction>,
+        /// Annotation data at the current location.
+        /// This is represented as a json formatted byte array to allow for flexible encoding of custom annotations.
+        annotation_json: Option<String>,
```
ianthetechie, line 74: May also change pending discussion.

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
```diff
@@ -0,0 +1,89 @@
+use super::models::AnyAnnotation;
+use crate::models::AnyAnnotationValue;
+use crate::routing_adapters::error::ParsingError;
+use serde_json::Value;
+use std::collections::BTreeMap;
+
+/// Get's the slice of annotations
+///
+/// Throws an [`ParsingError`] if the annotations are not present or the slice is out of bounds.
```
ianthetechie, line 9:
```suggestion
/// Returns a [`ParsingError`] if the annotations are not present or the slice is out of bounds.
```

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
```diff
@@ -0,0 +1,89 @@
...
+    start_index: usize,
+    end_index: usize,
+) -> Result<Vec<AnyAnnotationValue>, ParsingError> {
+    annotations
+        .ok_or(ParsingError::Annotations {
+            error: "No annotations".to_string(),
+        })?
+        .get(start_index..end_index)
+        .ok_or(ParsingError::Annotations {
+            error: "Annotations slice index out of bounds ({start_index}..{end_index})".to_string(),
```
ianthetechie, line 21: Maybe show the size of the vec we're indexing into in the message for better debuggability?

I also feel like these error types probably belong in some other bucket. These aren't parsing errors.

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
```diff
@@ -0,0 +1,89 @@
...
+///
+/// Throws an [`ParsingError`] if the annotations are not present or the slice is out of bounds.
+pub(crate) fn get_annotation_slice(
+    annotations: Option<Vec<AnyAnnotationValue>>,
+    start_index: usize,
+    end_index: usize,
+) -> Result<Vec<AnyAnnotationValue>, ParsingError> {
+    annotations
+        .ok_or(ParsingError::Annotations {
+            error: "No annotations".to_string(),
```
ianthetechie, line 17: It feels like this function should probably either...

1. not be called at all if there no annotations (and the parameter name can change to be non-optional), OR
2. always be called but return an empty array; is having no annotations really an error?

common/ferrostar/src/routing_adapters/error.rs
```diff
@@ -57,6 +57,8 @@ pub enum ParsingError {
     // TODO: Unable to find route and other common errors
     #[cfg_attr(feature = "std", error("Failed to parse route response: {error}."))]
     ParseError { error: String },
+    #[cfg_attr(feature = "std", error("Failed to parse annotations: {error}."))]
+    Annotations { error: String },
```
ianthetechie, line 61: Probably worth a look now that we have two variants revisiting `ParseError`. I think I wrote that and I now think the name may be too vague. Can we make it more specific?

common/ferrostar/src/routing_adapters/algorithms.rs
```diff
@@ -0,0 +1,26 @@
+use polyline::decode_polyline;
+
+use crate::models::GeographicCoordinate;
+
+use super::error::ParsingError;
+
+/// Parse a polyline-encoded geometry string into a list of geographic coordinates.
+/// If the polyline cannot be decoded, a [`ParsingError`] results.
+pub fn get_coordinates_from_geometry(
```
ianthetechie, line 9: Did this move from somewhere like the main `algorithms` module? If so, I think I probably put it there and that it became a bit of a kitchen sink :P

While might belong under the `routing_adapters` module (honestly not sure), it's more of a common utility conversion function than any algorithm.

Doesn't need to happen for this PR since we're still pre-1.0 and can always change it, but there's probably a better name for the module.

common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -1,18 +1,21 @@
 //! Response parsing for OSRM-compatible JSON (including Stadia Maps, Valhalla, Mapbox, etc.).
 
+pub(crate) mod algorithms;
```
ianthetechie, line 3: Similarly, this is more of a utility module than actual algorithms.

Does this need to be `pub(crate)`? Seems like it's more internal but maybe I missed something.

common/ferrostar/src/routing_adapters/osrm/models.rs
```diff
@@ -67,34 +69,44 @@ pub struct RouteLeg {
...
-pub struct MaxSpeed {
-    // TODO
+#[derive(Debug, Clone)]
+#[allow(dead_code)] // TODO: https://github.com/stadiamaps/ferrostar/issues/271
+pub enum MaxSpeed {
+    Unknown { unknown: bool },
+    Known { speed: f64, unit: MaxSpeedUnits },
+}
+
+#[allow(dead_code)] // TODO: https://github.com/stadiamaps/ferrostar/issues/271
```
ianthetechie, line 97: I think it's fine to leave the defaults (I think they are just warnings, right?). They remind us that there is still something unfinished.

Also, do you have any thoughts on using `SpeedLimit` instead of `MaxSpeed`? I think it's more descriptive and not so Valhalla-specific.

Unrelated: I thought for sure there was an "unlimited" speed. Maybe that's just internal to Valhalla but they might report autobahns with a different speed than unknown or a max value.

common/ferrostar/src/routing_adapters/osrm/mod.rs
ianthetechie: This is... a FANTASTIC summary. Could we codify this in a doc comment on the annotations property? (I'm thinking in the _Ferrostar_ generalized route models, not the OSRM ones. That way it's not buried and makes it clear in public docs how any other routing backend integrator could leverage these in a Ferrostar integration as well.)

ianthetechie (commented):
Few more comments... We're really close on this :D

common/ferrostar/src/models.rs
```diff
@@ -12,17 +12,16 @@ use alloc::{string::String, vec::Vec};
...
 #[cfg(all(feature = "std", not(feature = "web-time")))]
 use std::time::SystemTime;
 
 #[cfg(feature = "web-time")]
 use web_time::SystemTime;
 
-#[cfg(any(test, feature = "wasm-bindgen"))]
-use serde::Serialize;
-
+use std::collections::BTreeMap;
```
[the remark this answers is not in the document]
Archdoog: This is primarily driven by snapshot testing. Using a `HashMap` we'd have to rethink how we test all of the route deserialization stuff. It also helps visually compare things when you're looking a route step because the data will maintain it's column (not required but a nice to have).

common/ferrostar/src/routing_adapters/algorithms.rs
Archdoog: Yeah, definitely worth chatting about and maybe doing a follow up ticket on organization. This was exactly the consideration, the root level `algorithms.rs` was going to become pretty crazy.

common/ferrostar/src/routing_adapters/osrm/models.rs
Archdoog: Yeah, just warnings, so we could definitely revert and leave the todo. 

As for naming, I think we can table this one until the serialization of this object ticket is sorted. It's entirely reasonable to assume that we'll parse from `"max_speed"` on valhalla, mapbox type routes then convert to our more router flexible `SpeedLimit` object.

common/ferrostar/src/models.rs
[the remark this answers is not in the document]
ianthetechie: Hmmm.... I feel like data structure selection based on how they look in snapshot tests is a slippery slope. Let's see if we we can find a way to make the test output comparable. I'm a bit surprised that insta doesn't already account for this already TBH as `HashMap` is the go-to data structure for this sort of thing.

common/ferrostar/src/routing_adapters/osrm/models.rs
ianthetechie: Ooh! Sorry, I missed that this is an OSRM-specific module. It's the public one that I think _has_ to be more generally named. Though TBH I think naming the struct `SpeedLimit` here too is far more useful to the average reader than `MaxSpeed`. IT's a fine line though. A comment is enough I think.

common/ferrostar/src/models.rs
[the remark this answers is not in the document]
Archdoog: The issue is we're asserting after converting back to a string (or bytes if desired). So maybe we can restructure the process a bit to snapshot while it's still a hashmap.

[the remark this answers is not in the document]
ianthetechie: Ok; let's add that to the discussion agenda for our next meeting. Probably easier to talk through it live and maybe I'll try live coding a few ideas.

android/core/src/main/java/com/stadiamaps/ferrostar/core/models/AnnotationValue.kt
```diff
@@ -0,0 +1,12 @@
+package com.stadiamaps.ferrostar.core.models
+
+import com.squareup.moshi.Json
+
+data class AnnotationValue(
```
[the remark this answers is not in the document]
Archdoog: This no longer exists. Resolving

android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
```diff
@@ -70,7 +72,12 @@ class DefaultNavigationViewModel(
           .map { coreState ->
             lastLocation =
                 when (coreState.tripState) {
-                  is TripState.Navigating -> coreState.tripState.snappedUserLocation
+                  is TripState.Navigating -> {
+                    Log.d("NavigationViewModel", "current annotations: ${coreState.tripState.annotation(
```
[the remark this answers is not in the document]
Archdoog: This no longer exists. Resolving

android/core/src/main/java/com/stadiamaps/ferrostar/core/models/AnnotationValue.kt
```diff
@@ -0,0 +1,12 @@
...
+  val distance: Double,
+  val duration: Double,
+  @Json(name = "max_speed_mps")
+  val maxSpeedMps: Double?,
+  @Json(name = "speed_mps")
+  val speedMps: Double
```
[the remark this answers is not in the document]
Archdoog: This no longer exists. Resolving

common/ferrostar/src/routing_adapters/osrm/mod.rs
Archdoog: This is now commented in the code that runs the slicing process 👍

common/ferrostar/src/routing_adapters/osrm/models.rs
Archdoog: This no longer exists. Resolving

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
Archdoog: With regards to `ParsingError::Annotations`: The main goal of this error is to say: "Your API returned something out of spec for annotations" (with whatever reasoning). We're treating this different to normal parsing errors though, because we can easily still run a navigation session even if the annotations have to be tossed because they're malformed.

common/ferrostar/src/models.rs
```diff
@@ -337,6 +334,23 @@ impl RouteStep {
...
+
+    /// Get the annotation data at a specific point along the step.
+    ///
+    /// `at_coordinate_index` is the index of the coordinate in the step geometry.
+    pub fn get_current_annotation_json(&self, at_coordinate_index: u64) -> Option<Vec<u8>> {
+        match &self.annotations {
+            None => return None,
+            Some(annotations) => {
+                if at_coordinate_index as usize >= annotations.len() {
+                    panic!("Index {at_coordinate_index} out of bounds for annotations");
```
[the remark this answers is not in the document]
Archdoog: Yep, and this code is now gone. Resolving

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -68,6 +68,11 @@ impl NavigationController {
         let spoken_instruction = current_route_step
             .get_current_spoken_instruction(progress.distance_to_next_maneuver)
             .cloned();
+        let annotation_bytes = if let Some(current_index) = current_step_geometry_index {
+            current_route_step.get_current_annotation_json(current_index)
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
Archdoog: Resolved 👍

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -145,6 +151,12 @@ impl NavigationController {
                         let spoken_instruction = current_step
                             .get_current_spoken_instruction(progress.distance_to_next_maneuver)
                             .cloned();
+                        let annotation_bytes =
+                            if let Some(at_coordinate_index) = current_step_geometry_index {
+                                current_step.get_current_annotation_json(*at_coordinate_index)
+                            } else {
+                                None
+                            };
```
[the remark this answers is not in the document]
Archdoog: Resolved 👍

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -256,6 +272,13 @@ impl NavigationController {
                             .get_current_spoken_instruction(progress.distance_to_next_maneuver)
                             .cloned();
 
+                        let annotation_bytes =
+                            if let Some(at_coordinate_index) = current_step_geometry_index {
+                                current_step.get_current_annotation_json(at_coordinate_index)
+                            } else {
+                                None
+                            };
```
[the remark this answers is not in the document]
Archdoog: Resolved 👍

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
Archdoog: All of these `Some` ifs and matches are now gone 👍

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
```diff
@@ -0,0 +1,83 @@
...
+    start_index: usize,
+    end_index: usize,
+    annotations: Option<Vec<AnnotationValue>>,
+) -> Option<Vec<AnnotationValue>> {
+    return if let Some(annotations) = &annotations {
+        let annot_len = annotations.len();
+
+        println!("Step Indexes: (start_index = {start_index}, end_index={end_index}, annotation.len: {annot_len})");
+
+        let slice = &annotations[start_index..end_index];
```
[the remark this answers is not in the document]
Archdoog: It's long gone 👍

common/ferrostar/src/routing_adapters/osrm/algorithms.rs
Archdoog: This function gone 👍

common/ferrostar/src/routing_adapters/osrm/mod.rs
Archdoog: Fixed & renamed to utilities.

apple/Tests/FerrostarCoreTests/__Snapshots__/ValhallaCoreTests/testValhallaRouteParsing.1.txt
```diff
@@ -80,6 +91,7 @@
         - spokenInstructions: 0 elements
         - visualInstructions: 0 elements
       ▿ RouteStep
+        - annotations: Optional<Array<String>>.none
```
Archdoog, line 94: Manually redacting this until we solve the next ticket of deserializing all of this.

common/ferrostar/src/models.rs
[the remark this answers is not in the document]
Archdoog: Reverted to `HashMap` and added:

1. `sort_maps` insta setting for tests that validate while it's a HashMap. See https://docs.rs/insta/1.40.0/insta/struct.Settings.html
2. Redact the annotations elsewhere.

common/ferrostar/src/models.rs
```diff
@@ -295,6 +289,8 @@ pub struct RouteStep {
     pub visual_instructions: Vec<VisualInstruction>,
     /// A list of prompts to announce (via speech synthesis) at specific points along the step.
     pub spoken_instructions: Vec<SpokenInstruction>,
+    /// A list of json encoded strings representing annotations between each coordinate along the step.
+    pub annotations: Option<Vec<String>>,
```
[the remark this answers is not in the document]
Archdoog: Because this looses the indexes of the `Vec<String>`, we've decided to leave it as a string for now.

ianthetechie (approved):

PR #267, opened 2024-09-25
Android demo api keys

ianthetechie:
This makes the Android demo app a bit more like the iOS one with API key management separated out. Finally found a good way to do this with a Gradle plugin.

### DIFF
--- a/android/README.md
+++ b/android/README.md
@@ -7,3 +7,15 @@ This directory tree contains the Gradle workspace for Ferrostar on Android.
 * `demo-app` - A minimal demonstration app.
 * `google-play-services` - Optional functionality that depends on Google Play Services (like the a fused location client wrapper). This is a separate module so that apps are able to "de-Google" if necessary.
 * `maplibreui` - Map-related user interface components built with MapLibre.
+
+## Running the demo app
+
+To run the demo app, you'll need a Stadia Maps API key
+(free for development and evaluation use; no credit card required; get one at https://client.stadiamaps.com/).
+You can also modify it to work with your preferred maps and routing vendor by editing `AppModule.kt`.
+
+Set your API key in `local.properties` to run the demo app:
+
+```properties
+stadiaApiKey=YOUR-API-KEY
+```
--- a/android/README.md
+++ b/android/README.md
@@ -1,0 +1,9 @@
+# Ferrostar Android
+
+This directory tree contains the Gradle workspace for Ferrostar on Android.
+
+* `composeui` - Jetpack Compose UI elements which are not tightly coupled to any particular map renderer.
+* `core` - The core module is where all the "business logic", location management, and other core functionality lives.
+* `demo-app` - A minimal demonstration app.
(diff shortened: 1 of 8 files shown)


android/README.md
```diff
@@ -0,0 +1,21 @@
+# Ferrostar Android
+
+This directory tree contains the Gradle workspace for Ferrostar on Android.
+
+* `composeui` - Jetpack Compose UI elements which are not tightly coupled to any particular map renderer.
+* `core` - The core module is where all the "business logic", location management, and other core functionality lives.
+* `demo-app` - A minimal demonstration app.
+* `google-play-services` - Optional functionality that depends on Google Play Services (like the a fused location client wrapper). This is a separate module so that apps are able to "de-Google" if necessary.
```
Archdoog, line 8:
```suggestion
* `google-play-services` - Optional functionality that depends on Google Play Services (like a fused location client wrapper). This is a separate module so that apps are able to "de-Google" if necessary.
```

android/README.md
```diff
@@ -0,0 +1,21 @@
...
+* `maplibreui` - Map-related user interface components built with MapLibre.
+
+## Running the demo app
+
+To run the demo app, you'll need a Stadia Maps API key
```
Archdoog, line 13: Maybe we should create a fallback demo app ticket for iOS & Android. Just so devs can use the simulated route w/ demo tiles without having to sign up to stadia and get a key. Obviously this would be reduced demo functionality. While it's not a huge hurdle, it would keep the minimum requirements to contribute a little lower.

android/demo-app/src/main/java/com/stadiamaps/ferrostar/AppModule.kt
```diff
@@ -59,7 +77,7 @@ object AppModule {
                 NavigationControllerConfig(
                     StepAdvanceMode.RelativeLineStringDistance(
                         minimumHorizontalAccuracy = 25U, automaticAdvanceDistance = 10U),
-                    RouteDeviationTracking.StaticThreshold(25U, 10.0),
+                    RouteDeviationTracking.StaticThreshold(15U, 25.0),
```
Archdoog, line 80: Good reminder here to also write up ideal route deviation setups on iOS & Android. I went back to iOS for the first time in probably months and was appalled about how bad my setting was versus how well android was working 🤣

Archdoog (commented):
Just some minor things. Could easily be deferred to future ticket(s).

android/README.md
ianthetechie: Agreed. Captured it #270.

android/demo-app/src/main/java/com/stadiamaps/ferrostar/AppModule.kt
ianthetechie: Oh, interesting! Actually yeah I don't know since I have my own iOS demo app now haha. It's also tricky because on iOS you have pretty standard hardware with reasonably good geolocation services, whereas on Android, well, I'm intentionally testing on something with very bad geolocation ;) 

These few PRs as well as the not-yet-PR'd `autocomplete-search` branch will make the demo app _actually_ usable so that we can all improve the defaults rather than hacking up stuff in our own separate apps.

PR #268, opened 2024-09-25
Support arbitrary Valhalla options (formerly we only merged costing opts)

ianthetechie:
This is required to do a surprising number of things like change the units to miles, which completely slipped my mind 😅

### DIFF
--- a/common/ferrostar/src/routing_adapters/valhalla.rs
+++ b/common/ferrostar/src/routing_adapters/valhalla.rs
@@ -84,6 +84,8 @@ impl ValhallaHttpRequestGenerator {
     /// Options in this constructor are a JSON fragment representing any
     /// options you want to add along with the request.
     ///
+    /// # Examples
+    ///
     /// ```
     /// # use ferrostar::routing_adapters::valhalla::ValhallaHttpRequestGenerator;
     /// let options = r#"{
@@ -92,7 +94,7 @@ impl ValhallaHttpRequestGenerator {
     ///             "vehicle_type": "golf_cart"
     ///         }
     ///     }
-    /// }"#.to_string();
+    /// }"#;
     ///
     /// // Without options
     /// let request_generator_no_opts = ValhallaHttpRequestGenerator::with_options_json(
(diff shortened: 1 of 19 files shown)


common/ferrostar/src/routing_adapters/valhalla.rs
```diff
@@ -114,8 +136,14 @@ impl RouteRequestGenerator for ValhallaHttpRequestGenerator {
...
                 "costing": &self.profile,
                 "locations": locations,
-                "costing_options": &self.costing_options,
             });
+
+            if let Some(options) = self.options.as_object() {
+                for (k, v) in options.iter() {
+                    args[k] = v.clone();
+                }
+            }
```
ianthetechie, line 145: This is the main change. Pretty much everything else is just renaming.

common/ferrostar/src/routing_adapters/valhalla.rs
```diff
@@ -30,32 +43,41 @@ pub struct ValhallaHttpRequestGenerator {
...
             profile,
-            costing_options: costing_options.unwrap_or(json!({})),
+            options: options.unwrap_or(json!({})),
         }
     }
 
-    pub fn with_costing_options_json(
+    /// Creates a new Valhalla request generator given an endpoint URL, a profile name,
+    /// and options (which will update the minimal defaults set internally).
+    /// NOTE: If options is `None` or a non-object, it will be interpreted as an empty object.
```
Archdoog, line 67: "it will be interpreted as an empty object." I would suggest we describe how this impacts the minimal defaults set internally. This phrasing makes it hard to understand how `None` impacts the internal defaults. E.g is None just the defaults or is None actually no options at all?

Archdoog (commented):
One minor comment. Otherwise looks good 👍

common/ferrostar/src/routing_adapters/valhalla.rs
ianthetechie: Excellent point. I just pushed  a revision which I think clarifies things (there is no effect; it's just ignored).

ianthetechie: Now that you mention in though.... maybe we can do better... and make it fail in that case since that's a bit surprising. What we had before wasn't very obvious either.

ianthetechie: Ok; that's better. Should be all cleaned up. It isn't something that needs explaining now ;) The constructors now either make illegal state unrepresentable or return result types.

PR #276, opened 2024-09-27
[ios] pull down InstructionView handle to show remaining steps

### LINKED ISSUE #32: [iOS] Standard view of upcoming maneuvers
While not all apps will necessarily _need_ this, I expect that most will want a predesigned view that the user can activate to get a list of the upcoming maneuvers along the route.

michaelkirk:
FIXES #32

https://github.com/user-attachments/assets/ef3f3f2d-921e-457a-8307-73dff466dce5

https://github.com/user-attachments/assets/002a5665-e1d6-4801-bbe0-af6a48c4afde


apple/Sources/FerrostarCore/Mock/TestFixtureFactory.swift
```diff
@@ -0,0 +1,92 @@
+import FerrostarCoreFFI
+import Foundation
+
+public protocol TestFixtureFactory {
```
michaelkirk, line 4: I might have gone overboard here, but I expect it to pay dividends to make it easier to create scaffolding data for our previews and unit tests. If you think it's too much, I can inline something less clever for my specific use case.

apple/Sources/FerrostarMapLibreUI/Views/Overylays/PortraitNavigationOverlayView.swift
```diff
@@ -45,44 +47,58 @@ struct PortraitNavigationOverlayView: View, CustomizableNavigatingInnerGridView
     }
 
     var body: some View {
-        VStack {
-            if case let .navigating(_, _, _, _, progress: progress, _, visualInstruction: visualInstruction,
-                                    _) = navigationState?.tripState,
+        ZStack(alignment: .top) {
```
michaelkirk, line 50: The layout changes here are non-trivial.

We want to be able to expand the instruction view *over top* of the other content when it's expanded. But when it's not expanded, we want it to behave as if it's in a vstack with the NavigationInnerGridView.

My solution was to put the InstructionView in a ZStack above the grid, and then plumb out the InstructionView's "unexpanded height" to push down the NavigatingInnerGridView.

I struggled a bit with implementing this layout in SwiftUI. It's something I think would be natural to express in old school iOS layout constraints. Maybe there's a better way?

apple/Sources/FerrostarSwiftUI/Views/InstructionsView.swift
```diff
@@ -24,72 +28,106 @@ public struct InstructionsView: View {
...
+    ///   - sizeWhenNotExpanded: Size of the InstructionsView when minimized - useful for allocating space for this view
+    /// in your layout.
     public init(
         visualInstruction: VisualInstruction,
         distanceFormatter: Formatter = DefaultFormatters.distanceFormatter,
         distanceToNextManeuver: CLLocationDistance? = nil,
+        remainingSteps: [RouteStep]? = nil,
         primaryRowTheme: InstructionRowTheme = DefaultInstructionRowTheme(),
         secondaryRowTheme: InstructionRowTheme = DefaultSecondaryInstructionRowTheme(),
-        showPillControl: Bool = false
```
michaelkirk, line 36: Rather than a boolean, the pill is now shown whenever there are "next steps".

apple/Sources/FerrostarSwiftUI/Views/Maneuver/DefaultIconographyManeuverInstructionView.swift
```diff
@@ -54,6 +54,8 @@ public struct DefaultIconographyManeuverInstructionView: View {
                     maneuverModifier: maneuverModifier
                 )
                 .frame(maxWidth: 48)
+                // REVIEW: without this, the first image in the vstack was rendering very small. Curiously subsequent items in the vstack looked reasonable.
```
michaelkirk, line 57: At the end of the day I don't think it's dangerous to include this change, but it's curious.

Here's what it looks like with this change reverted:

https://github.com/user-attachments/assets/cf2b9280-fe2e-4971-803f-621e3b8b4d4d

michaelkirk:
Converting to a draft, I'd like to tweak the handle styling slightly and add a snapshot test for the expanded maneuver list.

michaelkirk:
Ok! I applied a couple last style tweaks, and updated the demo videos in the description.

apple/Sources/FerrostarCore/Mock/TestFixtureFactory.swift
ianthetechie: Oh boy 😂 I have an allergic reaction to seeing the term "factory" in a Swift codebase, but I'll try to come up with some actually constructive feedback after I've had some time to look at the rest of the changes hahaha.

michaelkirk: Would a factory by any other name smell as bad?

ianthetechie: I'm mostly joking; PTSD from working in the JVM ecosystem haha. This describes what it does, and it's test code, not the public API, so I'm fine with it haha.

apple/Sources/FerrostarMapLibreUI/Views/Overylays/PortraitNavigationOverlayView.swift
ianthetechie: Yeah, I'm not aware of a better way offhand either. It's one of the areas where SwiftUI and others struggle compared to springs-and-struts or auto layout.

apple/Sources/FerrostarCore/Mock/TestFixtureFactory.swift
ianthetechie: Edit: Can this be internal? For tests, we can use testable imports.

ianthetechie: Figured out a resolution for this. It's used only for previews, so we can just move this to the `FerrostarSwiftUI` module and make everything in the file internal :)

apple/Sources/FerrostarSwiftUI/Views/InstructionsView.swift
ianthetechie: I think it still needs to be configurable but `showPillControl` was the wrong thing to toggle ;) I'll play with a few ideas and commit something.

apple/Sources/FerrostarSwiftUI/Views/InstructionsView.swift
```diff
@@ -24,72 +32,110 @@ public struct InstructionsView: View {
     ///   - visualInstruction: The visual instruction to display.
     ///   - distanceFormatter: The formatter which controls distance localization.
     ///   - distanceToNextManeuver: The distance remaining for the step.
+    ///   - remainingSteps: All steps remaining in the route, including the current step
     ///   - primaryRowTheme: The theme for the primary instruction.
     ///   - secondaryRowTheme: The theme for the secondary instruction.
-    ///   - showPillControl: If true, shows a pill control (to indicate an action/expansion).
+    ///   - isExpanded: Whether the instruction view starts expanded
+    ///   - sizeWhenNotExpanded: Size of the InstructionsView when minimized - useful for allocating space for this view
```
ianthetechie, line 39: I wish SwiftUI had a better model for this sort of thing. It's technically a binding, but really its sole purpose is _publishing_ values back out to observers. Maybe we can clarify this a bit in the comment (if I'm understanding it correctly)?

apple/Sources/FerrostarSwiftUI/Views/InstructionsView.swift
ianthetechie: Ok... I feel moderately strongly that passing parameters like this which are then promptly assigned to state is an anti-pattern in SwiftUI, no matter how tempting. (Side note: it's clear that you understand what's happening since your comment explains that this is only the initial value, but I think that puts you in the top like 2% of devs by SwiftUI knowledge :D)

I think that we can sidestep the problem by making this a binding instead of a boolean value. This way developers can still do what you intended, but with less potential confusion and _more_ flexibility:

* You can set your own state to have the default they want
* You can directly control whether it's expanded, or observe whether it is and change UI accordingly
* You can even use `.constant(true)` to force it to stay open if you really want

Ignore my comment on making the "show pill control" toggleable. You elegantly address this in the `TopDrawerView`, and you can always pass an empty set of remaining steps to this view.

apple/Sources/FerrostarMapLibreUI/Views/Overylays/LandscapeNavigationOverlayView.swift
```diff
@@ -11,6 +11,9 @@ struct LandscapeNavigationOverlayView: View, CustomizableNavigatingInnerGridView
 
     private var navigationState: NavigationState?
 
+    @State private var isInstructionViewExpanded: Bool = false
+    // TODO: Same as portrait?
```
ianthetechie, line 15: Did you intend to also make this state + add dynamic padding in landscape like you did for portrait?

apple/Sources/FerrostarMapLibreUI/Views/Overylays/PortraitNavigationOverlayView.swift
ianthetechie: There are a few conflicts unfortunately with the main branch since @Archdoog just cleaned up a bunch of stuff to clean up the pattern matching mess 😅

ianthetechie (commented):
Ok, this is really close now :) I've cleaned up a few things and left some final comments for discussion!

apple/Sources/FerrostarSwiftUI/Views/InstructionsView.swift
michaelkirk: Yes, you are understanding correctly — this is a `Binding` because we want the **child** component to communicate something to its **parent**. I think your additional documentation makes this clearer. 👍

michaelkirk:
> I think that we can sidestep the problem by making this a binding instead of a boolean value. This way developers can still do what you intended, but with less potential confusion and more flexibility:

Funny enough, I actually had it as a binding initially, but got annoyed at not being able to fiddle with the canvas Previews, and decided to change it. 

But really, exposing the "isExpanded" parameter was all just a last minute thing I added to enable snapshot tests, so I'm pretty happy to go whatever way you prefer with it.

> (Side note: it's clear that you understand what's happening since your comment explains that this is only the initial value, but I think that puts you in the top like 2% of devs by SwiftUI knowledge :D)

Thank you, but this is a miscalculation on your part. 😆  I'm new to SwiftUI and any apparent competence is purely coincidental. I'm almost constantly confounded by it.

apple/Sources/FerrostarMapLibreUI/Views/Overylays/LandscapeNavigationOverlayView.swift
michaelkirk: No - I don't think we need to. 

Here's a crude schematic explaining why:

Before this PR, in the portrait layout we used to have:
```
VStack {
  InstructionView
  GridView  // <- grid view is below the instruction view 
  ArrivalView
}
```

To allow the instructionView to expand *over top* of the other content, we now have:
```
ZStack(align: .top) {
  InstructionView
  VStack {
     // !!! We now need to add top padding to account for the InstructionView height
     // otherwise the GridView would be obscured behind the unexpanded instruction view
    GridView.padding(.top, XX_InstructionView.minimizedHeight_XX)
    ArrivalView
  }
}
```

However, things are easier with the landscape view. 
The GridView is in its own column, so it can't be covered by the ArrivalView.

Landscape:
```
HStack {
  VStack {
    InstructionView
    ArrivalView
  }
  // GridView is in its own column, so won't be covered by the InstructionView
  GridView 
}
```

Does that make sense and answer your question?

michaelkirk:
> There are a few conflicts unfortunately with the main branch since @Archdoog just cleaned up a bunch of stuff to clean up the pattern matching mess 😅

I have rebased and pushed. 

If this is not a rebase and force push kind of town, let me know, and I will merge with main next time.

ianthetechie:
Thanks! Giving it a final check now!

> I'm almost constantly confounded by it.

ha! Yeah, SwiftUI is... pretty confounding. I found [Thinking in SwiftUI](https://www.objc.io/books/thinking-in-swiftui/) helpful for making it SLIGHTLY less confounding... (No joke, I read the sections on state management like 25 times over while doing the initial sketches of the MapLibre SwiftUI DSL.)

> If this is not a rebase and force push kind of town, let me know, and I will merge with main next time.

All good; thanks for checking 👍

apple/Sources/FerrostarMapLibreUI/Views/Overylays/LandscapeNavigationOverlayView.swift
ianthetechie: Yes, perfectly! Thanks!

apple/Sources/FerrostarSwiftUI/Views/InstructionsView.swift
```diff
@@ -24,72 +30,111 @@ public struct InstructionsView: View {
...
         self.secondaryRowTheme = secondaryRowTheme
-        self.showPillControl = showPillControl
+        _isExpanded = isExpanded
+        _sizeWhenNotExpanded = sizeWhenNotExpanded
     }
 
-    public var body: some View {
-        VStack {
+    private var nextVisualInstructions: [(VisualInstruction, RouteStep)] {
+        guard let remainingSteps, !remainingSteps.isEmpty else {
```
ianthetechie, line 61: Just noticed a subtle bug here. We try to index from 1 below, so we need to look at the actual count.

apple/Sources/FerrostarSwiftUI/Views/InstructionsView.swift
```diff
@@ -57,8 +57,14 @@ public struct InstructionsView: View {
...
     }
 
+    /// The visual instructions *after* the one for the current maneuver,
+    /// paired with the route step.
+    ///
+    /// Note that in the case of steps with multiple instructions,
+    /// only the first is returned in this list.
+    /// This is sufficient for display of the upcoming maneuver list.
     private var nextVisualInstructions: [(VisualInstruction, RouteStep)] {
-        guard let remainingSteps, !remainingSteps.isEmpty else {
```
michaelkirk: I prefer your change because it's clearer, but I don't see how the previous version was a bug. 

If remaining steps had one thing in it, then `remainingSteps[1...] == []` so the result is the same, right?

Or am I missing something?

ianthetechie: So I went off to code up a Swift Fiddle, because OBVIOUSLY Swift is zero-indexed, and accessing index 1 of an array of length 1 is out of bounds access... 

> You can subscript an array with any integer from zero up to, but not including, the count of the array. Using a negative number or an index equal to or greater than count triggers a runtime error.

But' here's a fiddle that proves your original code actually returns empty array as you claim: https://swiftfiddle.com/aw3i6znzvjcjzi3pdoefmu23sy. I've been writing Swift since the 0.x days and I have to say, this is the most surprised I've been in a very long time! 🤯

Scouring the Apple docs, I couldn't figure out why this works as it does, but your code was indeed correct 😂 If you've got any links explaining why this is the way it is, I'd be grateful to learn!

michaelkirk:
```swift 
let foo = ["a"]
print("foo[0] => \(foo[0])") // "a"
// print("foo[1] => \(foo[1])") // Fatal error: Index out of range

print("foo[0...] => \(foo[0...])") // ["a"]
print("foo[1...] => \(foo[1...])") // []

// And my favorite - while this works (and seems like it should):
print("foo[0...] => \(foo[0...])") // ["a"]
print("foo[0..0] => \(foo[0...0])") // ["a"]

// This blows up
print("foo[1...] => \(foo[1...])") // []
// print("foo[1..1] => \(foo[1...1])") // Fatal error: Array index is out of range

// I think this is a hint to the riddle
// print("foo[2...] => \(foo[2...])") // Swift/Range.swift:754: Fatal error: Range requires lowerBound <= upperBound

// Note the error! 
// It's obvious where the lower bound is coming from (2), 
// but where is the upper bound coming from that is greater than 2?

func example() {
  // So my guess is that accessing a RandomAccessCollection with PartialRangeFrom, 
  // converts the PartialRangeFrom to a Range (not a ClosedRange!), using the Array.length as the `upperBound` of the Range

  do {
    let rangeFrom: PartialRangeFrom = 2...
    let arrayLength = 1
    // this will panic when rangeFrom.start is greater than arrayLength - e.g. 2..<1
    let range: Range = rangeFrom.lowerBound..<arrayLength 
  }

  do {
    // but this will not panic, it'll just be empty:
    let rangeFrom: PartialRangeFrom = 1...
    let arrayLength = 1
    let range: Range = rangeFrom.lowerBound..<arrayLength 
    assert(range == 1..<1)
  }
}
```

Link to fiddle: https://swiftfiddle.com/l4jhe6atjjgkdnih6ot6rsyzgy

PR #294, opened 2024-10-12
IOS Mute button

MarekSabol:
Hi, I've developed it based on our discussion, hopefully you will be satisfied with the implementation :). In case there will be some changes needed, let me know :).

Snímka obrazovky 2024-10-12 o 17 53 04
Snímka obrazovky 2024-10-12 o 17 53 15

### DIFF
--- a/apple/DemoApp/Demo/DemoNavigationView.swift
+++ b/apple/DemoApp/Demo/DemoNavigationView.swift
@@ -13,10 +13,10 @@ private let initialLocation = CLLocation(latitude: 37.332726,
                                          longitude: -122.031790)
 
 struct DemoNavigationView: View {
-    private let navigationDelegate = NavigationDelegate()
+    private var navigationDelegate = NavigationDelegate()
     // NOTE: This is probably not ideal but works for demo purposes.
     // This causes a thread performance checker warning log.
-    private let spokenInstructionObserver = AVSpeechSpokenInstructionObserver(isMuted: false)
+    @State private var spokenInstructionObserver = AVSpeechSpokenInstructionObserver(isMuted: false)
 
     private var locationProvider: LocationProviding
     @ObservedObject private var ferrostarCore: FerrostarCore
@@ -34,6 +34,7 @@ struct DemoNavigationView: View {
 
     @State private var camera: MapViewCamera = .center(initialLocation.coordinate, zoom: 14)
     @State private var snappedCamera = true
+    @State private var isNavigationActive = false // New state variable for navigation status
 
     init() {
         let simulated = SimulatedLocationProvider(location: initialLocation)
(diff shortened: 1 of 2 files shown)


apple/DemoApp/Demo/DemoNavigationView.swift
```diff
@@ -13,10 +13,10 @@ private let initialLocation = CLLocation(latitude: 37.332726,
                                          longitude: -122.031790)
 
 struct DemoNavigationView: View {
-    private let navigationDelegate = NavigationDelegate()
+    private var navigationDelegate = NavigationDelegate()
```
ianthetechie, line 16: Does this need to be a `var`?

apple/DemoApp/Demo/DemoNavigationView.swift
```diff
@@ -148,6 +150,23 @@ struct DemoNavigationView: View {
...
+
+                        MuteButton(isMuted: $spokenInstructionObserver.isMuted)
+
+                    }
+
+                },
+
+                alignment: .topTrailing
+
+            )
```
ianthetechie, line 169: This works for the demo app, but doesn't do much for everyone building apps with the library components :)

Have a look at `PortraitNavigationOverlayView` and `LandscapeNavigationOverlayView` for how we handle this. We already built the whole grid layout and have ways to configure what's shown where, as well as overrides. Moving the mute button to the top trailing position there will also conveniently mean it's shown and hidden automatically in the future based on whether the user is navigating.

apple/DemoApp/Demo/MuteButtonView.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  MuteButton.swift
+//  Ferrostar Demo
+//
+//  Created by Marek Sabol on 08/10/2024.
+//
+
```
ianthetechie, line 7: 1. The mute button isn't specific to the demo app but is reusable, so it should move to the `FerrostarSwiftUI` module. All similar controls are under FerrostarSwiftUI > Views > Controls.
2. We remove the automatic Xcode header comments.

apple/DemoApp/Demo/MuteButtonView.swift
```diff
@@ -0,0 +1,34 @@
...
+                .aspectRatio(contentMode: .fit)
+                .frame(width: 18, height: 18)
+                .padding()
+                .foregroundColor(.black)
+                .background(Color.white)
+                .clipShape(Circle())
+                .shadow(radius: 10)
+        }
+        .padding(.trailing, 18) // Right
+        .padding(.top, 112)
```
ianthetechie, line 28: We can probably remove most of these after integrating into the overlays. See the other controls for examlpes.

MarekSabol:
I've opened new PR for adjusments Ios mute fix after discussion #296

PR #300, opened 2024-10-17
Fix proguard/r8 rules for Android

ahmedre:
Libraries that require proguard rules should publish them automatically to r8 by using consumerProguardFiles. Currently, Ferrostar declared this, but left the corresponding rules files empty. This patch moves the rules from proguard-rules.pro so that R8 automatically picks them up without apps using Ferrostar needing to do anything.

Moreover, it removes proguard-rules.pro and the corresponding release block from Android libraries, since whether the build type is release or debug (and whether or not r8 is enabled) is dependant on the app module.


ahmedre:
How I tested -
- built our app as release - ran it and it crashed.
- applied these changes locally, bumped the version, and did a `./gradlew publishToMavenLocal`.
- updated the app to use the updated snapshot version.
- built the app as release - ran it and it didn't crash.

Younes-Charfaoui (approved):

ianthetechie (approved):
Thanks for the PR! I just did some research on this and I think you're correct. Will merge after CI checks pass.

PR #303, opened 2024-10-21
Support for visual sub-maneuvers

ahmedre:
Within `BannerInstructions`, the MapBox API allows for a `sub_content` for describing sub-maneuvers. This data is used either for showing the step after the current one if it is to happen really quickly after the current step (ex turn left then turn right), or to specify lane information. This PR implements this within Ferrostar.


ahmedre:
relevant urls:
- [Directions API Playground](https://docs.mapbox.com/playground/directions/?coordinates=-74.014218,40.790704&coordinates=-74.057964,40.795581&geometries=polyline6&bannerInstructions=true&voiceInstructions=true&annotations=distance,duration,maxspeed,congestion,speed)
- [Maneuver Instructions Documentation](https://docs.mapbox.com/android/navigation/guides/ui-components/maneuver/#sub-maneuver)

I set this as a draft because I am not sure if this approach is correct, especially considering Valhalla responses (which I didn't think about too much), and about different types of subs, and about the overall model structuring. I figured it was easier to discuss over a PR than just text, so happy to make changes accordingly.

common/ferrostar/src/routing_adapters/osrm/models.rs
```diff
@@ -162,6 +162,8 @@ pub struct BannerInstruction {
     pub distance_along_geometry: f64,
     pub primary: BannerContent,
     pub secondary: Option<BannerContent>,
+    // sub-maneuver information
+    pub sub: Option<SubManeuverBannerContent>,
```
ahmedre, line 166: not sure if using `BannerContent` here is preferable or not, but many of these are different.

common/ferrostar/src/routing_adapters/osrm/models.rs
```diff
@@ -187,6 +189,23 @@ pub struct BannerContent {
...
+pub struct SubManeuverBannerContent {
+    pub text: String,
+    pub components: Vec<SubManeuverBannerContentComponent>,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct SubManeuverBannerContentComponent {
+    #[serde(rename = "type")]
+    pub component_type: Option<String>,
+    pub directions: Option<Vec<String>>,
```
ahmedre, line 202: i only focused on `directions` here, but didn't think too much about other `subs`, like just lists of `type` + `text`, or `type` + `image_base_url`, etc, as per the sample json in this file. should we handle these also?

common/ferrostar/src/models.rs
```diff
@@ -451,6 +451,18 @@ pub enum ManeuverModifier {
...
+/// The content of a visual instruction.
+#[derive(Debug, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(any(feature = "wasm-bindgen", test), derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "wasm-bindgen", serde(rename_all = "camelCase"))]
+#[cfg_attr(feature = "wasm-bindgen", derive(Tsify))]
+#[cfg_attr(feature = "wasm-bindgen", tsify(into_wasm_abi, from_wasm_abi))]
+pub struct LaneInfo {
+    pub active: bool,
+    pub directions: Vec<String>,
```
ianthetechie, line 463: I think `active_direction` might be worth adding too?

ianthetechie (commented):
Wow, thanks for the PR! I think this is pretty much the right direction; just a few comments on models.

I've just reviewed the Rust side of things for now; once we're aligned on data modeling, we can move up to the UI layer (wasn't expecting that in the PR too; thanks!) :D

> Also, I guess I am breaking backward compatibility, but am not sure about what the recommended way to fix this is.

You're talking about semver compatibility, right? Or was there something I missed / didn't review?

Generally speaking, this project is pre-1.0, and a significant number of the organizations actively using it are also engaged in development, so don't worry too much about breaking changes as long as they are announced.

common/ferrostar/src/models.rs
ahmedre: done.

common/ferrostar/src/models.rs
```diff
@@ -451,6 +451,18 @@ pub enum ManeuverModifier {
     SharpLeft,
 }
 
+/// The content of a visual instruction.
+#[derive(Debug, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(any(feature = "wasm-bindgen", test), derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "wasm-bindgen", serde(rename_all = "camelCase"))]
+#[cfg_attr(feature = "wasm-bindgen", derive(Tsify))]
+#[cfg_attr(feature = "wasm-bindgen", tsify(into_wasm_abi, from_wasm_abi))]
```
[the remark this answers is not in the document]
ahmedre: 😂 - done.

common/ferrostar/src/routing_adapters/osrm/models.rs
ianthetechie: 👍 

You know what's funny? I just realized we already have another example from Mapbox's public docs which I had forgotten about (see the `deserialize_banner_instruction` test) 😂 Let's keep both; good to have diversity of API tests in here. We could probably turn one or both of those into snapshot tests now that I think of it.

common/ferrostar/src/routing_adapters/osrm/models.rs
```diff
@@ -185,6 +187,23 @@ pub struct BannerContent {
...
+#[derive(Deserialize, Debug)]
+pub struct BannerContentComponent {
+    #[serde(rename = "type")]
+    pub component_type: Option<String>,
+    pub directions: Option<Vec<String>>,
+    pub active: Option<bool>,
+    pub image_base_url: Option<String>,
+    pub abbr: Option<String>,
+    pub abbr_priority: Option<u8>,
+    pub active_direction: Option<String>,
```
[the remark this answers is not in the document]
ianthetechie: Correct; intersections also have detailed lane info, which is what I originally thought this PR was about 😂 My current understanding (not having designed either API) is that these are complementary, but that the Mapbox usage is the more common one across navigation SDKs. The detailed intersection info may be useful in specific contexts though, so we modeled it just in case.

_typos.toml
```diff
@@ -1,5 +1,10 @@
+[default]
+extend-ignore-re = [
+  "VALHALLA_EXTENDED_OSRM_RESPONSE.*",
```
ianthetechie, line 3: 😂 I... did not expect this action to apply to _raw_ strings. Oh well. Sorry for any annoyance this caused!

common/ferrostar/src/routing_adapters/osrm/models.rs
```diff
@@ -162,6 +162,8 @@ pub struct BannerInstruction {
     pub distance_along_geometry: f64,
     pub primary: BannerContent,
     pub secondary: Option<BannerContent>,
+    // sub-maneuver information
```
ianthetechie, line 165:
```suggestion
    // Sub-maneuver information. This is used to give additional info
    // about the next maneuver or lane guidance.
```

ianthetechie (commented):
Fantastic; thanks!

I also had a quick look over the UI stuff and that all looked fine.

Is this PR still considered a draft? Were you also planning to push SwiftUI+Jetpack Compose views for the sub-instruction layout? (No worries if not; I know you guys also do a lot of custom UI; that can come in a later PR if it's not immediately available. This looks solid to merge as-is.)

ianthetechie:
Looks like a few failures CI caught which are easy to fix (missing params etc.)

ahmedre:
for the UI, happy to try this in another PR - we're still wrapping a `ManeuverView` from maplibre-navigation so would be a good opportunity for us to not depend on it anymore 👍

ianthetechie (approved):
Sounds good on the UI in a separate PR. Thanks! Can merge this as soon as CI passes.

PR #305, opened 2024-10-22
Remove the Gradle Secrets plugin

ianthetechie:
I forgot to put this in the last PR... Halfway through integrating autocomplete search, I found a better way of doing this on Android which didn't require depending on yet another Gradle plugin.

### DIFF
--- a/android/README.md
+++ b/android/README.md
@@ -14,7 +14,8 @@ To run the demo app, you'll need a Stadia Maps API key
 (free for development and evaluation use; no credit card required; get one at https://client.stadiamaps.com/).
 You can also modify it to work with your preferred maps and routing vendor by editing `AppModule.kt`.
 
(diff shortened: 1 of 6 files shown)


Archdoog (commented):
Looks good. Only note is, could we make this optional in this PR and default back to the demo tiles if not set?

I figure this is a good short PR to do that as well. (Happy to merge as is though first)

ianthetechie:
Ahh, good coll. Though at the moment, the search box can't work and we don't have simulated routes back yet.

ianthetechie:
App now works without an API key and doesn't crash ;)

image

Archdoog:
Ah fair. In that case we can just create a ticket

ianthetechie:
Yeah, we know we need to to simulated routes anyways (without recompiling) in the wake of this anyways.

PR #306, opened 2024-10-22
Add a skip.signing parameter for local publishing

ahmedre:
Today, due to the signing requirement, one can no longer publish to Maven local for testing without having credentials. This patch adds an optional flag for disabling signing for cases when someone wants to publish locally for testing.

### DIFF
--- a/android/composeui/build.gradle
+++ b/android/composeui/build.gradle
@@ -68,7 +68,9 @@ dependencies {
 
 mavenPublishing {
     publishToMavenCentral(SonatypeHost.CENTRAL_PORTAL)
-    signAllPublications()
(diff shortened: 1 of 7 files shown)


android/composeui/build.gradle
```diff
@@ -68,7 +68,9 @@ dependencies {
 
 mavenPublishing {
     publishToMavenCentral(SonatypeHost.CENTRAL_PORTAL)
```
ahmedre, line 70: sorry to add more repetition here - at some point we should move this to a convention plugin and just apply it once so everyone gets it automatically.

ianthetechie (approved):

PR #326, opened 2024-10-30
Upcoming instructions android

### LINKED ISSUE #248: [Android] Standard view of upcoming maneuvers
Mirrors #32 but for Android.

ianthetechie:
Closes #248.

### DIFF
--- a/android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
+++ b/android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
@@ -9,8 +9,11 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.material3.Divider
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.MaterialTheme
@@ -56,10 +59,12 @@ fun InstructionsView(
 ) {
     var isExpanded by remember { mutableStateOf(false) }
     val screenHeight = LocalConfiguration.current.screenHeightDp
-    val collapsedHeight = 100  // Seems right... need to check this though
-    val animatedHeight by animateDpAsState(
-        targetValue = if (isExpanded) screenHeight.dp else collapsedHeight.dp
-    )
+    val collapsedHeight = 100.dp
+
+    val targetHeight = if (isExpanded) screenHeight.dp else collapsedHeight
+    val animatedHeight by animateDpAsState(targetValue = targetHeight)
+
+    val scrollState = rememberScrollState()
 
     Box(
         modifier = Modifier
@@ -67,9 +72,12 @@ fun InstructionsView(
             .height(animatedHeight)
             .background(theme.backgroundColor, RoundedCornerShape(10.dp))
             .clickable { isExpanded = !isExpanded }
-            .padding(8.dp)
+            .padding(16.dp)
     ) {
-        Column {
+        Column(
+            modifier = if (isExpanded) Modifier.verticalScroll(scrollState) else Modifier
+        ) {
+            // Primary content
             ManeuverInstructionView(
                 text = instructions.primaryContent.text,
                 distanceFormatter = distanceFormatter,
@@ -78,10 +86,13 @@ fun InstructionsView(
                 content = content
             )
             // TODO: Secondary content
-            if (isExpanded && remainingSteps != null) {
+
+            // Expanded content
+            if (isExpanded && remainingSteps != null && remainingSteps.count() > 1) {
+                Spacer(modifier = Modifier.height(8.dp))
+                HorizontalDivider(thickness = 1.dp)
                 remainingSteps.drop(1).forEach { step ->
                     step.visualInstructions.firstOrNull()?.let { upcomingInstruction ->
-                        HorizontalDivider(thickness = 1.dp)
                         Spacer(modifier = Modifier.height(8.dp))
                         ManeuverInstructionView(
                             text = upcomingInstruction.primaryContent.text,
@@ -91,6 +102,7 @@ fun InstructionsView(
                             content = content
                         )
                         Spacer(modifier = Modifier.height(8.dp))
+                        HorizontalDivider(thickness = 1.dp)
                     }
                 }
             }
--- a/android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
+++ b/android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
@@ -1,15 +1,27 @@
 package com.stadiamaps.ferrostar.composeui.views
 
 import android.icu.util.ULocale
+import androidx.compose.animation.core.animateDpAsState
 import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Divider
+import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.MaterialTheme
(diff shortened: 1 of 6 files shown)


android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
```diff
@@ -36,23 +49,59 @@ fun InstructionsView(
...
-      ManeuverImage(instructions.primaryContent, tint = MaterialTheme.colorScheme.primary)
+    remainingSteps: List<RouteStep>? = null,
+    contentBuilder: @Composable (VisualInstruction) -> Unit = {
+      ManeuverImage(it.primaryContent, tint = MaterialTheme.colorScheme.primary)
     }
 ) {
-  Column(
+  var isExpanded by remember { mutableStateOf(false) }
+  val screenHeight = LocalConfiguration.current.screenHeightDp
+  val collapsedHeight = 100.dp
```
ianthetechie, line 60: See lines 62 and 70. Is there a better way to do this? We need a "default" value it seems, but there's probably a better way.

Archdoog: If we want to get the actual height, this looks likely: https://stackoverflow.com/questions/66752369/get-height-of-element-jetpack-compose. That said, if we can pre-set it to a known value that's actually correct with and without various extra data (lanes, secondary instruction, etc). Then that could be fine. Less work for the UI to do if it's a static value (or set of).

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
```diff
@@ -36,23 +49,60 @@ fun InstructionsView(
...
+                contentBuilder(instructions)
+              }
+
+          // TODO: Secondary content
+
+          // Expanded content
+          if (isExpanded && remainingSteps != null && remainingSteps.count() > 1) {
+            Spacer(modifier = Modifier.height(8.dp))
+            HorizontalDivider(thickness = 1.dp)
+            remainingSteps.drop(1).forEach { step ->
```
Archdoog, line 90: I feel like this should probably be using a [LazyColumn](https://developer.android.com/develop/ui/compose/lists#lazylistscope). Something like?

```kotlin
LazyColumn(
    verticalArrangement = Arrangement.spacedBy(16.dp)
) {
    items(remainingSteps.drop(1)) { step ->
        ManeuverInstructionView(step)
    }
}
```

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
```diff
@@ -36,23 +49,60 @@ fun InstructionsView(
...
+                    text = upcomingInstruction.primaryContent.text,
+                    distanceFormatter = distanceFormatter,
+                    distanceToNextManeuver = step.distance,
+                    theme = theme) {
+                      contentBuilder(upcomingInstruction)
+                    }
+                Spacer(modifier = Modifier.height(8.dp))
+                HorizontalDivider(thickness = 1.dp)
+              }
+            }
```
Archdoog, line 103: Around here we probably also want the footer w/ a close like on iOS? Whatever "arrow up" option is included in the material icons we've included. This would make the lazy column look like:

```kotlin
LazyColumn(
    verticalArrangement = Arrangement.spacedBy(16.dp)
) {
    items(remainingSteps.drop(1)) { step ->
        ManeuverInstructionView(step)
        HorizontalDivider(thickness = 1.dp) // interestingly most android tables seem to omit dividers these days.
    }

    item { 
         // on click could be omitted given the whole list clicks, but may be handy if someone doesn't want tap anywhere.
         InstructionsFooter(onClick: { isExpanded = !isExpanded }) 
    }
}
```

Archdoog (commented):
Looks good. Proposed a small compose improvement for iOS parity.

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
ianthetechie: This doesn't address the horizontal rule, but it is probably marginally more efficient and lets us drop one manual spacer.

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
ianthetechie: image

How do you feel about this visual? I'm not sold on the button styling myself since it's pretty heavy, but it's how material theming works out of the box, and it feels like getting into a war if we want to fight it :/

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
ianthetechie: Nevermind; I found a way to do this using `heightIn` and `animateContentSize`; it's not really possible to do this correctly with a fixed collaped size.

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
ianthetechie: Updated video:

https://github.com/user-attachments/assets/2d781587-4559-4049-befb-0dc7e5e3d06d

ianthetechie: Discussed over Slack; found something better.

image

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
```diff
@@ -36,23 +57,70 @@ fun InstructionsView(
...
+                    Modifier.height(36.dp).fillMaxWidth().clickable { isExpanded = !isExpanded }) {
+                  Icon(
+                      if (isExpanded) {
+                        Icons.Default.KeyboardArrowUp
+                      } else {
+                        Icons.Default.KeyboardArrowDown
+                      },
+                      modifier = Modifier.align(Alignment.Center),
+                      contentDescription = "Show upcoming maneuvers")
+                }
```
Archdoog, line 121: Getting there. I still think it the height has to be extremely small when not expanded. Maybe even so much so that we use an offset to overlay the pill on the padding to keep the view tight. As its own footer composable, maybe something like:

```fun
@Composable
fun InstructionsViewFooter(
  isExpanded: Boolean,
  iconTint: Color
) {
  val height: Dp = if (isExpanded) 36.dp else 8.dp

  Box(
      modifier =
          Modifier.fillMaxWidth()
              .height(height)
          // TODO: Clickable if not handled by the outer InstructionsView
  ) {
            if (isExpanded) {
              Icon(Icons.Default.KeyboardArrowUp,
                modifier = Modifier.align(Alignment.Center),
                contentDescription = "Show upcoming maneuvers",
                tint = iconTint)
            } else {
              Box(
                  modifier = Modifier.fillMaxWidth()
                      .height(6.dp)
                      .background(iconTint, RoundedCornerShape(6.dp)))
            }
          }
}
```

Apologies if this doesn't work (cargoNdk is still failing). But a centered pill with minimal height when not expanded is the goal. I don't think the full height down arrow button will acceptable for most app's UIs (too much wasted space).

As an aside. Google maps doesn't even have the full instrument list these days of some reason. Maybe why it's hard to find an android example.

ianthetechie: Thanks; will mess around with this some more. The reason it is larger currently is because the touch target was too small otherwise.

ianthetechie: Resolved by making the entire banner a click target _for expansion_. This lets us still use a thin pill and offsets to keep a slim view while keeping it usable.

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/theme/InstructionRowTheme.kt
```diff
@@ -33,5 +33,5 @@ object DefaultInstructionRowTheme : InstructionRowTheme {
     @Composable get() = MaterialTheme.colorScheme.onSurface
 
   override val backgroundColor: Color
-    @Composable get() = MaterialTheme.colorScheme.surface
+    @Composable get() = MaterialTheme.colorScheme.surfaceContainerLow
```
ianthetechie, line 36: FWIW this came from the Material 3 [spec](https://m3.material.io/components/bottom-sheets/specs)

Archdoog: My only comment on this is: 

When material theme is used, the more complex the value is, the less likely someone has it in their theme. This isn't so much a problem. More just a warning when someone asks why their instructions row isn't changing color. It'll be because people rarely set those very nuanced values.

Archdoog (approved):
Looking good!

PR #327, opened 2024-10-30
Add fixes for SimulatedLocationProvider

ahmedre:
Instead of determining whether or not to start the simulation based on whether or not the job is null, instead, use the active state of the job to determine whether or not to start the simulator. This allows the SimulatedLocationProvider to be reused more than once, and also allows for addListener to happen before setSimulatedRoute is called without breaking the simulation.


ianthetechie (approved):
Good catch; Thanks!

PR #328, opened 2024-10-30
[common] Update to georust/geo v0.29.0

michaelkirk:
A lot of the line measure traits were re-worked.

Mostly this is just moving code around in hopes of similar methods across metric spaces being easier to find now that they are uniform.

It also enabled replacing some mostly copy/pasted implementations in geo with generic implementations, which enabled some new functionality - e.g. we can now `.densify::<Geodesic>`

One notable behavioral change: the output of the new `Bearing` trait is now uniformly 0...360.  Previously GeodesicBearing and HaversineBearing returned -180..180 while RhumbBearing returned 0...360. (see https://github.com/georust/geo/issues/1210 for more)

This actually uncovered a bug in ferrostar, reflected in the new output of
ferrostar/src/snapshots/ferrostar__simulation__tests__state_from_polyline.snap

Since previously we were casting a potentially negative number to u16 — now it's always positive (0...360).

There are some other changes I'd like to make, but this represents a mostly drop in replacement for the current behavior.

### DIFF
--- a/common/ferrostar/src/algorithms.rs
+++ b/common/ferrostar/src/algorithms.rs
@@ -12,8 +12,8 @@ use crate::{
     navigation_controller::models::TripProgress,
 };
 use geo::{
-    Closest, ClosestPoint, Coord, EuclideanDistance, GeodesicBearing, HaversineDistance,
-    HaversineLength, LineLocatePoint, LineString, Point,
+    Bearing, Closest, ClosestPoint, Coord, Distance, Euclidean, Geodesic, Haversine, Length,
+    LineLocatePoint, LineString, Point,
 };
 
 #[cfg(test)]
@@ -66,8 +66,8 @@ pub fn index_of_closest_segment_origin(location: UserLocation, line: &LineString
         // Find the line segment closest to the user's location
         .min_by(|(_, line1), (_, line2)| {
             // Note: lines don't implement haversine distances
-            let dist1 = line1.euclidean_distance(&point);
-            let dist2 = line2.euclidean_distance(&point);
+            let dist1 = Euclidean::distance(line1, &point);
+            let dist2 = Euclidean::distance(line2, &point);
             dist1.total_cmp(&dist2)
         })
         .map(|(index, _)| index as u64)
@@ -86,7 +86,7 @@ fn get_bearing_to_next_point(
     let next = points.next()?;
 
     // This function may return negative bearing values, but we want to always normalize to [0, 360)
-    let degrees = normalize_bearing(current.geodesic_bearing(next));
+    let degrees = normalize_bearing(Geodesic::bearing(current, next));
 
     Some(CourseOverGround {
         degrees,
@@ -161,7 +161,7 @@ fn snap_point_to_line(point: &Point, line: &LineString) -> Option<Point> {
     // Bail early when we have two essentially identical points.
     // This can cause some issues with edge cases (captured in proptest regressions)
     // with the underlying libraries.
-    if line.euclidean_distance(point) < 0.000_001 {
+    if Euclidean::distance(line, point) < 0.000_001 {
         return Some(*point);
     }
 
@@ -226,7 +226,7 @@ fn snap_point_to_line(point: &Point, line: &LineString) -> Option<Point> {
 /// ```
 pub fn deviation_from_line(point: &Point, line: &LineString) -> Option<f64> {
     snap_point_to_line(point, line).and_then(|snapped| {
-        let distance = snapped.haversine_distance(point);
+        let distance = Haversine::distance(snapped, *point);
 
         if distance.is_nan() || distance.is_infinite() {
             None
@@ -243,7 +243,7 @@ fn is_close_enough_to_end_of_linestring(
 ) -> bool {
     if let Some(end_coord) = current_step_linestring.coords().last() {
         let end_point = Point::from(*end_coord);
-        let distance_to_end = end_point.haversine_distance(current_position);
+        let distance_to_end = Haversine::distance(end_point, *current_position);
 
         distance_to_end <= threshold
     } else {
@@ -310,8 +310,8 @@ pub fn should_advance_to_next_step(
                         // If the user's distance to the snapped location on the *next* step is <=
                         // the user's distance to the snapped location on the *current* step,
                         // advance to the next step
-                        current_position.haversine_distance(&next_step_closest_point)
-                            <= current_position.haversine_distance(&current_step_closest_point)
+                        Haversine::distance(current_position, next_step_closest_point)
+                            <= Haversine::distance(current_position, current_step_closest_point)
                     } else {
                         // The user's location couldn't be mapped to a single point on both the current and next step.
                         // Fall back to the distance to end of step mode, which has some graceful fallbacks.
@@ -366,7 +366,7 @@ pub(crate) fn advance_step(remaining_steps: &[RouteStep]) -> StepAdvanceStatus {
 /// The result is given in meters.
 /// The result may be [`None`] in case of invalid input such as infinite floats.
 fn distance_along(point: &Point, linestring: &LineString) -> Option<f64> {
-    let total_length = linestring.haversine_length();
+    let total_length = linestring.length::<Haversine>();
     if total_length == 0.0 {
         return Some(0.0);
     }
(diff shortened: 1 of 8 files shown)


common/ferrostar/src/algorithms.rs
```diff
@@ -66,8 +49,8 @@ pub fn index_of_closest_segment_origin(location: UserLocation, line: &LineString
         // Find the line segment closest to the user's location
         .min_by(|(_, line1), (_, line2)| {
             // Note: lines don't implement haversine distances
```
michaelkirk, line 51: I think you might want https://docs.rs/geo/latest/geo/algorithm/cross_track_distance/trait.CrossTrackDistance.html

I think geo should one day export that as `Haversine::distance(line, point)`. Relatedly, see https://github.com/georust/geo/issues/1128

common/ferrostar/src/algorithms.rs
```diff
@@ -85,13 +68,8 @@ fn get_bearing_to_next_point(
...
     let next = points.next()?;
 
-    // This function may return negative bearing values, but we want to always normalize to [0, 360)
-    let degrees = normalize_bearing(current.geodesic_bearing(next));
-
-    Some(CourseOverGround {
-        degrees,
-        accuracy: None,
-    })
+    let degrees = Geodesic::bearing(current, next);
```
michaelkirk, line 71: I noticed there was a mix of `Geodesic` vs. `Haversine` calculations used throughout - is that intentional?

If you aren't super worried about accuracy Haversine will be a little quicker.

common/ferrostar/src/algorithms.rs
```diff
@@ -379,9 +357,9 @@ fn distance_along(point: &Point, linestring: &LineString) -> Option<f64> {
 
             // Compute distance to the line (sadly Euclidean only; no haversine_distance in GeoRust
             // but this is probably OK for now)
-            let segment_distance_to_point = segment.euclidean_distance(point);
+            let segment_distance_to_point = Euclidean::distance(&segment, point);
             // Compute total segment length in meters
-            let segment_length = segment_linestring.haversine_length();
+            let segment_length = segment_linestring.length::<Haversine>();
```
michaelkirk, line 362: For example, you could double down on the more accurate but slower `segment_linestring.length::<Geodesic>()`

common/ferrostar/src/models.rs
```diff
@@ -171,8 +171,17 @@ pub struct CourseOverGround {
...
 
 impl CourseOverGround {
-    pub fn new(degrees: u16, accuracy: Option<u16>) -> Self {
-        Self { degrees, accuracy }
+    /// # Arguments
+    ///
+    /// - degrees: The direction in which the user's device is traveling, measured in clockwise degrees from
+    ///            true north (N = 0, E = 90, S = 180, W = 270).
+    /// - accuracy: the accuracy of the course value, measured in degrees.
+    pub fn new(degrees: f64, accuracy: Option<u16>) -> Self {
```
michaelkirk, line 179: This is a larger change than necessary, but it cleaned things up a bit. 

A couple things:

1. Are you comfortable trusting input to be 0...360 now that `geo::Bearing` does so?
2. It's maybe a little weird that this `new` takes in an f64 while storing it as u16, but it does reflect how we're using the API in practice.

I can make changes if you'd prefer.

common/ferrostar/src/snapshots/ferrostar__simulation__tests__state_from_polyline.snap
```diff
@@ -8,7 +9,7 @@ current_location:
     lng: -149.543469
   horizontal_accuracy: 0
   course_over_ground:
-    degrees: 0
+    degrees: 288
```
michaelkirk, line 12: This reflects a bug in the previous implementation where negative bearings were being coalesced to 0 in [location_simulation_from_coordinates](https://github.com/stadiamaps/ferrostar/pull/328/files#diff-5668545870bc04900016eb8c61c7df2af165eb8dde8b4072d42396986d2d2e66L109)

common/ferrostar/src/algorithms.rs
ianthetechie: Yes, it was at the time. Unfortunately I can't remember now the specifics, but I believe it was related to edge cases I had discovered the hard way. For example, there's a comment around here about how one or the other will panic on an exact intersection. I agree we should be able to use haversine for most cases. For things involving direct computation of distances, perhaps geodesics make more sense, since there could be a larger error over things like a cross-country route through the US.

common/ferrostar/src/algorithms.rs
ianthetechie: Yeah, exactly. That's probably worth evaluating properly. I've opened #329 to follow up on this. This is _probably_ a case where Geodesics could make sense. On the other hand, I am pretty sure Valhalla uses haversine exclusively and is generally fine 😂

common/ferrostar/src/snapshots/ferrostar__simulation__tests__state_from_polyline.snap
ianthetechie: Good catch, and thanks for the explanation! That explains the same question I had on https://github.com/stadiamaps/ferrostar/pull/323#discussion_r1821766112 yesterday!

common/ferrostar/src/models.rs
ianthetechie: Ooh, thanks for asking!

1. Yes. [iOS](https://developer.apple.com/documentation/corelocation/cllocationdirection) and [Android](https://developer.android.com/reference/android/location/Location#getBearing()) also make guarantees about the range of valid values (iOS uses negative to indicate an error/unavailability, and we already handle that [here](https://github.com/stadiamaps/ferrostar/blob/065a7d2ab94ae1227d593a254ea5896cc4315be1/apple/Sources/FerrostarCore/Extensions/CoreLocationExtensions.swift#L16). I DO think we should document this though just so it's clear to users. The debug_assert is also a good call. After a think over it, I think this is right since it's going to be pretty hard to get a negative value on most platforms.
2. Yeah, agreed. It probably makes sense to accept a float and then do conversion in just one place so that we know it's consistent.

common/ferrostar/src/algorithms.rs
ianthetechie: That... is exactly what we want and I just didn't know what it was named 😅

ianthetechie: Upon further inspection, this sounds like a great idea, but what I really wanted is highly specific to the [Line](https://docs.rs/geo-types/0.7.13/geo_types/geometry/struct.Line.html) implementation. Perhaps `Line` is actually mis-named, as it is really describing a line _segment_. Consider the case of a perfect NW line `[(0,0), (1,1), ...]`. The algorithm using cross track distance uses the mathematical definition of a line as infinite and passing between the points. This means that even at point (100, 100), the algorithm as specified will return index zero, or point (0, 0)!!

That is not at all what we expect. What we actually want is not the closest abstract line that matches, but the _segment_ nearest to us. I'll add some clarifying comments to the code here.

common/ferrostar/src/models.rs
```diff
@@ -171,8 +171,17 @@ pub struct CourseOverGround {
...
+        debug_assert!(degrees >= 0.0 && degrees <= 360.0);
```
ianthetechie, line 180: Pretty sure this should be < 360, not <=... https://github.com/georust/geo/pull/1216/files#diff-7119b05faab26ac9655d8487127708ac29437993afa42df72da07de446285d2aR38

common/ferrostar/src/algorithms.rs
michaelkirk: You are better at reading documentation than me! I mistakenly thought that CrossTrackDistance was a measure to a segment.

common/ferrostar/src/algorithms.rs
michaelkirk: None of the methods should panic! 

You weren't by chance thinking of `VincentyDistance` were you? 

Vincenty is sort of between `Geodesic` and `Haversine` on the accuracy vs speed axis, but has the significant downside that it does not converge for nearly antipodal points, and thus returns an Err (but shouldn't panic).

common/ferrostar/src/models.rs
ianthetechie: EDIT: Looks like I have to file a bug in Android for contradictory docs. The bearing setter does mod 360 there too but the getter docs are contradictory on whether the 360 is inclusive or exclusive haha.

common/ferrostar/src/algorithms.rs
ianthetechie: Yeah, it wasn't Vincenty. It was with haversine_closest_point. I had documented the issue here: https://github.com/stadiamaps/ferrostar/blob/6133f12a12a187097d1b22b1e30acbcfa5e281ac/common/ferrostar/src/algorithms.rs#L173. 

Upstream issue: https://github.com/georust/geo/issues/1084.

Someone else beat me to a PR: https://github.com/georust/geo/pull/1119. It looks like it made it into a release but I didn't get around to checking. Will see if it works ;)

ianthetechie (approved):

common/ferrostar/src/algorithms.rs
michaelkirk: Oh right, that. I'm sorry about that. 😞 

That should be fixed as of 0.28.0, released in February.

PR #330, opened 2024-10-31
iOS Route Overview

ianthetechie:
Opening this as a draft PR because there are still a few problems with it, but I want feedback on the overall approach. I'll leave comments to guide the first review.

Ignore the terrible quality of simulator animated GIFs... it is not really that choppy, but it DOES take too long to get back to the user's real position because of intermediate camera updates, which is a problem we'll need to fix.

Simulator Screen Recording - iPhone 16 Pro - 2024-10-31 at 19 03 03

### DIFF
--- a/apple/Tests/TestSupport/AssertSnapshot.swift
+++ b/apple/Tests/TestSupport/AssertSnapshot.swift
@@ -24,7 +24,7 @@ public extension XCTestCase {
             .background(Color(red: 130 / 255, green: 203 / 255, blue: 114 / 255))
 
         assertSnapshot(of: view,
-                       as: .image(perceptualPrecision: 0.98),
+                       as: .image(precision: 0.99),
                        named: name,
                        record: record,
                        file: file,
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -22,12 +22,12 @@ jobs:
 
     - name: Check format
       run: swiftformat . --lint
-      
+
   build-ferrostar:
     runs-on: macos-15
     permissions:
       contents: write  # To auto-commit Package.swift and binding changes
-    
+
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-ios-build-lib
       cancel-in-progress: true
@@ -84,7 +84,7 @@ jobs:
         scheme: [
           "iOS Demo"
         ]
-    
+
     steps:
     - name: Checkout repo
       uses: actions/checkout@v4
@@ -116,7 +116,7 @@ jobs:
       run: xcodebuild -scheme "${{ matrix.scheme }}" build -configuration Debug -skipMacroValidation \
         CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO | xcbeautify && exit ${PIPESTATUS[0]}
       working-directory: apple/DemoApp
-      
+
     - name: 'Upload Artifact'
       uses: actions/upload-artifact@v4
       with:
@@ -173,11 +173,11 @@ jobs:
         name: Library-Package.swift
         path: Package.swift
         retention-days: 5
-        
+
     - name: 'Upload Test Results Artifact'
       uses: actions/upload-artifact@v4
       if: always()
       with:
-        name: ${{ matrix.scheme }}-test-results
+        name: ${{ matrix.scheme }}.xcresult
         path: ${{ matrix.scheme }}.xcresult
         retention-days: 5


apple/Sources/FerrostarMapLibreUI/Extensions/NavigationState.swift
```diff
@@ -0,0 +1,30 @@
...
+                    coord.lng
+                )),
+                ne: CLLocationCoordinate2D(
+                    latitude: max(acc.ne.latitude, coord.lat),
+                    longitude: max(acc.ne.longitude, coord.lng)
+                )
+            )
+        }
+
+        return MapViewCamera.boundingBox(bounds, edgePadding: UIEdgeInsets(top: 20, left: 100, bottom: 20, right: 100))
```
ianthetechie, line 28: The padding values are silly because the implementation in the MapLibre SwiftUI DSL seems to be broken.

apple/Sources/FerrostarMapLibreUI/Views/DynamicallyOrientingNavigationView.swift
```diff
@@ -106,8 +107,13 @@ public struct DynamicallyOrientingNavigationView: View, CustomizableNavigatingIn
                         showZoom: true,
                         onZoomIn: { camera.incrementZoom(by: 1) },
                         onZoomOut: { camera.incrementZoom(by: -1) },
-                        showCentering: !camera.isTrackingUserLocationWithCourse,
-                        onCenter: { camera = navigationCamera },
+                        cameraControlState: camera.isTrackingUserLocationWithCourse ? .showRouteOverview {
+                            if let overviewCamera = navigationState?.routeOverviewCamera {
+                                camera = overviewCamera
+                            }
+                        } : .showRecenter { // TODO: Third case when not navigating!
```
ianthetechie, line 114: TBD the best way to express this logic given the limits of view code; probably in an extension / helper function.

apple/Sources/FerrostarSwiftUI/Views/Controls/MuteUIButton.swift
```diff
@@ -12,6 +12,7 @@ public struct MuteUIButton: View {
     }
 
     public var body: some View {
+        // TODO: Use NavigationUIButton?
```
ianthetechie, line 15: Noticed that the mute PR didn't use the convention that others did; probably a quick fix.

apple/Sources/FerrostarSwiftUI/Views/GridViews/NavigatingInnerGridView.swift
```diff
@@ -1,6 +1,23 @@
...
+    ///
+    /// The action is responsible for resetting the camera
+    /// to a state that follows the user.
+    case showRecenter(() -> Void)
+
+    /// Shows the route overview button.
+    ///
+    /// The action is responsible for transitioning the camera to an overview of the route.
+    case showRouteOverview(() -> Void)
+}
```
ianthetechie, line 19: This is the first significant design decision. I think that an enum makes the camera control states significantly cleaner to express. We can add more later if necessary, but the common ones are here.

apple/Sources/FerrostarSwiftUI/Views/GridViews/NavigatingInnerGridView.swift
```diff
@@ -86,6 +96,29 @@ public struct NavigatingInnerGridView: View, CustomizableNavigatingInnerGridView
...
+                    .shadow(radius: 8)
+                case let .showRouteOverview(onRouteOverview):
+                    NavigationUIButton(action: onRouteOverview) {
+                        Image(systemName: "point.bottomleft.forward.to.point.topright.scurvepath")
+                            .resizable()
+                            .aspectRatio(contentMode: .fit)
+                            .frame(width: 18, height: 18)
+                    }
+                    .shadow(radius: 8)
+                }
```
ianthetechie, line 120: The important thing to note here is that we have moved the recenter button to be in the top corner. The logic behind that is that we *probably* only ever care about having one camera button visible at a time in my opinion (differing opinions welcome :D). So the camera button slot will either be a recenter button, a route overview button, or hidden, depending on state driven higher up.

ianthetechie (commented):
Here are some things to look at in a first pass before I get too far down the rabbit hole.

apple/Sources/FerrostarSwiftUI/Views/Controls/MuteUIButton.swift
ianthetechie: Addressed it #342

apple/Sources/FerrostarMapLibreUI/Views/DynamicallyOrientingNavigationView.swift
```diff
@@ -117,6 +123,8 @@ public struct DynamicallyOrientingNavigationView: View, CustomizableNavigatingIn
                         topTrailing?()
                     } midLeading: {
                         midLeading?()
+                    } bottomLeading: {
+                        bottomLeading?()
```
ianthetechie, line 127: Undo this change; recenter will stay in lower left corner.

apple/Sources/FerrostarSwiftUI/Views/GridViews/NavigatingInnerGridView.swift
ianthetechie: Discussed; we will move the recenter button to the bottom left corner to keep with convention.

ianthetechie: Re-discussed months later and decided to keep it top right 😂

ianthetechie (commented):
I think this needs a bit more work before merging.

# Route overview button shows even when not navigating

And, as a slight variation, the recenter button shows up when you exit navigation, but this is less of a problem.

simulator_screenshot_22CCDDA1-17ED-4281-ACB3-FE48A3832A93



# Inconsistent observation of camera state

When moving the camera manually, it does not immediately realize that the camera is no longer tracking. It seems only a _second_ movement triggers this.

simulator_screenshot_E5EBB252-80F3-46D6-8F78-69E794C0DD90

# The fly effect when recentering from the overview is quite jarring


https://github.com/user-attachments/assets/0b4578b3-4ade-49c9-97f4-cce4a1976443

Archdoog:
@ianthetechie it lives! It's not as nice as android, but it's passable IMO.

cc @bolsinga since you were asking about it.

apple/Tests/TestSupport/AssertSnapshot.swift
```diff
@@ -24,7 +24,7 @@ public extension XCTestCase {
             .background(Color(red: 130 / 255, green: 203 / 255, blue: 114 / 255))
 
         assertSnapshot(of: view,
-                       as: .image,
+                       as: .image(precision: 0.99),
```
Archdoog, line 27: The test artifacts were throwing a precision issue with color divergence on CI. 99% precision is still higher than most of the iOS snapshot testing I do CI.

apple/Sources/FerrostarSwiftUI/Views/GridViews/NavigatingInnerGridView.swift
Archdoog: Let's review this alongside android on the call. Since it's been so long.

PR #342, opened 2024-11-03
Fix button inconsistencies

ianthetechie:
Pretty straightforward. Does all the things that probably should have been done the first time ;)

* Builds on `NavigationUIButton`, which we designed specifically for this consistency
* Adds shadow modifier that we used to have everywhere else.
* (Possibly contentious) Moves shadow modifiers inside the views. We have these strewn all over the place and it just makes things confusing and incosistent IMO; these views are already quite opinionated and often have half a dozen modifiers internally anyways. I vaguely remember removing these at one point due to spurious snapshot differences across OS versions, but we have to lock these anyways for CI, so I'd like to revisit if that's possible. This is all in a separate commit so can roll back easily enough if you'd prefer not to go this far.


Archdoog (approved):
Looking good per the snapshot tests 👍

PR #343, opened 2024-11-04
Pin JDK toolchain

ianthetechie:
Here's an attempt to pin down one free variables in troubleshooting build tooling issues on some machines. Adopted from the Android docs here: https://developer.android.com/build/jdks#toolchain.

I think at a minimum we can at least specify the `JavaLanguageVersion` to force using JDK 17 consistently. I am NOT sure about specifying the vendor; this may be going a bit too far. Perhaps @haysmike has some input on this.

### DIFF
--- a/android/composeui/build.gradle
+++ b/android/composeui/build.gradle
@@ -21,6 +21,13 @@ android {
         consumerProguardFiles "consumer-rules.pro"
     }
 
+    java {
+        toolchain {
+            languageVersion = JavaLanguageVersion.of(17)
+            vendor = JvmVendorSpec.ADOPTIUM
+        }
+    }
+
     compileOptions {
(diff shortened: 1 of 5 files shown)


ianthetechie:
@ahmedre forwarded my a useful write-up this morning on why this probably isn't a great idea to enforce at compile time. Also incidentally some helpful bits on how to keep your JDKs up to date with homebrew.

Let's discuss today on the call.

Archdoog:
Yeah, I feel like I would avoid this as well. I think we've got a workaround for the cargo cli issue until more info comes along. Maybe we can just pin the comment in our issues board?

ianthetechie:
Sounds good; closing for now. Il'l see about pinning the issues for visibility.

PR #345, opened 2024-11-04
Android route overview

### LINKED ISSUE #158: [Android] Route overview
Add a route overview button + functionality to the MapLibre UI toolkit on Android.

ianthetechie:
https://github.com/user-attachments/assets/6f7f4ce9-fd83-4b40-8495-4d818d08db18

Closes #158


android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/config/VisualNavigationViewConfig.kt
```diff
@@ -0,0 +1,34 @@
+package com.stadiamaps.ferrostar.composeui.config
+
+// TODO: Nest cases
```
ianthetechie, line 3: No longer a TODO

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/gridviews/NavigatingInnerGridView.kt
```diff
@@ -62,13 +91,7 @@ fun NavigatingInnerGridView(
...
       },
       bottomStart = {
-        if (showCentering) {
-          NavigationUIButton(onClick = onClickCenter) {
-            Icon(
-                Icons.Filled.Navigation,
-                contentDescription = stringResource(id = R.string.recenter))
-          }
-        }
+        // TODO: Fill this with something else optionally
```
ianthetechie, line 94: Revert this behavior; hide zoom when not following.

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/views/overlays/PortraitNavigationOverlayView.kt
```diff
@@ -64,14 +68,35 @@ fun PortraitNavigationOverlayView(
...
+        } else {
+          val bbox = uiState.routeGeometry?.boundingBox()
+          if (viewModel.isNavigating() && cameraIsTrackingLocation && bbox != null) {
+            CameraControlState.ShowRouteOverview {
+              val scale = density.density
+              camera.value =
+                  MapViewCamera.BoundingBox(
+                      LatLngBounds.from(bbox.north, bbox.east, bbox.south, bbox.west),
+                      padding =
+                          CameraPadding(20.0 * scale, 200.0 * scale, 100.0 * scale, 150.0 * scale))
```
ianthetechie, line 83: Pass in a data class of the instructions AND progress view sizes.

We also need to hide the road name!

ianthetechie (commented):
Few comments from the discussion but have a good path forward.

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/views/overlays/PortraitNavigationOverlayView.kt
ianthetechie: Scratch that; figured out something better :P Will comment up the new diff.

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/extensions/VisualNavigationViewConfig.kt
```diff
@@ -0,0 +1,65 @@
...
+        val bbox = uiState.routeGeometry?.boundingBox()
+        if (bbox != null) {
+          val scale = LocalDensity.current.density
+          val progressViewHeight = navigationViewMetrics.progressViewSize.height.value.toDouble()
+          val instructionsViewHeight =
+              navigationViewMetrics.instructionsViewSize.height.value.toDouble()
+          val layoutDirection = LocalLayoutDirection.current
+
+          // Bottom padding must take the recenter button into account
+          val bottomPadding = (progressViewHeight + this.buttonSize.height.value + 50) * scale
```
ianthetechie, line 39: Let it be known that I really don't like that we have to do this as a result of moving the button, further reducing the amount of space available for the map. I still think the UI would make more sense if we left the recenter button in the top end grid cell.

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/extensions/VisualNavigationViewConfig.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.maplibre.compose.camera.MapViewCamera
+import com.maplibre.compose.camera.models.CameraPadding
+import com.stadiamaps.ferrostar.composeui.config.CameraControlState
+import com.stadiamaps.ferrostar.composeui.config.VisualNavigationViewConfig
+import com.stadiamaps.ferrostar.core.NavigationUiState
+import com.stadiamaps.ferrostar.core.boundingBox
+import com.stadiamaps.ferrostar.maplibreui.NavigationViewMetrics
+
+@Composable
+fun VisualNavigationViewConfig.cameraControlState(
```
ianthetechie, line 19: 1. This is as good a place as any to drop this I think... I think an extension is better than a free function, but there's no place we colud put this that would remove a substantial number of parameters.
2. Note that this can't live in the same file as the declaration in the `composeui` module, since almost everything about it is MapLibre specific.

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/views/overlays/LandscapeNavigationOverlayView.kt
```diff
@@ -36,9 +43,10 @@ import kotlinx.coroutines.flow.asStateFlow
 fun LandscapeNavigationOverlayView(
     modifier: Modifier,
     camera: MutableState<MapViewCamera>,
-    navigationCamera: MapViewCamera = navigationMapViewCamera(),
+    navigationCamera: MapViewCamera,
```
ianthetechie, line 46: I believe that making this a non-optional parameter was a net negative for code quality for not much benefit. It's way too easy to forget to pass it with this many layers, and in fact it was not actually passed down previously. I haven't changed the `progressViewSize` parameter in this PR, but I also strongly suspect that the same is true of that as well and we should require it at compile itme.

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/views/overlays/LandscapeNavigationOverlayView.kt
```diff
@@ -74,11 +93,17 @@ fun LandscapeNavigationOverlayView(
           showMute = config.showMute,
           isMuted = uiState.isMuted,
           onClickMute = { viewModel.toggleMute() },
+          buttonSize = config.buttonSize,
+          cameraControlState =
```
ianthetechie, line 97: Here's the (IMO fairly elegant) solution I ended up with. We can avoid 1) repeating a lot of code, and 2) leaking out the instruction view size (propagating _more_ boilerplate) than if we moved it up a level (which was my first instinct. I'm pretty happy with this since it's minimal code, with all parameters readily available, and users of the overlay just get it "for free" (fewer composable parameters).

ianthetechie (commented):
This should be ready for review now! I've updated the video attached to the PR description showing the updated UX.

android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoNavigationViewModel.kt
```diff
@@ -1,27 +1,36 @@
...
 import uniffi.ferrostar.UserLocation
 
-// NOTE: We are aware that this is not a particularly great ViewModel.
-// We are working on improving this. See the discussion on
-// https://github.com/stadiamaps/ferrostar/pull/295.
-class DemoNavigationViewModel : ViewModel(), NavigationViewModel {
+class DemoNavigationViewModel(
+    // This is a simple example, but these would typically be dependency injected
+    private val ferrostarCore: FerrostarCore = AppModule.ferrostarCore,
+) : ViewModel(), NavigationViewModel {
```
ianthetechie, line 27: I have revamped the view model so it is no longer crap.

android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoNavigationScene.kt
```diff
@@ -122,15 +111,9 @@ fun DemoNavigationScene(
...
-        viewModel.stopNavigation()
-        val vm = DemoNavigationViewModel()
-        viewModel = vm
-
-        vm.startLocationUpdates(locationProvider)
-      },
+      onTapExit = { viewModel.stopNavigation(stopLocationUpdates = false) },
       userContent = { modifier ->
-        if (!viewModel.isNavigating()) {
+        if (!vmState.isNavigating()) {
```
ianthetechie, line 116: These evil bits are gone!

android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
```diff
@@ -299,9 +301,11 @@ class FerrostarCore(
     }
   }
 
-  fun stopNavigation() {
+  fun stopNavigation(stopLocationUpdates: Boolean = true) {
     foregroundServiceManager?.stopService()
-    locationProvider.removeListener(this)
+    if (stopLocationUpdates) {
+      locationProvider.removeListener(this)
```
ianthetechie, line 307: New argument that doesn't break source compatibility; lets us keep the location provider running, which makes life easier for single map view apps.

android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
```diff
@@ -218,10 +225,7 @@ class FerrostarCore(
    * @throws UserLocationUnknown if the location provider has no last known location.
    */
   @Throws(UserLocationUnknown::class)
-  fun startNavigation(
-      route: Route,
-      config: NavigationControllerConfig? = null
-  ): DefaultNavigationViewModel {
+  fun startNavigation(route: Route, config: NavigationControllerConfig? = null) {
```
ianthetechie, line 228: Major change 🚨 

The state flow was already exposed. And @Archdoog was telling me for some time that this was a bit of an antipattern 😅 So this resolves long-standing tech debt. We no longer return view models here, which actually encourages people to write their own to purpose if needed, and the "default" is still quite usable for modal activities.

ianthetechie (commented):
Ended up discovering some bugs in real world testing, which led me to realize the view model swopping antipattern in the demo app HAD to go ;)

android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoNavigationScene.kt
```diff
@@ -47,15 +44,10 @@ fun DemoNavigationScene(
...
-  val executor = remember { Executors.newSingleThreadScheduledExecutor() }
-
   // Keeps the screen on at consistent brightness while this Composable is in the view hierarchy.
   KeepScreenOnDisposableEffect()
 
-  // NOTE: We are aware that this is not a particularly great pattern.
-  // We are working on improving this. See the discussion on
-  // https://github.com/stadiamaps/ferrostar/pull/295.
-  var viewModel by remember { mutableStateOf<NavigationViewModel>(DemoNavigationViewModel()) }
+  val viewModel = remember { DemoNavigationViewModel() }
```
Archdoog, line 50: `DemoNavigationViewModel` can move to the AppModule to more accurately represent likely architecture.

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/extensions/VisualNavigationViewConfig.kt
Archdoog: Personally, I'd probably just ignore the button heights in the bounding box calculation. I think just filling the entire rectangle between the instructions and progress view with a reasonable padding probably succeeds in _most_ cases. The chances of a button hiding important route visual are already low & a user can easily adjust the map by pan if need be.

PR #346, opened 2024-11-04
Add React setup instructions to documentation

AdarshSahani003:
## This PR enhances the **Getting Started Guide** by adding setup instructions for **React**, alongside the existing Vue setup guide.

---

##  What’s New

- **Clear Setup Steps**: Comprehensive instructions for setting up a React project using **Ferrostar**.
- **Comparison Highlights**: Points out key differences between **Vue** and **React** setups for better understanding.

---

##  Why This Matters

Including React instructions makes the guide more **accessible** and **developer-friendly** for those who prefer React, broadening the usability of the framework.

---

###  Feedback and Suggestions

I’d love to hear your **thoughts, feedback, or suggestions**! Your input is valuable and will help improve the guide further.

### DIFF
--- a/guide/src/web-getting-started.md
+++ b/guide/src/web-getting-started.md
@@ -1,9 +1,9 @@
 # Getting Started on the Web
 
-This section of the guide covers how to integrate Ferrostar into a web app.
+This section of the guide covers how to integrate **Ferrostar** into a web app using **React**.
 While there are limitations to the web [Geolocation API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API)
 (notably no background updates),
-PWAs and other mobile-optimized sites
+**Progressive Web Apps (PWAs)** and other mobile-optimized sites
 can be a great solution when a native iOS/Android app is impractical or prohibitively expensive.
 
 We'll cover the "batteries included" approach, but flag areas for customization and overrides along the way.
@@ -33,7 +33,7 @@ Then add `wasm()` and `topLevelAwait()` to the `plugins` section of your Vite co
 
 TODO
 
-## Add Ferrostar web components to your web app
+## Add Ferrostar web components to your React App
 
 The Ferrostar web SDK uses the [Web Components](https://developer.mozilla.org/en-US/docs/Web/API/Web_components)
 to ensure maximum compatibility across frontend frameworks.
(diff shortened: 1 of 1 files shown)


guide/src/web-getting-started.md
```diff
@@ -1,9 +1,9 @@
 # Getting Started on the Web
 
-This section of the guide covers how to integrate Ferrostar into a web app.
+This section of the guide covers how to integrate **Ferrostar** into a web app using **React**.
```
ianthetechie, line 3: This is oddly specific; we shouldn't make the setup exclusive to React. I think we can leave the original language and get into specifics later in the document since most things are common web standards.

guide/src/web-getting-started.md
```diff
@@ -33,7 +33,7 @@ Then add `wasm()` and `topLevelAwait()` to the `plugins` section of your Vite co
 
 TODO
 
-## Add Ferrostar web components to your web app
+## Add Ferrostar web components to your React App
```
ianthetechie, line 36: Same as the above on being highly specific to react; the original headline is fine I think.

guide/src/web-getting-started.md
```diff
@@ -43,35 +43,44 @@ You can import the components just like other things you’re used to in JavaScr
...
+id="ferrostar"
+valhallaEndpointUrl="https://api.stadiamaps.com/route/v1"
+styleUrl="https://tiles.stadiamaps.com/styles/outdoors.json"
+profile="bicycle"
+style={{ display: 'block', width: '100%', height: '100%' }} ></ferrostar-map>
+);
+};
+
+export default MyMapComponent;
 ```
```
ianthetechie, line 66: Why delete the HTML example? This still works great :) We can add details about react but not delete the original.

guide/src/web-getting-started.md
```diff
@@ -43,35 +43,44 @@ You can import the components just like other things you’re used to in JavaScr
...
 
-Here we have used Stadia Maps URLs, which should work without authentication for local development.
-(Refer to the [authentication docs](https://docs.stadiamaps.com/authentication/)
-for network deployment details; you can start with a free account.)
```
ianthetechie, line 61: Un-delete this please :)

ianthetechie (changes requested):
This seems like it *replaces* the existing web standards language and examples with React, which is a bit weird. This should _add_ information about React specifics, not remove existing good info :)

I think you also missed a section further down ("Configuration explained") which has a section on Vue. React does not (as far as I know) support web components natively, so this would be a perfect spot to talk about whatever the best way to do things is in React for binding properties.

guide/src/web-getting-started.md
AdarshSahani003: Sorry, I misunderstood about the changes, and made React Specific.
Ok i am going to undo these un-necessary changes.

guide/src/web-getting-started.md
AdarshSahani003: Ok

guide/src/web-getting-started.md
AdarshSahani003: Sorry for this change, as i said i am going to undo this change and make sure that this should be along with React and not  to replace the HTML example.

Is it fine now like this :
--

Now you can use Ferrostar in your Web component like this:

## HTML

```html
<ferrostar-map
id="ferrostar"
  styleUrl="https://tiles.stadiamaps.com/styles/outdoors.json"
  profile="bicycle"
></ferrostar-map>
```

## React

```javascript
import React from 'react';
import { FerrostarMap } from "@stadiamaps/ferrostar-webcomponents";

const MyMapComponent = () => {
return (
<ferrostar-map
id="ferrostar"
styleUrl="https://tiles.stadiamaps.com/styles/outdoors.json"
profile="bicycle"
style={{ display: 'block', width: '100%', height: '100%' }} ></ferrostar-map>
);
};

export default MyMapComponent;
```

guide/src/web-getting-started.md
```diff
@@ -43,35 +43,57 @@ You can import the components just like other things you’re used to in JavaScr
...
   valhallaEndpointUrl="https://api.stadiamaps.com/route/v1"
   styleUrl="https://tiles.stadiamaps.com/styles/outdoors.json"
   profile="bicycle"
 ></ferrostar-map>
 ```
 
-Here we have used Stadia Maps URLs, which should work without authentication for local development.
-(Refer to the [authentication docs](https://docs.stadiamaps.com/authentication/)
-for network deployment details; you can start with a free account.)
+## React
```
ianthetechie, line 61: I believe that the first example should remain generic, and we should have separate sections for framework-specific (React, Vue, etc.) code. This appears to replace the generic code, which is applicable to a large number of users, with a React-specific example.

ianthetechie (changes requested):
Could you have a look over the formatting again and ensure it matches the styling of other guides? We don't really use horizontal rules, and we prefer to have the code indented per language norms as it's very hard to read otherwise.

Out of curiosity, do you have a readily reproducible example (separate git repo, CodeSandbox, etc.) so that we could test it out? We aren't regular users of React on the core team, so we'd want to be able to easily verify that this works.

PR #350, opened 2024-11-05
Fix issue with expanded full screen layout for instructions

ianthetechie:
I'm not quite sure how this got past my testing, but there was something wrong with the layout in the last release, either as a result of complex interaction or perhaps because of how I did the lazy column the first time around. I still don't fully understand why, but even debugging the layout inspector live, the previous iteration would "push" the pill offscreen at which point it was dropped from the composable completely if the list of instructions were long!

Colllapsed:

image

Expanded (large):

image

Detail showing the layout inspector to show how the pill is carefully aligned:

image


android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/views/InstructionsView.kt
```diff
@@ -64,60 +67,78 @@ fun InstructionsView(
...
+            modifier =
+                Modifier.offset {
+                      IntOffset(
+                          0,
+                          if (isExpanded) {
+                            (-4 * density).toInt()
+                          } else {
+                            (-8 * density).toInt()
+                          })
+                    }
```
ianthetechie, line 138: Unfortunately the new layout structure required some changes here. Android Studio gave a lint warning that I should use this closure version of offset since it depends on state. TIL.

Archdoog (approved):
I noticed this issue as well 👍

PR #351, opened 2024-11-05
Fix state flow of mute UI updates on Android

ianthetechie:
Another small bug fix. This one a bit more obvious.

In the present design, the UI state of the view model is exclusively from FerrostarCore's state. But the mute button being pressed is certainly a state update worth knowing about ;) This is less obvious of a problem when you're moving around, but very obvious when you're stationary / go inside, since the UI will appear to freeze until there's a navigation related update.

### DIFF
--- a/android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
+++ b/android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
@@ -103,7 +103,7 @@ class DefaultNavigationViewModel(
 
   private var userLocation: UserLocation? = locationProvider.lastLocation
   private var muteState: MutableStateFlow<Boolean?> =
-      MutableStateFlow(spokenInstructionObserver?.isMuted)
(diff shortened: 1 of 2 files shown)


android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
```diff
@@ -99,18 +102,20 @@ class DefaultNavigationViewModel(
 ) : ViewModel(), NavigationViewModel {
 
   private var userLocation: UserLocation? = locationProvider.lastLocation
+  private var muteState: MutableStateFlow<Boolean?> =
+      MutableStateFlow(spokenInstructionObserver?.isMuted)
```
Archdoog, line 106: There may be issues with this, but was curious since this is usually what you see:

```suggestion
      remember { mutableStateOf(spokenInstructionObserver?.isMuted) }
```

Archdoog (approved):

PR #360, opened 2024-11-15
Fix/android/view builder and speed json

Archdoog:
- Converts Demo's ViewModel to extend `DefaultNavigationViewModel`
- Fixes "speed" vs "value" key in `SpeedUnit`
- Internalizes view builder params so that you can't accidentally try to customize them.
- Adds a public get for custom view (used in maplibreui).


android/core/src/main/java/com/stadiamaps/ferrostar/core/annotation/SpeedSerializationAdapter.kt
```diff
@@ -18,7 +18,7 @@ class SpeedSerializationAdapter : JsonAdapter<Speed>() {
         is Speed.NoLimit -> writer.name("none").value(true)
         is Speed.Unknown -> writer.name("unknown").value(true)
         is Speed.Value ->
-            writer.name("value").value(speed.value).name("unit").value(speed.unit.text)
+            writer.name("speed").value(speed.value).name("unit").value(speed.unit.text)
```
Archdoog, line 21: @ahmedre I believe the actual key should be `"speed"` when there's a value. E.g.

```json
{
  "speed": 56,
  "unit": "km/h"
},
```

Let me know if there are additional cases. Definitely a bit of an obnoxious json object 😄.

Also, do we want to try catch the annotations parsing? On iOS we just convert a json parsing error to nil/null for annotations and an onError callback so that developers can log the failure. We figured better to let navigation succeed even in the case of a degraded/malformed annotations. See https://github.com/stadiamaps/ferrostar/blob/61e299a66a1afc02c046ff52e1d6ff48a1f3cffa/apple/Sources/FerrostarCore/Annotations/AnnotationPublisher.swift#L70-L75

android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/config/NavigationViewComponentBuilder.kt
```diff
@@ -19,14 +19,16 @@ import com.stadiamaps.ferrostar.composeui.views.components.TripProgressView
 import com.stadiamaps.ferrostar.core.NavigationUiState
 
 data class NavigationViewComponentBuilder(
-    val instructionsView: @Composable (modifier: Modifier, uiState: NavigationUiState) -> Unit,
-    val progressView:
+    internal val instructionsView:
```
Archdoog, line 22: This avoids users accidentally trying to do: 

```kotlin
NavigationViewComponentBuilder.Default()
    .instructionsView { }
```

Which is very easy to do and highly problematic.

android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoNavigationViewModel.kt
```diff
@@ -24,7 +25,8 @@ import uniffi.ferrostar.UserLocation
 class DemoNavigationViewModel(
     // This is a simple example, but these would typically be dependency injected
     private val ferrostarCore: FerrostarCore = AppModule.ferrostarCore,
-) : ViewModel(), LocationUpdateListener, NavigationViewModel {
+    annotationPublisher: AnnotationPublisher<*> = valhallaExtendedOSRMAnnotationPublisher()
+) : DefaultNavigationViewModel(ferrostarCore, annotationPublisher), LocationUpdateListener {
```
Archdoog, line 29: @ianthetechie now that this view model uses the `DefaultNavigationViewModel`. We should be able to refactor it to use the existing uiState from ferrostar and implement its own additional state for things like route loading, etc. Just to highlight how easy it is to customize without interrupting the default navigation behavior.

ianthetechie (approved):
Looks good to me but I'll wait for @ahmedre to comment since he may have an opinion based on what they're building 👍

android/core/src/main/java/com/stadiamaps/ferrostar/core/annotation/SpeedSerializationAdapter.kt
ahmedre: yes, you are right - my mistake! agree re: try/catching the parsing with an error callback 👍

ahmedre:
On the `DemoNavigationViewModel` usage of `DefaultNavigationViewModel` - I think it is a step forward, but I think it is not as flexible as we have with just using `FerrostarCore` directly, because in this case, for example, both need to emit a `NavigationUiState` object.

We can swap `NavigationUiState` for a `sealed` class to solve this problem specifically for the demo use case of additional states (which is what we also did in our implementation, but using `FerrostarCore` directly), but if we want to add lane support or speed limit support later, we'll need to either add these to `NavigationUiState` for everyone, or have our own ui state, or have two flows (in which case are we sure they're in sync, or would someone just need to `combine` them together to get a combination).

If we want to go this route, should we consider making `NavigationUiState` an interface with some required fields and let the return type be any instance? The downside of this approach though is that people would need an instance of check, and/or we'd need to do something with generics to avoid that.

ianthetechie:
This looks good to me; going to merge now but happy to discuss other ways to make things clearer on today's call @ahmedre

PR #370, opened 2024-11-25
Run SwiftFormat on commit

Patrick-Kladek:
### Description

To reduce the feedback loop for a failed build when formatting is not acceptable we agreed to run the formatter as part of the commit. Using https://github.com/hallettj/git-format-staged allows us to only format staged files, leaving other files untouched.

### Tasks

- [x] Add pre commit hook
- [x] configure git hook on `./build-ios.sh`

### Infos for Reviewer

I tested this without SwiftFormat installed: The commit will fail with an error message explaining to install SwiftFormat. You still have the option to ignore the hook and commit directly.

Git, by default, does not distribute hooks when a repository is cloned. This design choice is intentional to prevent arbitrary code execution on users’ machines for security reasons. Therefore we need the workaround, where we configure the git hooks path as part for `build-ios.sh`

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -3,8 +3,12 @@
 if which swiftformat >/dev/null; then
 	./.githooks/git-format-staged --formatter "swiftformat stdin --stdinpath '{}'" "apple/*.swift"
 else
-	echo "error: SwiftFormat not installed"
-	echo "error: Install it via 'brew install swiftformat'"
-	echo "error: or download from https://github.com/nicklockwood/SwiftFormat"
-	exit 1
+	staged_swift_files=$(git diff --cached --name-only --diff-filter=ACMR | grep '\.swift$')
+	
+	if [[ -n "$staged_swift_files" ]]; then
+		echo "error: commit contains swift files & SwiftFormat is not installed"
+		echo "error: Install it via 'brew install swiftformat'"
+		echo "error: or download from https://github.com/nicklockwood/SwiftFormat"
+		exit 1
+	fi
 fi
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,2 +1,10 @@
 #!/bin/bash
-./.githooks/git-format-staged --formatter "swiftformat stdin --stdinpath '{}'" "apple/*.swift"
+
+if which swiftformat >/dev/null; then
+	./.githooks/git-format-staged --formatter "swiftformat stdin --stdinpath '{}'" "apple/*.swift"
+else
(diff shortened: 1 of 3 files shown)


.githooks/pre-commit
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+if which swiftformat >/dev/null; then
+	./.githooks/git-format-staged --formatter "swiftformat stdin --stdinpath '{}'" "apple/*.swift"
+else
+	echo "error: SwiftFormat not installed"
+	echo "error: Install it via 'brew install swiftformat'"
+	echo "error: or download from https://github.com/nicklockwood/SwiftFormat"
+	exit 1
```
ianthetechie, line 9: Would it be possible to either skip this check if no Swift changes are staged OR skip the check on non-Apple platforms? All the core devs here use macOS, but I know some end-users do occasionally develop on Windows or Linux if they are android-specific, and this would exclude them from being able to contribute patches easily.

ianthetechie (commented):
This looks good to me except for one thing.

FYI @Archdoog, I discussed the problem of how we (try to) push commits from CI that reform your code with the HudHud team last week, since you had complained about how it'd rewrite your branch causing local conflicts, and @ahmedre also ended up having an issue with this causing an inscrutable CI failure, since the action doesn't work on a cross-org fork.

Patrick suggested that we try this approach, which they have been using for 6+ months without issue. It's not as drastic as blanket running swiftformat (etc.) locally, and my understanding is that it only operates on staged files.

While I've been scarred by past experiences with git commit hooks from random JavaScript repos, I think this is a reasonable approach to try. If you're doing Swift development, there's like a 90%+ chance you already have swiftformat installed, and it doesn't use anything else outside git and the Python standard library.

Any objections? We can discuss in tomorrow's meeting if needed.

.githooks/pre-commit
Patrick-Kladek: good point, I didn't consider non-Apple developers.

Patrick-Kladek:
I'm using https://fork.dev as my git client, so I can show some use cases:

## Failure when swift file is committed and swiftformat is not installed
failure swiftformat not installed

## Success after installing swiftformat
swift success

## Android should be unaffected
java success

Patrick-Kladek:
@ianthetechie could this cause any issue with Windows? for a long time windows didn't support bash but that changed in recent years.

ianthetechie (approved):
We'll just merge this for now and if it causes any issues with a Windows user, we'll cross that bridge when we come to it :)

PR #384, opened 2024-12-03
Properly update sub model with type and modifier

ahmedre:
The type and modifier fields are part of the official MapBox spec for allowed sub fields. In practice, they are mostly unused in favor of using them inside of components instead, but they are allowed and we are already parsing them, so this patch properly propagates them.


ahmedre:
snippet from the documentation:
CleanShot 2024-12-03 at 14 27 40@2x

[source](https://docs.mapbox.com/api/navigation/directions/#route-leg-object).

ianthetechie (approved):
Good catch!

PR #385, opened 2024-12-03
Expose exits on RouteSteps

ahmedre:


ahmedre:
I was checking and exits can be in 2 places - one is here (which already exists), and the second is inside of the various `BannerInstruction.BannerContent` (for `primary`, `secondary`, and `sub`). For the second place, I can open another PR - should we expose `BannerContent`s instead?

In a previous patch, we mapped various fields / types within the `BannerContent` into lanes, so wondering if this is the approach we should continue (i.e. add fields denoting the `exit-numbers` fields in addition to `lanes`, etc), or if we want to keep `lanes` and also add the actual info, or if we want to remove `lanes` altogether in favor of just exposing what is now parsed as `Vec<BannerContentComponent>`?

i.e. option 1:

```rust
pub struct VisualInstructionContent {
    pub text: String,
    pub maneuver_type: Option<ManeuverType>,
    pub maneuver_modifier: Option<ManeuverModifier>,
    pub roundabout_exit_degrees: Option<u16>,
    pub lane_info: Option<Vec<LaneInfo>>,
    pub exit_numbers: Option<Vec<String>>,
}
```


or option 2:

```rust
pub struct VisualInstructionContent {
    pub text: String,
    pub maneuver_type: Option<ManeuverType>,
    pub maneuver_modifier: Option<ManeuverModifier>,
    pub roundabout_exit_degrees: Option<u16>,
    pub lane_info: Option<Vec<LaneInfo>>,
    pub details: Option<Vec<BannerContentComponent>>,
}
```

or option 3 being option 2 with removing `lane_info`.

common/ferrostar/src/models.rs
```diff
@@ -311,6 +311,11 @@ pub struct RouteStep {
     pub duration: f64,
     /// The name of the road being traveled on (useful for certain UI styles).
     pub road_name: Option<String>,
+    /// A list of exits (name or number), separated by semicolons.
+    ///
+    /// NOTE: This annotation is not in the official spec, but is a common extension used by Mapbox
+    /// and Valhalla.
+    pub exits: Option<String>,
```
ianthetechie, line 318: I think we can/should probably just do the splitting here. For the higher level models, there's no reason to push this work down to the user of the library.

ianthetechie:
> I was checking and exits can be in 2 places - one is here (which already exists), and the second is inside of the various BannerInstruction.BannerContent (for primary, secondary, and sub).

Ahh, that's a good point. I don't think these are guaranteed to be the same, so we can handle them each separately. For example, a single route step could technically involve multiple exits with carefully timed banners in a complex highway interchange, so the banners will be more granular than the route step list of strings.

As for the question of how to model the components... that's a good one 😅  A few observations:

1. We currently [only parse lane info from `sub` instructions](https://github.com/stadiamaps/ferrostar/blob/0310469e449d26f45cf4530480728718efd56ad3/common/ferrostar/src/routing_adapters/osrm/mod.rs#L225). I'm not sure that is correct based on the docs; it's not clear whether these could technically show up elsewhere too.
2. I'd prefer to keep `BannerContentComponent` as an implementation detail of OSRM and provide a higher level model _if possible_. I think this makes it a bit more future-proof and easier to evolve over time. I had just typed up something elaborate, but upon my 5th read of the docs, realized that all we need is indeed a list of strings as you suggest, so my vote is for option 1. (Their addition of the local word for "exit" as a whole component is interesting... I'm not sure if I think that's clever or weird and maybe slightly wasteful TBH.)

Slight tangent: What are your thoughts on removing the `Option` wrapper... There are some situations where it actually matters which flavor of "nothing" you have, but I don't think it matters here.

ianthetechie (commented):
Any chance you have a test case we could commit for snapshotting? That'd be useful.

ahmedre:
> We currently [only parse lane info from sub instructions](https://github.com/stadiamaps/ferrostar/blob/0310469e449d26f45cf4530480728718efd56ad3/common/ferrostar/src/routing_adapters/osrm/mod.rs#L225). I'm not sure that is correct based on the docs; it's not clear whether these could technically show up elsewhere too.

Not sure tbh - the only thing I found is the Mapbox API itself, which does return it in `sub`s -

```json
                  "sub": {
                    "components": [
                      {
                        "active": false,
                        "directions": ["straight"],
                        "type": "lane",
                        "text": ""
                      },
                      {
                        "active": false,
                        "directions": ["straight"],
                        "type": "lane",
                        "text": ""
                      },
                      {
                        "active": false,
                        "directions": ["straight"],
                        "type": "lane",
                        "text": ""
                      },
                      {
                        "active_direction": "slight right",
                        "active": true,
                        "directions": ["slight right"],
                        "type": "lane",
                        "text": ""
                      }
                    ],
                    "text": ""
                  },
```

ahmedre:
with respect to exits, i found an api call from Mapbox showing this - it's coming up in primary -

```json
                  "primary": {
                    "components": [
                      {"type": "exit", "text": "Exit"},
                      {"type": "exit-number", "text": "9"},
                      {"type": "text", "text": "Highland Park"}
                    ],
                    "type": "off ramp",
                    "modifier": "right",
                    "text": "Exit 9 Highland Park"
                  },
```

common/ferrostar/src/models.rs
ahmedre: done.

ahmedre:
> Any chance you have a test case we could commit for snapshotting? That'd be useful.

added a test - this one can be used with the second pr where we parse the banner component exit info as well, we'll just have to update the insta result.

common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -266,13 +266,19 @@ impl RouteStep {
                 .collect()
         });
 
+        let exits = match value.exits.clone() {
+            Some(exit_text) => exit_text.split(';').map(|s| s.to_string()).collect(),
```
ianthetechie, line 270: Probably should have a `trim` in here just in case ;)

ianthetechie:
> with respect to exits, i found an api call from Mapbox showing this - it's coming up in primary -

I can confirm Valhalla does the same. I'm going to do some minor cleanup, but I think this can be merged today. Thanks!

ianthetechie:
Oh no! I can't actually push changes since this is on an org-owned fork 😅 But I think I can just push the branch elsewhere and PR that way.

ianthetechie:
Closed as completed via #391

Issue #263, opened 2024-09-24
Add logging to common API failure points


Archdoog:

ianthetechie:
Bit more detail here from the call. The ask here is to add a logging crate which can be configured to dump to stderr so devs can get insight into what's happening in the Rust core from platform code.

ianthetechie:
@Archdoog similarly, I think the PR more or less addresses the issues without getting too crazy and doing stuff the user might not want. For example, tracing I think makes a lot of sense when we aren't already sending a `Result` back up, but since we are sending up a result, then I'm not sure the _library_ code needs to do any more.

PR #387, opened 2024-12-05
Step advance edge case improvements

ianthetechie:
From our call earlier today, here's where we are at so far.

It does appear to solve the problem for the edge cases we discussed, but currently struggling with what appears to be a new edge case at the very end of the route.

- [x] Resolve logging discussion
- [x] Update the guide

[Screen_recording_20241205_163157.webm](https://github.com/user-attachments/assets/73ac4af5-7728-4bf8-907f-499fcbbb3d52)


common/ferrostar/src/navigation_controller/snapshots/ferrostar__navigation_controller__tests__full_route_state_snapshot-2.snap
```diff
@@ -0,0 +1,14132 @@
...
+          - lat: 45.37561
+            lng: -122.588432
+          - lat: 45.375657
+            lng: -122.58831
+          - lat: 45.375703
+            lng: -122.588114
+        distance: 295
+        duration: 12.258
+        road_name: ""
+        instruct
```
ianthetechie, line 14124: Apologies in advance for the mega-diff that will crash your browser, but this is the part that doesn't look quite right. I think this should be empyt at the end of the route, but it's not.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -431,15 +451,22 @@ mod tests {
...
                         assert!(
                             index < geom_length,
                             "index = {index}, geom_length = {geom_length}"
                         );
                     }
+
+                    // Regression test that we are never marked as off the route.
+                    // We used to encounter this with relative step advance on self-intersecting
+                    // routes, for example.
+                    assert_eq!(deviation, &RouteDeviation::NoDeviation);
```
ianthetechie, line 469: In the case of the self-intersecting route, an improper step advance would eventually result in a route deviation (which is why our test case eventually reroutes).

common/ferrostar/src/navigation_controller/snapshots/ferrostar__navigation_controller__tests__full_route_state_snapshot-2.snap
ianthetechie: EDIT: This moved to another file.

common/ferrostar/src/navigation_controller/fixtures/valhalla_self_intersecting_osrm_response.json
ianthetechie, line 1: This is the exact route we looked at on the call earlier + I've captured in the video FWIW

common/ferrostar/src/navigation_controller/snapshots/ferrostar__navigation_controller__tests__self_intersecting_relative_linestring_min_line_distance.snap
```diff
@@ -0,0 +1,14132 @@
...
+            lng: -122.588661
+          - lat: 45.375324
+            lng: -122.588722
+          - lat: 45.375415
+            lng: -122.588676
+          - lat: 45.375483
+            lng: -122.588617
+          - lat: 45.375559
+            lng: -122.58852
+
```
ianthetechie, line 14124: Re-raising now that the files are named sensibly. This is confusing. I would expect the final simulation step, which even says it is directly at the end, to have no remaining waypoints.

ianthetechie:
This should be ready for review now @Archdoog :)

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -130,18 +130,43 @@ pub enum StepAdvanceMode {
...
+#[derive(Debug, Copy, Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+#[cfg_attr(feature = "wasm-bindgen", derive(Deserialize, Tsify))]
+#[cfg_attr(feature = "wasm-bindgen", tsify(from_wasm_abi))]
+pub enum SpecialAdvanceConditions {
+    /// Allows navigation to advance to the next step as soon as the user
+    /// comes within this distance (in meters) of the end of the current step.
+    ///
+    /// This results in *early* advance when the user is near the goal.
+    AdvanceAtDistanceFromEnd(u16),
```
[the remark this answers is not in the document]
ianthetechie: I'm not necessarily opposed to adding warning logs when we think the user may be setting themselves up, but...

1. I think the last PR just configured the standard logging (`log` crate) in a way that logcat picks it up on Android. Is this what you mean by "new logging?"
2. If you don't read the docs, will you really read all 6,000 lines that logcat spews on a typical app launch? I don't want us to add to the noise. It's almost impossible to arrive at `SpecialAdvanceConditions` by accident, so I'm feeling like ensuring good doc comments and demo app code are probably the best (actually that's a good note; I still need to update docs!)
3. If we want to add logging, where should we add it? Maybe in `NavigationController::new`?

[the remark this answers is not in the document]
ianthetechie: Discussed on the call; let's defer this in a task to add a debug log on FerrostarCore init (similar to MapLibre etc.) with a summary of the settings.

Related: #263.

We should also open a task to test logging on iOS.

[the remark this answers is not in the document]
ianthetechie: Captured iOS in #399

common/ferrostar/src/algorithms.rs
```diff
@@ -222,14 +222,15 @@ fn is_close_enough_to_end_of_linestring(
     current_step_linestring: &LineString,
     threshold: f64,
 ) -> bool {
```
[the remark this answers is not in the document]
ianthetechie: Fixed.

PR #388, opened 2024-12-07
Added and tested android logger

Archdoog:
- Adds https://docs.rs/android_logger/latest/android_logger/ & log.

Screenshot 2024-12-06 at 4 28 10 PM


ianthetechie (approved):
Nice! Let's not forget to add some docs to the guide later too ;)

PR #394, opened 2024-12-10
Ferrostar support for Expo Modules & React Native (Android only for now)

bjtrounson:
Introduces a new directory `/expo` to the repo which contains a expo module wrapper for Ferrostar.

The directory is under expo as this module requires the expo package to be installed in your react native project. I also was thinking down the line @ianthetechie might want to look at creating Turbo Module with uniffi bindings for Ferrostar which could live in a react-native directory.
 
Expo modules can be integrated into existing bare react native projects by following this guide here: https://docs.expo.dev/bare/installing-expo-modules/#automatic-installation

**_The only major issue I have found in this module, is when switching between pages and back to the map. The map will disappear  and not reload until the app is relaunched._**

iOS is currently also not supported but I don't think it will be a huge task to do. We only need to sort out how to configure the SPM packages in CocoaPods. Then we can scaffold out a very similar implementation to Android.

### Somethings that should probably will need to be done before fully releasing this
- npm package (this is if we intend to release this as an npm package for others to use.)
- Automated releases
- Fallback when launched on an iOS device (currently I think the app will just crash, I haven't tried it yet.)
- Documentation (looking at doing this on 16th & 17th of December.)


expo/android/src/main/java/expo/modules/ferrostar/ExpoOfflineManagerModule.kt
```diff
@@ -0,0 +1,101 @@
+package expo.modules.ferrostar
```
ianthetechie, line 1: Can _probably_ be deleted?

expo/android/src/main/java/expo/modules/ferrostar/records/BoundingBox.kt
```diff
@@ -0,0 +1,17 @@
+package expo.modules.ferrostar.records
```
ianthetechie, line 1: Let's figure out how to generate these automatically.

expo/example/App.tsx
```diff
@@ -0,0 +1,31 @@
+import { Ferrostar, FerrostarProvider } from "expo-ferrostar";
+import { View } from "react-native";
+
+export default function App() {
+  return (
+            <Ferrostar
+                style={{ flex: 1 }}
+                locationMode="default"
+                styleUrl="https://api.maptiler.com/maps/basic-v2/style.json?key=DQTRKy0N0WUe90KUjN4i"
```
ianthetechie, line 9: whoops ;)

ianthetechie (commented):
Overall solid start. As discussed on the call, the big thing we want to figure out is how to autogenerate all of the core bindings, since these will go out of sync pretty quickly (not to mention it's a lot of typing!).

bjtrounson:
Closing this for #401 as it seems to be a better approach forward than this.

PR #398, opened 2024-12-16
Parse exits from banner components

ahmedre:
This is a follow up to #391, which parses out the exit numbers from the banner components. Note that this explicitly ignores all types other than "exit-number" for now.


common/ferrostar/src/routing_adapters/osrm/snapshots/ferrostar__routing_adapters__osrm__tests__parse_osrm_with_exits.snap
```diff
@@ -129,6 +128,8 @@ snapshot_kind: text
             maneuver_modifier: slight right
             roundabout_exit_degrees: ~
             lane_info: ~
+            exit_numbers:
+              - "10"
```
ahmedre, line 132: this is the result of the fix. given the input in the sample Valhalla response and others:

```json
"primary": {
                    "type": "off ramp",
                    "modifier": "slight right",
                    "text": "OR 213 South: Oregon City, Molalla",
                    "components": [
                      {
                        "text": "Exit",
                        "type": "exit"
                      },
                      {
                        "text": "10",
                        "type": "exit-number"
                      },
                      {
                        "text": "OR 213 South: Oregon City, Molalla",
                        "type": "text"
                      },
                      {
                        "text": "/",
                        "type": "delimiter"
                      },
                      {
                        "text": "OR 213 South",
                        "type": "text"
                      }
                    ]
                  },
```

this only parses out `exit-number` fields. note that, while I have only seen this in `primary` components in both the Valhalla api responses and Mapbox's API responses, it's theoretically possible in all, so it's returned from all of them.

@ianthetechie do you think this is ok, or should we keep it primary only? if this is okay, should we update `lanes` to not be `sub` only?

ianthetechie: I've also only seen this in primary, so I guess it's fine as-is for now. We can always revisit later if necessary.

common/ferrostar/src/models.rs
```diff
@@ -633,6 +633,8 @@ pub struct VisualInstructionContent {
     pub roundabout_exit_degrees: Option<u16>,
     /// Detailed information about the lanes. This is typically only present in sub-maneuver instructions.
     pub lane_info: Option<Vec<LaneInfo>>,
+    /// Detailed information about exits.
```
ianthetechie, line 636:
```suggestion
    /// The exit number (or similar identifier like "8B").
```

ianthetechie (approved):

ahmedre:
I had to manually update and force push to re-generate the `ferrostar.swift` file after the comment change. I couldn't figure out how to pull it with your suggestion merged so manually re-did it - git pull --rebase didn't get the changes

ianthetechie:
huh weird.... 🤷

Issue #425, opened 2025-01-14
Documentation for react native


bjtrounson:
Related to #401.

Having some documentation on the for the React Native implementation is need and example app. 

Some testing to see if the generated Uniffi bindings still work when the library is packaged still need to happen.

ianthetechie:
This should probably be two separate issues: one for docs and another for publishing to npm ;)

PR #436, opened 2025-01-31
Change package.json for React Native to allow for optional deps for views

bjtrounson:
This is a small PR and related to #425 

I've been wondering about a way to properly do this within the package.json and was looking through some other projects and found the way https://github.com/computerjazz/react-native-draggable-flatlist/blob/main/package.json handles this. 

It seems peerDependencies seems the way to go, look through the documentation on NPM and it seems we can label some of the peer deps as optional which I think will work well for the view related packages (React Native MapLibre and so on.)

Documentation if anyone else want to have a look in case I've missed something https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependencies

This PR just makes that change. Another way to potential do this is to separate the view and core logic into to two separate package which we upload separately onto NPM

### DIFF
--- a/react-native/package.json
+++ b/react-native/package.json
@@ -92,14 +92,33 @@
     "react-native-builder-bob": "^0.32.0",
     "release-it": "^17.10.0",
     "turbo": "^1.10.7",
-    "typescript": "^5.2.2"
+    "typescript": "^5.2.2",
+    "@maplibre/maplibre-react-native": "^10.0.0",
+    "@turf/bbox": "^7.2.0",
+    "react-native-svg": "^15.10.1",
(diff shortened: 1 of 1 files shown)


ianthetechie (approved):
This makes sense to me! Seems like a reasonable first step at least.

I haven't had much chance to think through the npm packaging of the React Native module yet, but this is an interesting idea, since you can explicitly denote optional dependencies.

I think the main argument for separating the packages is that users wouldn't need to download code which they won't use. Is that worth the overhead of maintaining multiple npm projects in the workspace? I'm actually not sure, but it does seem to make things significantly more annoying to work with.

Issue #437, opened 2025-01-31
Route Polyline Configuration


ben-burwood:
Just wanted to gauge some interest before starting work on a PR...

The BorderedPolyline implementation currently has no adaptability. For example, in our use case, we often have a map matched route in residential areas, effectively drawing a line on every road on the map, whilst navigating this makes it hard to the user to distinguish 'completed' vs remaining sections of the route.
Customisation for this could be to set the travelled section of the route to vanish, fade out, thin the line, etc...

Similarly to how the `VisualNavigationViewConfig` works, we could extract a RoutePolyline Builder, with reference to the NavigationUiState to the top level of the NavigationViews to achieve this - obviously this is Android specific but I'm sure parity could be kept with iOS in a similar implementation?

By defaulting to the current implementation, this could be implemented without breaking changes.

Would be interested if anyone has any concerns with this idea

ianthetechie:
Short answer: yes, there is interest!

Longer answer: there is a pretty long running discussion within MapLibre (it ends up being a near recurring agenda item) about how we can improve this at the renderer level. Today it would be possible to render two lines (one for the full route, and another on top for what's left). This unfortunately isn't optimal but it can probably be made to work. Ideally we'd like to see deeper support for this at the style spec and then renderer level to do this efficiently and at a higher refresh rate directly in MapLibre. 

In the interim though, I think we'd welcome PRs with alternate approaches that work "well enough" (if you have any ideas besides the multi layer approach, let me know)!

cc @Archdoog in case you have anything to add, and I think @nyurik was going to do a write up somewhere to get a discussion going on how to do this in the style spec.

Also don't worry too much about breaking changes. We are not 1.0 and everyone using it at this stage is OK with breaking changes as long as we are improving things :)

ianthetechie:
btw send me a message if you're interested to join our weekly dev meeting on Tuesday (02:00 UTC; not great if you're in Europe but works well for US and Asia)

ben-burwood:
Thought that might be the case r.e. MapLibre!

For an interim solution, I have a working POC (just Android currently) with a new injected _NavigationPathBuilder_ instead of constructing the BorderedPolyline directly in the MapView. This feels quite adaptable and leaves the default behaviour unchanged, but allowing me to override the view with multiple lines in different states.
I'll open a draft PR to get some thoughts on the approach over the weekend before firming up impl for the other platforms.

Ah, shame I'm UK based so that wouldn't work

Issue #361, opened 2024-11-15
Puck lag mitigation


ianthetechie:
As things stand today, the puck can lag a bit behind the user's current location. We've added a little dot that updates in real time with the user's location in both demo apps as a visualization of this (and to enable debugging of route snapping vs GPS drift). Here is a brain dump of what I know so far.

# Underlying problems

A few underlying problems contribute to the observed lag.

## (In)frequency of location updates

Most mobile devices will deliver a location update approximately once per second. This is not fast enough to create a smooth motion without some sort of additional work.

## Camera animation

The MapLibre camera mode (track user with bearing) that draws the puck attempts to address this infrequency of updates with an animation. This sounds like a great idea in theory, but it means the user's location may be up to 1 second ahead of the puck. Or at least, that's what the _intent_ of the MapLibre code is, as I understand it.

There are some bugs in this animation code though. I tried looking at it myself about 6 months ago, but wasn't able to find a glaringly obvious issue.

### Fast updates cause problems

On iOS in particular (I forget if Android exhibits this), rapid updates to the user's location actually make the problem significantly worse. I believe this is because it constantly interrupts the animation with a new one, and given how easing etc. work, it actually makes it harder to catch up. If you deliver updates twice/second, for example, you would have a persistent lag that's even longer. And if your update timing is variable, the lag will not be consistent either.

**Root cause 1**: Since updates are typically not delivered at a frequency of > 1Hz, the user's location will lag behind by definition.

**Root cause 2**: in my analysis, something is subtly wrong with how MapLibre animates its camera, preventing delivery of sub-second location estimates (ex: Kalman filters, external vehicle sensors, etc.)

## (Related) Inaccuracy of location updates

It's pretty regular that the device confidently reports very wrong GPS data. I think that some amount of filtering is going to be required by many applications. iOS is interesting in particular, because the data you get is already smoothed in some way. Android has a mediocre fused location client, and an extremely rough location API, but at least it does let you get more "raw" data which is more amenable to building corrective layers on top of.

This isn't a direct cause of puck lag normally, but is a related challenge that will need to be considered.

# Approach to getting better tracking

[...]

PR #439, opened 2025-01-31
Route Polyline Configuration

ben-burwood:
As discussed in #437, I have put together a minimal change to allow overriding of the current BorderedPolyline implementation for displaying the Route on the Map.
This allows for a custom NavigationPathBuilder _(not precious on the name...)_ to be injected into the NavigationView and be re-composed with changes to the UiState.

Found that having currentStepGeometryIndex exposed in the NavigationUiState was needed for the tracking in my implementation so have added this.

This is just Android for now, happy to work on implementation on the other platforms to work in a similar way to this if others would fine this useful.

### DIFF
--- a/android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/routeline/NavigationPathBuilder.kt
+++ b/android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/routeline/NavigationPathBuilder.kt
@@ -1,0 +1,23 @@
+package com.stadiamaps.ferrostar.maplibreui.routeline
+
+import androidx.compose.runtime.Composable
+import com.mapbox.mapboxsdk.geometry.LatLng
+import com.maplibre.compose.ramani.MapLibreComposable
+import com.stadiamaps.ferrostar.core.NavigationUiState
+
+data class NavigationPathBuilder(
+    internal val navigationPath:
+        @Composable
+        @MapLibreComposable
+        (uiState: NavigationUiState) -> Unit
+) {
+  companion object {
+    fun Default() =
+        NavigationPathBuilder(
+            navigationPath = { uiState ->
+              uiState.routeGeometry?.let { geometry ->
+                BorderedPolyline(points = geometry.map { LatLng(it.lat, it.lng) }, zIndex = 0)
+              }
+            })
+  }
+}
--- a/android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
+++ b/android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
@@ -58,8 +58,10 @@ data class NavigationUiState(
     val isMuted: Boolean?,
     /** The name of the road which the current route step is traversing. */
     val currentStepRoadName: String?,
-    /** The index of the closest coordinate to the user's snapped location.
-     *  The index is Relative to the *current* (i.e. first in remainingSteps) RouteStep Geometry */
+    /**
+     * The index of the closest coordinate to the user's snapped location. The index is Relative to
+     * the *current* (i.e. first in remainingSteps) RouteStep Geometry
+     */
     val currentStepGeometryIndex: Int?,
     /** The remaining steps in the trip (including the current step). */
     val remainingSteps: List<RouteStep>?,
--- a/android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
+++ b/android/core/src/main/java/com/stadiamaps/ferrostar/core/NavigationViewModel.kt
@@ -7,6 +7,7 @@ import com.stadiamaps.ferrostar.core.annotation.AnnotationPublisher
 import com.stadiamaps.ferrostar.core.annotation.AnnotationWrapper
 import com.stadiamaps.ferrostar.core.annotation.NoOpAnnotationPublisher
 import com.stadiamaps.ferrostar.core.extensions.currentRoadName
+import com.stadiamaps.ferrostar.core.extensions.currentStepGeometryIndex
 import com.stadiamaps.ferrostar.core.extensions.deviation
 import com.stadiamaps.ferrostar.core.extensions.progress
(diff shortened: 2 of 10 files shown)


ben-burwood:
### Implementation Example 

The line is fairly chunky due to the known lag in MapLibres location tracking, also longer geometry sections may stay on the screen after the Location has passed them... so by no means a perfect implementation...

https://github.com/user-attachments/assets/7ad489ad-ecec-437b-b3f6-3e475e694766

```kotlin
val CustomNavigationPathBuilder =
    NavigationPathBuilder(
        navigationPath = { uiState: NavigationUiState ->
            val remainingGeometry = uiState.remainingSteps?.flatMap { it.geometry } ?: uiState.routeGeometry

            val currentGeometryIndex = uiState.currentStepGeometryIndex?.let {
                uiState.remainingSteps?.first()?.let { currentStep ->
                    val currentGeometry = currentStep.geometry[it]
                    remainingGeometry?.indexOfFirst { it == currentGeometry}
                }
            } ?: 0

            // Render Remaining Route Line
            val remainingRouteLine = remainingGeometry.let {
                if (uiState.isNavigating()) it?.subList(currentGeometryIndex, it.size)
                else it
            }?.map { LatLng(it.lat, it.lng) }
            remainingRouteLine?.let {
                BorderedPolyline(points = it, zIndex = 1)
            }

            // Render Complete Route Line
            val completeRouteLine = uiState.routeGeometry?.map { LatLng(it.lat, it.lng) }
            completeRouteLine?.let {
                BorderedPolyline(points = it, zIndex = 0, lineWidth = 5f, color="#999999")
            }
        })
```

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/routeline/NavigationPathBuilder.kt
```diff
@@ -0,0 +1,23 @@
...
+) {
+  companion object {
+    fun Default() =
+        NavigationPathBuilder(
+            navigationPath = { uiState ->
+              uiState.routeGeometry?.let { geometry ->
+                BorderedPolyline(points = geometry.map { LatLng(it.lat, it.lng) }, zIndex = 0)
+              }
+            })
+  }
```
ianthetechie, line 22: Ahh, that's clever. So basically the idea if I understand correctly is that the "default" doesn't change current behavior, but if you want to experiment with other drawing styles (including but not limited to "route eating") then you have the freedom to do so?

ianthetechie (commented):
Cool! Just one initial question to make sure I understand the intent of the PR (I was initially drawn to the route eating demo and then realized that probably wasn't the point haha).

I'll do a bit of thinking on names today + make some suggestions on how we name and/or document it, but I think this is probably mergeable with minimal effort if I got it right.

# Misc thoughts on "route eating"

I assume you've already seen https://github.com/stadiamaps/ferrostar/issues/361. There are a few challenges that make "route eating" really complicated ;)

As you've already observed, I think it'll be tricky to update the "remaining" polyline. You could perhaps try to update it to an interpolated value halfway between the previous update and the current position? This would have some problems obviously, but _maybe_ it's good enough?

Modifying the sources faster (say 10x/sec) with interpolated values to try and "skate to where the puck is going" _could_ work. You'd basically be making a bunch of small updates interpolating like the previous paragraph, but we'd need to keep an eye on whether this impacts performance.

Longer term we want to see MapLibre gain support for "high frequency" location updates, which could drive such interpolated updates. These are linked issues because high frequency estimated location updates will improve UI smoothness, reduce lag, and be part of a real high quality "route eating" polyline display.

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/routeline/NavigationPathBuilder.kt
ben-burwood: Yeah exactly that. I've tried to keep it general so a user can take the same approach whether they just want to override the appearance of the line (color, width), or _attempt_ to override functionality (as per my example).

ben-burwood:
Completely right on the 'route eating', spent a little while trying to improve it over the weekend but to no avail - so thanks for your write up so I know it's not just me!

Having said that, even with it's current janky implementation, the changes in this PR are still beneficial for our use case and something that we would likely use in some form - the other option I considered was only showing the next x steps on the route instead of the complete route - this also works quite well for us in cleaning up the map.

android/maplibreui/src/main/java/com/stadiamaps/ferrostar/maplibreui/routeline/NavigationPathBuilder.kt
```diff
@@ -0,0 +1,23 @@
+package com.stadiamaps.ferrostar.maplibreui.routeline
+
+import androidx.compose.runtime.Composable
+import com.mapbox.mapboxsdk.geometry.LatLng
+import com.maplibre.compose.ramani.MapLibreComposable
+import com.stadiamaps.ferrostar.core.NavigationUiState
+
+data class NavigationPathBuilder(
```
ianthetechie, line 8: Thoughts on renaming to `PathOverlayBuilder`? I think this communicates what it does a bit better. We should probably also have a docstring too ;)

cc @Archdoog

ianthetechie: Discussed on the call today and @Archdoog came up with `RouteOverlayBuilder` to keep it consistent with how we talk about this elsewhere in the navigation stack.

ben-burwood:
Hi both,
Thanks for supporting in getting this through :)
I've also looked into implementing this on iOS, but can see that a rough similar implementation is already in place with the RouteStyleLayers. I'd be happy to implement it equivalent to Android with injecting all of this into the NavigationMapView, but perhaps you want to keep some parts of the current implementation? Anyway let me know if I can help on this, else I'll  leave it with you.

ianthetechie (approved):
LGTM; thanks!

Issue #452, opened 2025-02-09
Android TextToSpeech Repeated Voice Instruction Queue Bug - Infinite Loop


Archdoog:
When navigating on android, there are certain cases where the Android TextToSpeech wrapper starts reciting the same instruction after a re-route. Given the timing after re-route, this tends to be the initial "Depart" or "Continue on Something Rd." instruction. Notes about what happens:

1. For the first (or first couple of) occurrences (e.g. multiple re-routes) it will repeat the instruction several times, but the queue will eventually clear and appear to resume somewhat normal service.
2. After many occurrences, the instruction will be recited in an infinite (or near infinite) loop. Just saying the first instruction over and over again for the entire session until you either: mute, end navigation or _funnily enough rotate the screen._

After this happens, even if you un-mute the next recitation will kick off a new cascade of repeated voice instructions. Given that we've taken some passes at this bugs, I'm starting to wonder if something's leaking, there's a thread safety issue. This guess is driven by the fact that initially it will self resolve, but later in the same session it often cascades out of control to an infinite-ish loop. Theory is if we're leaking something, we may have a listener that's running voice 1x per cycle, then 2x, 4x, 8x, and so on as more callers leak.

ianthetechie:
Out of curiosity, did you encounter this around an off route recalculation? 

I actually just thought of a few ways we can probably improve that and ensure users don't accidentally get in a loop making a bunch of useless reroute requests. I should be able to open a PR for that tomorrow. cc @ben-burwood since you may have hit one of those cases inadvertently too.

ianthetechie:
Some notes after a look at this on the call today...

It seems that it is indeed happening around recalculation events anecdotally per both @Archdoog and @bjtrounson. We looked at the code and think there are possible race conditions in `handleStateUpdate` if the method is called from multiple threads. We haven't yet identified whether this is actually what's happening, but we should take steps to ensure this can't happen.

Additionally, we should probably store the `Job` returned by `_scope.launch`, as we store the task on iOS.

Finally, we noted that while we haven't yet seen this on iOS, there is definitely a possibility that the same bug exist, especially given that both implementations are as structurally similar as the platform APIs allow.

bjtrounson:
As mentioned in our call today here is the recording of the behavior I got while mocking location updates down a route. 

https://github.com/user-attachments/assets/18196e50-4fe5-4e7e-a460-287ae33f18b4

[...]

PR #453, opened 2025-02-10
Recalculation gating

ianthetechie:
This adds a new gate to prevent spurious recalculations. We'll avoid recalculating unless the user has moved by at least 50m (configurable threshold) from the *last* point that we triggered a recalculation


ianthetechie:
cc @ben-burwood

ben-burwood:
Funnily enough this is the exact thing I have been trialling in our app since we fell into that spiral last week! So definitely support this functionality being included in the SDK. Implementation as you have here looks good 👍

Archdoog:
This is also likely part of the problem with #452. Many rapid re-routes blowing up the TTS queue.

PR #457, opened 2025-02-11
feat: separate waypoint logic

Archdoog:
- Separates waypoint advance logic out of step advance fn. This reduces the conditional burden required to advance a waypoint. It also enables flexible waypoint advance logic.


common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -169,12 +169,24 @@ pub enum SpecialAdvanceConditions {
     MinimumDistanceFromCurrentStepLine(u16),
 }
 
+// Condition to advance the remaining waypoints.
```
ianthetechie, line 172:
```suggestion
/// Controls when a waypoint should be marked as complete.
///
/// While a route may consist of thousands of points, waypoints are special.
/// A simple trip will have only one waypoint: the final destination.
/// A more complex trip may have several intermediate stops.
/// Just as the navigation state keeps track of which steps remain in the route,
/// it also tracks which waypoints are still remaining.
///
/// Tracking waypoints enables Ferrostar to reroute users when they stray off the route line.
/// The waypoint advance mode specifies how the framework decides
/// that a waypoint has been visited (and is removed from the list).
///
/// NOTE: Advancing to the next *step* and advancing to the next *waypoint*
/// are separate processes.
/// This will not normally cause any issues, but keep in mind that
/// manually advancing to the next step does not *necessarily* imply
/// that the waypoint will be marked as complete!
```

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -215,6 +194,16 @@ impl NavigationController {
                     &current_step_linestring,
                     remaining_steps,
                 );
+
+                // Trim the remaining waypoints if we should advance.
```
ianthetechie, line 198:
```suggestion
                // Trim the remaining waypoints if needed.
```

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -336,6 +325,31 @@ impl NavigationController {
...
+                    let current_location: Point = snapped_user_location.coordinates.into();
+                    let next_waypoint: Point = waypoint.coordinate.into();
+                    match self.config.waypoint_advance {
+                        WaypointAdvanceMode::WaypointWithinRange(range) => {
+                            Haversine::distance(current_location, next_waypoint) < range
+                        }
+                    }
+                } else {
+                    false
+                }
```
ianthetechie, line 348: Nice work! I think you can make this even more compact with the [`is_some_and`](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and) like this:

```rust
remaining_waypoints.first().is_some_and(|waypoint| {
    // Rest of the inner code
}
```

Then you can drop the whole if/else wrapper and it's a single expression inside :)

Archdoog:
Discussion for testing. Here's a debug dump:

```
Distance: 122.67532720196205
Should advance? with [Waypoint { coordinate: GeographicCoordinate { lat: 40.777985, lng: -74.040048 }, kind: Break }]
Should advance? user Point(Coord { x: -74.040718, y: 40.777033 }), next_waypoint [Waypoint { coordinate: GeographicCoordinate { lat: 40.777985, lng: -74.040048 }, kind: Break }]
Distance: 119.95247025652922
Should advance? with []
Should advance? with []
Should advance? with []
```

ianthetechie:
hmm... looks like the implementation broke something then? I would expect no changes in the test results with the given settings.

ianthetechie:
Also, given the fundamental nature of the changes, shouldn't there also be some TypeScript changes in here?

ianthetechie (approved):

PR #460, opened 2025-02-18
fix: clear utterance queue on replace route

Archdoog:
- Clears utterance queue on replace route in android core.

Potential fix for #452


ianthetechie (approved):
Good catch 👏

PR #461, opened 2025-02-18
Feat: Advanced Step Advance

Archdoog:
Introduces a customizable stateful capable step advance system. This allows highly configurable step advance modes.

Historically, there were multiple step advance modes that worked well for most cases, but had some minor side effects or failure modes. Some of these included: 

- The puck appearing to make the turn before the user went through a step, 
- The navigation controller getting stuck if the user skipped a step entirely out of cellular data.
- Other slightly eager or sluggish responses to events. E.g. forcing a re-route on under/overpasses.

This system allows both the construction of dynamic stateful step advance modes, as well as logic and/or configurations. 

An example use case could be an improved step advance mode like the DistanceEntryAndExit condition, with an OR for short circuit if the user is no longer anywhere near the step.


common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -554,11 +605,8 @@ mod tests {
     fn test_extended_relative_linestring() {
```
Archdoog, line 549: This test (and a couple of others) are failing w/ 

```
stored new snapshot /Users/jacob/Git/ferrostar/common/ferrostar/src/navigation_controller/snapshots/ferrostar__navigation_controller__tests__extended_relative_linestring.snap.new
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.

thread 'navigation_controller::tests::test_extended_relative_linestring' panicked at /Users/jacob/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/insta-1.43.1/src/runtime.rs:679:13:
snapshot assertion for 'extended_relative_linestring' failed in line 607
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The issue occurs at the end of the route when we need to exit early because step advance modes are no longer necessary (our user is snapped to the last location of the route - aka _the last point of the 2nd to last step_)

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -530,9 +574,16 @@ mod tests {
             simulation_state = new_simulation_state;
             state = new_state.clone();
             states.push(new_state);
+
+            // Exit if there are no more locations.
+            // TODO: This test simulation doesn't let us complete the actual route for
+            // step_advance_conditions that require the user to exit the step (e.g. DistanceEntryAndExitCondition)
+            if simulation_state.remaining_locations.is_empty() {
```
Archdoog: This allows the tests to avoid timeout when the panic isn't operable.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -395,13 +417,27 @@ impl NavigationController {
...
-            user_location,
-            summary,
-        }
+    /// Checks if the user is at the end of the route based on their snapped position.
+    ///
+    /// Returns `true` if the user is snapped to the last coordinate of the second-to-last step,
+    /// indicating they have effectively completed the route. We can't check the last step, because
+    /// it is an arrival step. Which is essentially single point step that copies of last point
+    /// of the 2nd to last step.
+    fn is_at_end_of_route(
```
Archdoog: This function probably isn't ideal. But we need an exit early for route completion.

Several of the step advance modes only work on the logic that the user actually passes the end point. Which isn't true in testing and might not be true IRL.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -246,25 +258,32 @@ impl NavigationController {
...
-                    &current_step_linestring,
-                    remaining_steps.get(1),
-                    &location,
-                    self.config.step_advance,
-                ) {
+                let intermediate_nav_state =
+                    NavState::new(intermediate_state, step_advance_result.next_iteration);
+
+                // Exit early if the user is snapped to the last coordinate in the route.
+                if Self::is_at_end_of_route(&remaining_steps, current_step_geometry_index) {
```
Archdoog: Here's the exit early.

common/ferrostar/src/navigation_controller/step_advance/conditions.rs
```diff
@@ -0,0 +1,685 @@
...
+        let user_location_far = user_location(0.0005, 0.001, 5.0);
+
+        // Now should advance because we've satisfied both conditions sequentially
+        let result2 = next_condition.should_advance_step(user_location_far, route_step, None);
+
+        assert!(
+            result2.should_advance,
+            "Should advance when user has first reached end of step and then moved away"
+        );
+    }
```
Archdoog, line 758: Write a 3rd test that iterates the part 2 of `test_entry_and_exit_condition_advance` twice to make sure it holds the true boolean when it fails to advance for result2.

common/ferrostar/src/simulation.rs
```diff
@@ -118,7 +118,7 @@ pub enum LocationBias {
 #[cfg_attr(feature = "wasm-bindgen", tsify(into_wasm_abi, from_wasm_abi))]
 pub struct LocationSimulationState {
     pub current_location: UserLocation,
-    remaining_locations: Vec<GeographicCoordinate>,
+    pub(crate) remaining_locations: Vec<GeographicCoordinate>,
```
Archdoog: Convert this to an impl of `simulation.has_reached_end()` or similar?

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -8,12 +10,101 @@ use crate::models::{
...
+#[cfg(feature = "wasm-bindgen")]
+impl From<NavState> for JsNavState {
+    fn from(value: NavState) -> Self {
+        Self {
+            trip_state: value.trip_state,
+            step_advance_condition: value.step_advance_condition.to_js(),
+        }
+    }
+}
+
```
Kandarp05, line 112: You don't need to do this, just add
```
#[cfg_attr(feature = "wasm-bindgen", tsify(into_wasm_abi, from_wasm_abi))]
```
to NavState and it does the conversion for you

refer to #619

ianthetechie: I don't think so, but maybe I've missed something. Your `NavState` is constructed from types that are directly representable. The step advance condition is not; we are only creating a bridge that enables representing a subset of possible states in JS.

android/demo-app/src/main/java/com/stadiamaps/ferrostar/AppModule.kt
```diff
@@ -90,12 +90,8 @@ object AppModule {
...
                 NavigationControllerConfig(
                     WaypointAdvanceMode.WaypointWithinRange(100.0),
-                    StepAdvanceMode.RelativeLineStringDistance(
-                        minimumHorizontalAccuracy = 25U,
-                        specialAdvanceConditions =
-                            // NOTE: We have not yet put this threshold through extensive real-world
-                            // testing
-                            SpecialAdvanceConditions.MinimumDistanceFromCurrentStepLine(10U)),
+                    stepAdvanceDistanceEntryAndExit(10u, 5u, 32u),
+                    stepAdvanceDistanceToEndOfStep(10u, 32u),
```
ianthetechie: To properly round this out, we'll need to update the following in the book:

* Android quick start
* iOS quick start
* Web quick start
* Add a new page under customization discussing step advance in slightly more detail

apple/Tests/FerrostarCoreTests/FerrostarCoreTests.swift
```diff
@@ -124,7 +124,8 @@ final class FerrostarCoreTests: XCTestCase {
             locationProvider: SimulatedLocationProvider(),
             navigationControllerConfig: .init(
                 waypointAdvance: .waypointWithinRange(100.0),
-                stepAdvance: .manual,
+                stepAdvanceCondition: stepAdvanceManual(),
```
ianthetechie, line 127: Bookmarking here as well as example code for the docs updates.

common/ferrostar/src/algorithms.rs
```diff
@@ -567,116 +443,6 @@ proptest! {
         }
     }
 
-    #[test]
```
ianthetechie, line 570: It looks like these property tests disappeared entirely, which is a bummer...

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -208,72 +152,41 @@ impl Navigator for NavigationController {
...
+                        NavState::apply_complete(user_location, summary.clone())
                     }
                 }
             }
-            TripState::Complete {
-                user_location,
-                summary,
-            } => Self::completed_trip_state(*user_location, summary.clone()),
+            // Pass through any other states.
+            _ => state.clone(),
```
ianthetechie: Generally we should prefer explicit exhaustive matching of match statements. The complier can then yell at us (politely) when we add a new variant and don't explicitly handle it.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -283,38 +196,112 @@ impl Navigator for NavigationController {
...
+    /// - `location`: The user's current location.
+    /// - `current_step`: The current route step.
+    /// - `remaining_steps`: The remaining route steps.
+    /// - `remaining_waypoints`: The remaining waypoints.
+    ///
+    /// Returns:
+    /// - `TripState`: The intermediate trip state.
+    fn create_intermediate_trip_state(
+        &self,
+        trip_state: &TripState,
```
ianthetechie: I'm changing the intermediate type of this to take an owned value. The intent is to "replace" the old value, so this seems appropriate and enforces that we don't accidentally use a stale value we held on to.

common/ferrostar/src/navigation_controller/models.rs
Kandarp05: Oh I see, my bad

common/ferrostar/src/navigation_controller/step_advance/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct StepAdvanceResult {
+    /// The step should be advanced.
+    pub should_advance: bool,
+    /// The next iteration of the step advance condition.
+    /// This allows us to copy the condition and its current state
+    /// to the next user location update/next interaction of the step
+    /// advance calculation.
+    ///
+    /// IMPORTANT! If the condition advances. This **must** be the clean/default state.
```
ianthetechie, line 36: Can you clean up / clarify this?

common/ferrostar/src/navigation_controller/step_advance/mod.rs
```diff
@@ -0,0 +1,188 @@
+use crate::{
```
ianthetechie, line 2: I think this module should probably have its own docstring `//!` and more of the types need at least basic docs since this is a non-trivial part of the public API now.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -533,7 +505,14 @@ mod tests {
             let new_state =
                 controller.update_user_location(new_simulation_state.current_location, &state);
 
-            match new_state {
+            // // Exit if there are no more locations.
+            // // TODO: This test simulation doesn't let us complete the actual route for
+            // // step_advance_conditions that require the user to exit the step (e.g. DistanceEntryAndExitCondition)
+            // if simulation_state.remaining_locations.is_empty() {
+            //     break;
+            // }
```
ianthetechie: Is this still TODO?

web/index.html
```diff
@@ -68,12 +68,18 @@
 
       // TODO: type + use TypeScript enum
       const config = {
-        stepAdvance: {
-          RelativeLineStringDistance: {
+        stepAdvanceCondition: {
```
ianthetechie, line 71: And another bookmark for JS documentation updates in the book.

common/ferrostar/src/navigation_controller/step_advance/conditions.rs
```diff
@@ -0,0 +1,759 @@
+use std::sync::Arc;
```
ianthetechie, line 1: I've removed the mark comments in this file. If you think that these should stay around for organization, I'd suggest we rather split into smaller submodule files instead. This is better for both organization and discovery.

common/ferrostar/src/navigation_controller/step_advance/conditions.rs
```diff
@@ -0,0 +1,759 @@
...
+        JsStepAdvanceCondition::DistanceEntryExit {
+            minimum_horizontal_accuracy: self.minimum_horizontal_accuracy,
+            distance_to_end_of_step: self.distance_to_end_of_step,
+            distance_after_end_step: self.distance_after_end_step,
+            has_reached_end_of_current_step: self.has_reached_end_of_current_step,
+        }
+    }
+}
+
+#[cfg(test)]
```
ianthetechie, line 369: We probably should have some property tests here? Those tend to surface subtle bugs ;)

ianthetechie (commented):
Phew... almost there ;)

common/ferrostar/src/navigation_controller/mod.rs
Archdoog: Removed

android/demo-app/src/main/java/com/stadiamaps/ferrostar/AppModule.kt
Archdoog: Added guide items 👍

common/ferrostar/src/navigation_controller/step_advance/mod.rs
Archdoog: Rewrote this to better describe what it's for/how it works.

common/ferrostar/src/navigation_controller/mod.rs
Archdoog: Removed

common/ferrostar/src/navigation_controller/step_advance/mod.rs
```diff
@@ -1,3 +1,4 @@
+//! Step advance condition traits and implementations
```
ianthetechie, line 1:
```suggestion
//! Step advance condition traits and implementations.
```

web/index.html
```diff
@@ -68,12 +68,18 @@
...
+          DistanceEntryExit: {
+            minimumHorizontalAccuracy: 25,
+            distanceToEndOfStep: 10,
+            distanceAfterEndStep: 5,
+            hasReachedEndOfCurrentStep: false,
+          },
+        },
+        arrivalStepAdvanceCondition: {
+          DistanceToEndOfStep: {
+            distance: 10,
```
Archdoog:
```suggestion
            distanceToEndOfStep: 30,
            distanceAfterEndStep: 5,
            hasReachedEndOfCurrentStep: false,
          },
        },
        arrivalStepAdvanceCondition: {
          DistanceToEndOfStep: {
            distance: 30,
```

Archdoog (commented):
One missed one.

PR #475, opened 2025-03-03
On-Trip-State-Change for Web Components

tedi4t:


igal1c0de4n (approved):
looks ok on my end

ianthetechie (changes requested):
This change looks fine. Two quick things:

1. I don't quite understand all of the re-binding of `this` to `this`. I'm vaguely aware that JavaScript has a rather counterintuitive way of redefining `this` in many contexts. Was this a bug in how we did it in just this one place, or is it something we should also have a look over elsewhere?

2. I tried to fix the style checker issues but got an error when trying to push to your branch. GitHub claims that maintainers can edit the PR, but unless this is a recent change, this is impossible on org-owned forks. Just run `npm format:fix` and push :)

tedi4t:
@ianthetechie PR is updated. You are right, we don't need bind here, my bad

ianthetechie (approved):
Looks good; thanks!

Issue #427, opened 2025-01-21
Sync up react native and web formatting


bjtrounson:
I've created some custom formatter e.g.  for React Native that could be used for the web component library. 

We probably need to create a separate package with these formatters in them that both projects and depend on.

PR #481, opened 2025-03-04
Web distance and duration formatters

bjtrounson:
Pretty simple PR here I've ripped out the formatters used in react native into a separate package here https://www.npmjs.com/package/@maptimy/platform-formatters

This PR just changes the trip-progress-view.ts and instructions-view.ts to use the new formatters. 

This is one step to resolving this issue #427 we then need to sync up the icons.


web/package-lock.json
ianthetechie, line 1: That's.... a large diff 😅 I looked at the package code over at https://github.com/maptimy/platform-formatters/blob/main/js/package.json and it looked fine, but I wonder if the issue is that npm doesn't know how to deal with platform-specific dependencies or something.

bjtrounson: Oh woah, didn't actually notice this large diff. I wondering if it's install react-native and it's peer deps since I have it labeled as a peer dependency. 

I could try removing it and seeing how it goes.

bjtrounson: @ianthetechie I've just done a version that has the peer dep meta set as optional. It seems to still add the react-native related packages to the lock file but flags them all as optional and **doesn't** install them into the node_modules directory.

ianthetechie:
Should be good now; will merge once CI passes

PR #482, opened 2025-03-04
Add a route border under route for web

bjtrounson:
Another simple PR for web that just adds a nice white border under the route line.

I've also changes the route color to be less harsh blue like the current one is, I can revert that one if need be. We could find the route color of the Android/iOS version and use that color so it's synced.

### DIFF
--- a/web/src/ferrostar-map.ts
+++ b/web/src/ferrostar-map.ts
@@ -347,10 +347,27 @@ export class FerrostarMap extends LitElement {
         "line-cap": "round",
       },
       paint: {
-        "line-color": "#3700B3",
+        "line-color": "#0099ff",
         "line-width": 8,
       },
     });
+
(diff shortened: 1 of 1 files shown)


web/src/ferrostar-map.ts
```diff
@@ -347,11 +347,28 @@ export class FerrostarMap extends LitElement {
         "line-cap": "round",
       },
       paint: {
-        "line-color": "#3700B3",
+        "line-color": "#0099ff",
```
ianthetechie, line 350: Found the color from iOS using a screen color meter:

```suggestion
        "line-color": "#3478f6",
```

ianthetechie (approved):
Thanks; this looks great! I went and found the blue color from iOS, which looks a bit better IMO.

PR #486, opened 2025-03-11
Add adaptive distance calculation algorithm with dynamic Haversine/Geodesic switching

### LINKED ISSUE #329: Evaluate conditions for preferring geodesics over haversine
Currently most of the distance calculations in Ferrostar are done with the haversine formula. This makes some simplifying assumptions, but it is widely used for its speed. We should probably keep doing this.

However, for cases where there could be a larger accumulation of errors, perhaps we should be using geodesics. An example of this is where one or more calculations add up to a trans-continental distance (as they might when computing the length of route sections).

We never really evaluated acceptable uses for each, and currently use a mix of both (partially due to bugs we found, such as panics on direct intersections in geo rust). We should research + standardize on this going forward.

Kandarp05:
This PR introduces an adaptive distance calculation approach that dynamically switches between Haversine and Geodesic formulas based on geographical conditions.

Fixes #329 

**Implementation approach:**
- Evidence-based algorithm accumulates factors from latitude and distance
- Defaults to faster Haversine formula for most cases
- Switches to Geodesic when needed (high latitudes, long distances)
- Uses decay mechanism to prevent algorithm thrashing
- Constants organized in typed structs for better maintainability

**Testing**
- Added comprehensive tests for basic functionality
- Edge cases covered (near poles, date line crossing)
- Stability tests to verify appropriate switching behavior

Let me know if this approach makes sense or if you'd prefer a different direction for the algorithm's design.

### DIFF
--- a/common/ferrostar/src/algorithms.rs
+++ b/common/ferrostar/src/algorithms.rs
@@ -12,8 +12,7 @@ use crate::{
     navigation_controller::models::{SpecialAdvanceConditions, TripProgress},
 };
 use geo::{
-    Bearing, Closest, Coord, Distance, Euclidean, Geodesic, Haversine, HaversineClosestPoint,
-    Length, LineLocatePoint, LineString, Point,
+    Bearing, Closest, Coord, Distance, Euclidean, Geodesic, Haversine, HaversineClosestPoint, Length, LineLocatePoint, LineString, Point
 };
 
 #[cfg(test)]
@@ -28,17 +27,11 @@ use std::time::SystemTime;
 #[cfg(all(test, feature = "web-time"))]
 use web_time::SystemTime;
 
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub enum AdaptiveAlgorithm {
-    Haversine,
-    Geodesic,
-}
-
 /// Enum for calcuation methods
 /// 
 /// The `Haversine` method is a simple formula that calculates the distance between two points on a sphere.
 /// The `Geodesic` method uses the Vincenty formula to calculate the distance between two points on an ellipsoid.
-/// The `Adaptive` method uses evidence-based apprach to dynamically switch between the two methods.
+/// The `Adaptive` method uses evidence-based approach to dynamically switch between the two methods.
 #[derive(Clone, Debug)]
 pub enum DistanceCalculation {
     ///Always use the Haversine formula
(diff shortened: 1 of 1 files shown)


ianthetechie (commented):
Hey @Kandarp05! Sorry I forgot to respond on your issue comment (please feel free to mention/poke again; sometimes things get busy!).

I think what I'd like to see before a PR with a complex adaptive method for switching is a discussion of the tradeoffs first! The issue is to first evaluate things before implementing. An acceptable outcome is "do nothing!" Valhalla, which calculates trans-continental routes for many industry giants, doesn't ever use anything more than Haversine. I'll carry on the discussion over there on the issue. Then we can decide whether any code changes are needed.

Kandarp05:
Thanks for the feedback! I understand that I should have discussed before making code changes, will make sure this doesn't happen again.

My reasoning behind this adaptive method was:
- Haversine is fast but less accurate over long distances, especially at high altitudes.
-  Geodesics is more accurate but computationally expensive.
-  My adaptive method aims to balance performance & accuracy dynamically.

However, the thing about Valhalla sticking to Haversine despite handling large-scale routing. Do they find accuracy tradeoff negligible, or is it purely for performance reasons?
Also is there some kind of benchmark or something we can use to determine if this method is worth implementing?

ianthetechie:
Yeah, that makes sense.

Basically I just want to make sure we aren't importing a ton of complexity that we have to maintain later. An adaptive model definitely makes sense in some mission critical systems where perf is priority one (see ZFS's Adaptive Replacement Cache for an even more extreme example), but here we can probably live with a much simpler heuristic. For example, we could use Vincenty for calculating things like route segment distances to get more accurate numbers on a transcontinental route. But use Haversine for anything where the error range isn't important.

> However, the thing about Valhalla sticking to Haversine despite handling large-scale routing. Do they find accuracy tradeoff negligible, or is it purely for performance reasons?

I'm _pretty_ sure it's for performance + the fact that _most_ vehicle routes are gonna be between like 5km and 150km. 1000km routes are possible, but a lot more rare.

> Also is there some kind of benchmark or something we can use to determine if this method is worth implementing?

Excellent question! A few ideas:

1. First, let's test what the error range is. Let's get a test route that's HUGE. Like New York to LA. You can generate a route on this page: https://stadiamaps.com/demos/simple-routing. Open the web inspector then pull out the JSON from the response after the route renders 😂 Then calculate the length of the polyline with Haversine and Vincenty and see how far off they are. Might be worth trying a few latitudes.
2. Next, to benchmark the performance difference, I'd just use criterion to run a benchmark of both methods. (Load up the route line into an appropriate geometry, then benchmark calculating the length using both Haversine and Geodesic: https://docs.rs/geo/latest/geo/algorithm/line_measures/trait.Length.html. You can do the benchmarking with Criterion.

PR #487, opened 2025-03-14
Repair compiler warnings

bolsinga:
```
Expression of type '[any StyleLayerDefinition]' is unused
```

The closure is defined to return `Void`, so the existing empty statement returns that value. Remove the statement to remove the warning.

### DIFF
--- a/apple/Sources/FerrostarCarPlayUI/UI/CarPlayNavigationView.swift
+++ b/apple/Sources/FerrostarCarPlayUI/UI/CarPlayNavigationView.swift
@@ -46,7 +46,6 @@ public struct CarPlayNavigationView: View, SpeedLimitViewHost,
                     camera: $camera,
                     navigationState: ferrostarCore.state
                 ) { _ in
-                    userLayers
                 }
                 .navigationMapViewContentInset(.landscape(within: geometry, horizontalPct: 0.65))
             }


ianthetechie (commented):
Thanks for the PR!

@Archdoog and I just looked at it on the call and figured out what's up here. The userLayers is supposed to be here, but we realized that our own API is a bit confusing. There are actually TWO possible closures you can pass to the NavigationMapView, and, well, this is the wrong one ;)

This other closure exists to hack around some camera issues in MapLibre around map loading, which we're discussing over here: https://github.com/maplibre/swiftui-dsl/issues/75.

bolsinga:
Interesting. @Archdoog let me know about this repository; and one of the first things I thought to tackle was to remove the compiler warnings. I certainly don't understand the flow the the library yet, so thanks for the tip. Should I close this PR, since it's wrong? Thanks.

Archdoog:
I think we'll whip up a GitHub issue to remove the `onStyleLoaded` init args from the SwiftUI views in this repo so that they only have a single `@MapViewContentBuilder` closure. Once that's posted one of us can close this PR.

This was a slight slip up toward Android Jetpack Compose convention, where there are no view modifiers and everything goes in the init args. However, Swift correctly compiled to the first onStyleLoaded closure, even though I thought I was using the `@MapViewContentBuilder` closure by virtue of the type I passed it. One of those rare cases where the ability to omit a `return` and the fact it was `Void` allows you to pass the compiler check with something that's wrong 😅.

Here's the issue: (NavigationMapView.swift#L43)[https://github.com/stadiamaps/ferrostar/blob/f6575379cc36c6027be63f84dbf26a7195a4402c/apple/Sources/FerrostarMapLibreUI/Views/NavigationMapView.swift#L43]
  
This PR was super helpful in diagnosing an issue that was not obvious, and might've taken quite a while to debug down the line 👍

bolsinga:
Aha. One thing that could be done is to enable compiler warnings are errors, but only on CI, so a PR can't be landed that way. However if you build it locally (by way of importing the package) it will build fine with the warnings. The logic is to prevent frustrating downstream package client build errors, and keep the package clean for CI.

Archdoog:
I've revised this in: https://github.com/stadiamaps/ferrostar/pull/499/files#diff-09aeef8c81d7b324f4f561bcf71a753310c8a4a984b0aabe19be2af79bf272ceR52.

I think we can close this for now 👍

PR #488, opened 2025-03-17
Vite update to version 6

tedi4t:

PR #489, opened 2025-03-18
Fix Vite 6 build issues

ianthetechie:
Fix breakage from #488 (mistakenly merged because we didn't have enough checks as part of CI).


Archdoog (approved):

ianthetechie:
FYI @tedi4t it looks like the issue blocking the Vite upgrade was that `vite-plugin-top-level-await` needs an upgrade. I've fixed it.

Issue #313, opened 2024-10-27
Log Invalid VoiceInstruction JSON


Archdoog:
When voice isn't working because the JSON model is invalid or missing, we should throw an error. This will help prove and debug issues where no voice instructions are coming through regardless of mute state. It's especially important because we're attempting to support a variety of backend APIs.

ianthetechie:
This whole set of related issues is probably "good first issue" material for anyone looking to get started on the Rust side.

ianthetechie:
@Archdoog I think we just merged a PR that substantially addresses this issue by raising a more informative error. Any chance you'd be able to either confirm that this fixes things or specify a bit more clearly what you're looking for? There isn't really a single accepted logging framework on most platforms, so I think throwing exceptions with useful info is probably good enough.

PR #491, opened 2025-03-21
Improve OSRM API error message handling

### LINKED ISSUE #170: RouteResponseParser Error strategy
When using tools like `createOsrmResponseParser`, your API your API may return a standard data structure (e.g. Valhalla's `{"code": "InvalidOptions", "message": "Options are invalid.", "routes": []}`). We should offer a method to handle error parsing/throwing through the parser.

### LINKED ISSUE #263: Add logging to common API failure points

Kandarp05:
Fixes #170 

Significant step toward logging desired for #263 #313 

When routing APIs return error responses (like "no route found" or "distance exceeded"), the error message was being lost during processing. Instead of showing the actual API error message, users would see a generic JSON parsing error like:
```
Failed to parse route json object: missing field `routes`
```
OR
```
Failed to parse route json object: missing field 'routes' at line 1 column 85
```
### Problem
Previously, when routing APIs returned error responses, JSON parsing would fail before we could extract the actual error message. This happened because error responses often don't include routes and waypoints fields.

### Solution
We're taking advantage of the existing `#[serde(default)]` annotations on the `RouteResponse` struct fields. This makes missing fields use default values (empty vectors) instead of failing to parse.

Now the parsing succeeds even for error responses, allowing our code to check the code field and properly extract the error message.

Now the error looks like this
```
Uncaught (in promise) Routing adapter returned an unexpected status code: DistanceExceeded: Path distance exceeds the max distance limit..
```

### DIFF
--- a/common/ferrostar/src/routing_adapters/osrm/mod.rs
+++ b/common/ferrostar/src/routing_adapters/osrm/mod.rs
@@ -48,12 +48,9 @@ impl RouteResponseParser for OsrmResponseParser {
                 .map(|route| Route::from_osrm(route, &res.waypoints, self.polyline_precision))
                 .collect::<Result<Vec<_>, _>>()
         } else {
-            let error_description = match res.message {
-                Some(message) => format!("{}: {}", res.code, message),
-                None => res.code,
-            };
             Err(ParsingError::InvalidStatusCode {
-                code: error_description,
+                code: res.code,
+                description: res.message,
             })
         }
     }
@@ -467,10 +464,11 @@ mod tests {
         let result = parser.parse_response(error_json.as_bytes().to_vec());
 
         assert!(result.is_err());
-        if let Err(ParsingError::InvalidStatusCode { code }) = result {
+        if let Err(ParsingError::InvalidStatusCode { code, description }) = result {
+            assert_eq!(code, "NoRoute");
             assert_eq!(
-                code,
-                "NoRoute: No route found between the given coordinates"
+                description,
+                Some("No route found between the given coordinates".to_string())
             );
         } else {
             panic!("Expected InvalidStatusCode error with proper message");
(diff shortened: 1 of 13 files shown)


common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -48,7 +48,13 @@ impl RouteResponseParser for OsrmResponseParser {
                 .map(|route| Route::from_osrm(route, &res.waypoints, self.polyline_precision))
                 .collect::<Result<Vec<_>, _>>()
         } else {
-            Err(ParsingError::InvalidStatusCode { code: res.code })
+            let error_description = match res.message {
+                Some(message) => format!("{}: {}", res.code, message),
+                None => res.code,
+            };
+            Err(ParsingError::InvalidStatusCode {
+                code: error_description,
```
ianthetechie, line 56: I'd suggest keeping the code and _adding_ a description field. Also, the struct field doesn't match what it stores now though (it says `code` but it's really a combined description) :)

common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -448,4 +454,47 @@ mod tests {
...
+        assert!(result.is_err());
+        if let Err(ParsingError::InvalidStatusCode { code }) = result {
+            assert_eq!(
+                code,
+                "NoRoute: No route found between the given coordinates"
+            );
+        } else {
+            panic!("Expected InvalidStatusCode error with proper message");
+        }
+    }
```
ianthetechie, line 499: 👍 great tests; thanks!

web/src/ferrostar-map.ts
```diff
@@ -297,7 +297,12 @@ export class FerrostarMap extends LitElement {
     });
 
     const responseData = new Uint8Array(await response.arrayBuffer());
-    return this.routeAdapter.parseResponse(responseData);
+    try {
+      return this.routeAdapter.parseResponse(responseData);
+    } catch (e) {
+      console.error("Error parsing route response:", String(e));
```
ianthetechie, line 303: I *think* it is more idiomatic JavaScript to write it like this... any thoughts @haysmike?

```suggestion
      console.error(`Error parsing route response: ${e}`);
```

haysmike: Thanks for the ping! Here's what I'd do:
```suggestion
      console.error("Error parsing route response:", e);
```

This should print out both the string and the error message nicely in the console and allow you to expand/collapse the stack.

ianthetechie (approved):
Looks good assuming CI passes. Thanks!

Kandarp05:
Thanks for merging! Appreciate all the feedback—excited to keep contributing!

PR #492, opened 2025-03-24
show only navigation

tedi4t:


web/src/ferrostar-map.ts
```diff
@@ -25,9 +25,6 @@ export class FerrostarMap extends LitElement {
   @property()
   valhallaEndpointUrl: string = "";
 
-  @property()
-  styleUrl: string = "";
-
```
ianthetechie, line 30: I believe this property is still necessary for users that are not bringing their own custom MapLibre map component.

web/src/ferrostar-map.ts
```diff
@@ -221,54 +215,6 @@ export class FerrostarMap extends LitElement {
...
-      if (this.geolocateOnLoad) {
-        this.geolocateControl?.trigger();
-      }
-
-      if (this.configureMap !== undefined) {
-        this.configureMap(e.target);
-      }
-    });
-  }
-
```
ianthetechie, line 271: I think this also assumes that the user must supply a component. I guess I don't necessarily have an issue with that, but as things stand today, there is no de facto standard npm package you can install that gives you a web component for MapLibre; you need to write your own (e.g. with Vue).

web/src/ferrostar-map.ts
```diff
@@ -463,7 +409,6 @@ export class FerrostarMap extends LitElement {
       <div id="container">
         <div id="map">
           <!-- Fix names/ids; currently this is a breaking change -->
-          <slot id="map-content"></slot>
```
ianthetechie, line 466: If this slot goes away, how does the custom map get added to the DOM?

ianthetechie (commented):
I'm a bit confused by this PR overall. It doesn't work with the demo app in the repo at all anymore.

web/index.html
```diff
@@ -5,6 +5,31 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Ferrostar Web Demo</title>
     <link rel="stylesheet" href="./src/index.css" />
+    <link
+      href="https://unpkg.com/maplibre-gl@latest/dist/maplibre-gl.css"
+      rel="stylesheet"
+    />
+    <link
+      href="https://unpkg.com/@stadiamaps/maplibre-search-box/dist/style.css?inline"
```
ianthetechie, line 13:
```suggestion
      href="https://unpkg.com/@stadiamaps/maplibre-search-box/dist/maplibre-search-box.css"
```

web/index.html
```diff
@@ -97,6 +132,7 @@
             ferrostar.startNavigation(route, config);
           },
         });
+        mapInstance.addControl(searchBox, "top-left");
 
         ferrostar.center = { lng: -122.42, lat: 37.81 };
         ferrostar.zoom = 18;
```
ianthetechie, line 138: GitHub is stupid and doesn't let me comment on lines below this, but the following line should be removed now:

```javascript
        ferrostar.customStyles = searchBoxStyle;
```

Pending the result of the other large discussion on `ferrostar-map.ts`, we should also delete the `configureMap` assignment, since this is currently broken and should probably be deleted.

web/src/ferrostar-map.ts
ianthetechie: I'm OK with this, generally, but there are a few things about the current implementation that are half-baked:

* The geolocateControl no longer exists. I think this is fine if we make it the user's responsibility to handle this, but we should clean up the code in this file. Similarly, `geolocateOnLoad` is currently broken and should be either fixed or removed.
* The `configureMap` property is also unused; either hook it up or delete it.
* Is the `maplibreglStyles` necessary here? I don't think so, since we don't actually load the map within the component anymore.

web/index.html
```diff
@@ -61,8 +90,14 @@
       };
 
       async function onload() {
+        const mapInstance = new mapLibreGL.Map({
+          container: "mapElement",
+          style: "https://tiles.stadiamaps.com/styles/outdoors.json",
+        });
```
ianthetechie, line 96: Pending the outcome of the other discussion, it would be nice to configure the map similarly, with a geolocate control, auto-geolocation on load, center, and pitch.

ianthetechie (commented):
Thanks for your your work on this! I finally got around to properly evaluating it, and I think you have the right approach. Just a bunch of small clean-up things.

PS: Normally I'd just push a bunch of edits myself, but your PR is coming from an org-owned fork. There's a ~10 year old design issue with GitHub that means even though you allow maintainer edits, we can't make them 😅 If you open PRs from a personal fork, we can make changes and get stuff merged a bit faster. Thanks!

ianthetechie:
Superseded by #517

PR #493, opened 2025-03-26
Fix UniFFI binding regen for linux

Kandarp05:
Skip XCFramework generation and only generate FFI when `--ffi-only` is specified, enabling users to regenerate FFI on Linux


ianthetechie (approved):
Wow, thanks! Just tried it out and seems to work. This will really improve things for contributors working on Linux!

PR #499, opened 2025-04-01
feat: improved navigation view customization

Archdoog:
- Reduces size of default CurrentRoadNameView to better accommodate landscape.
- Adds view modifiers for customizing InstructionsView, ProgressView, and CurrentRoadNameView using consistent `func navigationView` convention.
- Enables sending a UserLocation into the startNavigation function on iOS. This more realistically handles typical usage, where the location provider's location is less likely to match the start of the provided route than one that's manually introduced.
- Adds testing target for MapLibreUI
- Moves some testing tooling to the `TestSupport` shared target.


apple/Sources/FerrostarCore/FerrostarCore.swift
```diff
@@ -252,12 +252,18 @@ public protocol FerrostarCoreDelegate: AnyObject {
     ///
     /// - Parameters:
     ///   - route: The route to navigate.
+    ///   - userLocation: The user's location. This should be as close to the users location and the start of the route
+    /// as possible. Giving fetch timing, this may not be possible, but the closer the better.
```
ianthetechie, line 256: Things to document:

* userLocation is an override/default, but it will try falling back to locationProvider.lastLocation
* Explain why stale locations may cause unexpected behavior.

ianthetechie (commented):
Looks good overall.

We'll pass on the snapshot testing for now since it's complicated; will have Ian and Jacob manually review.

apple/Sources/FerrostarCore/FerrostarCore.swift
ianthetechie:
```suggestion
    /// as possible. If the location is too stale, the user may be almost immediately flagged as off the route,
    /// triggering a recalculation.
    /// If this parameter is `nil`, the last location will be obtained from the configured location provider automatically.
    /// If no location is available, this method will throw an exception.
```

PR #514, opened 2025-04-05
handle Graphhopper requests

karussell:
Please review carefully as I'm not a native speaker of Rust or Kotlin :)

My initial approach was to convert the "native" GraphHopper response inside ferrostar, but due to some problems and because it makes other situations also simpler I've created a server-side solution, where the endpoint accepts a "native" GraphHopper (POST) request and returns the OSRM-alike response, which is already supported from ferrostar and e.g. maxspeed annotations worked out of the box. (I kept the early commits mainly for myself and I think this is fine as you are squashing commits anyway?)

Currently there are no tests as I wasn't sure what to test and also there is this notice "move out of Android Tests ASAP". Is this still valid? I could test e.g. that the request is properly formed - just let me know if I should e.g. copy ValhallaCoreTest.kt or if I should do something else.

And should we create a more generic route adapter creation like this?
```kotlin
RouteAdapter.newHttp(routingEngine, routingEndpointURL.toString(), profile, jsonAdapter.toJson(options))
```

Otherwise we have to duplicate the cases of the different routing engines at a higher level in all different environment. One tricky part for other engines could be the authentication as it is often **not** done via a simple API key in the URL (like used from the Stadia Maps and the Graphhopper API).

Let me know which one you prefer and I could try to make the web/wasm environment working.

Currently there is also no support for iOS and react-native, but it should be only a few lines as you can see in the Kotlin code. Unfortunately, I probably won't be of much help here.

### DIFF
--- a/guide/src/android-getting-started.md
+++ b/guide/src/android-getting-started.md
@@ -223,9 +223,19 @@ The `FerrostarCore` instance should live for at least the duration of a navigati
 Bringing it all together, a typical init looks something like this:
 
 ```kotlin
+var options =
+    mapOf(
+        "costingOptions" to
+            mapOf(
+                "bicycle" to
+                    mapOf(
+                        "use_roads" to 0.2
+                        )),
+        "units" to "miles")
 private val core =
       FerrostarCore(
-          routingEngine = RoutingEngine.Valhalla("https://api.stadiamaps.com/route/v1?api_key=YOUR-API-KEY", "auto"),
+          wellKnownRouteProvider = WellKnownRouteProvider.Valhalla("https://api.stadiamaps.com/route/v1?api_key=YOUR-API-KEY", "bicycle")
+		    .withJsonOptions(options),
           httpClient = httpClient,
           locationProvider = locationProvider,
           foregroundServiceManager = foregroundServiceManager,
@@ -241,13 +251,12 @@ private val core =
       )
 ```
 
-`FerrostarCore` exposes a number of convenience constructors for common cases,
-such as using a Valhalla [Route Provider](./route-providers.md#bundled-support),
-and automatically subscribes to location updates from the `LocationProvider`.
+### Route Providers
 
-There are a LOT of options, but don’t worry; everything is documented.
-Check out the [Navigation Behavior](configuring-the-navigation-controller.md)
-customization chapter for details.
+You’ll need a route provider when you set up your `FerrostarCore` instance.
+Several are available via the `WellKnownRouteProvider` enum,
+or you can [configure your own from scratch](./route-providers.md).
+Refer to the [commercial vendors](./vendors.md) page for some known working integrations.
 
 ## Set up voice guidance
 
--- a/guide/src/ios-getting-started.md
+++ b/guide/src/ios-getting-started.md
@@ -93,15 +93,36 @@ and use it in SwiftUI using a `private @StateObject` or `@ObservedObject`
 in most situations.
 
 `FerrostarCore` provides several initializers,
-including convenient initializers for Valhalla routing backends
+including convenience helpers for well-known route providers
 and the ability to provide your own custom routing (ex: for local/offline use).
+
+Here's a full example configuration:
+
+```swift
+let core = try FerrostarCore(
+    wellKnownRouteProvider: .valhalla(
+        endpointUrl: "https://api.stadiamaps.com/route/v1?api_key=\(sharedAPIKeys.stadiaMapsAPIKey)",
+        profile: "bicycle"
+    )
+    .withJsonOptions(options: ["costing_options": ["bicycle": ["use_roads": 0.2]]]),
+    locationProvider: locationProvider,
+    navigationControllerConfig: config,
+    // This is how you can set up annotation publishing;
+    // We provide "extended OSRM" support out of the box,
+    // but this is fully extendable!
+    annotation: AnnotationPublisher<ValhallaExtendedOSRMAnnotation>.valhallaExtendedOSRM(),
+    widgetProvider: FerrostarWidgetProvider()
+)
+```
+
+Let's dive into the configuration options a bit further.
 
 ### Route Providers
 
-You’ll need to decide on a route provider when you set up your `FerrostarCore` instance.
-For limited testing, FOSSGIS maintains a public server with the URL `https://valhalla1.openstreetmap.de/route`.
-For production use, you’ll need another solution like a [commercial vendor](./vendors.md)
-or self-hosting.
+You’ll need a route provider when you set up your `FerrostarCore` instance.
+Several are available via the `WellKnownRouteProvider` enum,
+or you can [configure your own from scratch](./route-providers.md).
+Refer to the [commercial vendors](./vendors.md) page for some known working integrations.
 
 ### Set up Voice Guidance
 
--- a/guide/src/vendors.md
+++ b/guide/src/vendors.md
@@ -7,10 +7,10 @@ PRs welcome to expand the list or to add support!
 
 * [Stadia Maps](https://stadiamaps.com/)
 	* Global hosted Valhalla API (free keys for evaluation available)
-	* Custom deployments with proprietary data integrations possible
-* [GIS • OPS](https://gis-ops.com/)
-	* Routing specialists, with an emphasis on Valhalla
-	* Specializes in custom deployments and integration of proprietary data
+	* Traffic and other proprietary data integrations available
+* [GraphHopper](https://www.graphhopper.com/)
+    * Global hosted GraphHopper API (free keys for evaluation available)
+	* Traffic available via add-on
 
 ## Basemaps
 


android/core/src/androidTest/java/com/stadiamaps/ferrostar/core/ValhallaCoreTest.kt
```diff
@@ -248,7 +248,8 @@ class ValhallaCoreTest {
         }
     val core =
         FerrostarCore(
-            valhallaEndpointURL = URL(valhallaEndpointUrl),
+            routingEndpointURL = URL(valhallaEndpointUrl),
+            routingEngine = "valhalla",
```
Archdoog, line 252: We should probably consider a strategy where this is a sealed class with options. E.g.

```kt
sealed class RoutingEngine {
    data class Valhalla(val endpoint: String) : RoutingEngine
    data class GraphHopper(val endpoint: String): RoutingEngine
    // TODO: we might be able to move the CustomRouteProvider in here somehow as well.
}
```

For swift this would look something like:

```swift
enum RoutingEngine {
    case valhalla(String)
    case graphHopper(String)
}
```

android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
```diff
@@ -157,8 +158,12 @@ class FerrostarCore(
       options: Map<String, Any> = emptyMap(),
   ) : this(
       RouteProvider.RouteAdapter(
-          RouteAdapter.newValhallaHttp(
-              valhallaEndpointURL.toString(), profile, jsonAdapter.toJson(options))),
+          if (routingEngine == "graphhopper")
```
Archdoog, line 161: Accepting the sealed class above, this would become. `when` statement.

Archdoog (commented):
Some initial comments. Overall looking simple enough. Will let @ianthetechie critique the rust stuff.

common/ferrostar/src/routing_adapters/snapshots/ferrostar__routing_adapters__graphhopper__tests__request_body_with_custom_profile.snap
```diff
@@ -0,0 +1,34 @@
+---
+source: ferrostar/src/routing_adapters/graphhopper.rs
+expression: "generate_body(USER_LOCATION, WAYPOINTS.to_vec(),\nSome(r#\"{\n            \"ch.disable\": true,\n            \"custom_model\": {\n                \"distance_influence\": 15,\n                \"speed\": [\n                    {\n                        \"if\": \"road_class == MOTORWAY\",\n                        \"limit_to\": \"100\"\n                    }\n                ]\n            }\n        }\"#))"
+---
```
ianthetechie, line 4: @karussell can you give these snapshots a quick check? These are generated in the routing adapter tests I just added. Below is JSON that would be generated and sent to the server. I couldn't find any obvious docs on the navigate API but looking through the code, I think this is OK...

common/ferrostar/src/routing_adapters/graphhopper.rs
```diff
@@ -0,0 +1,322 @@
...
+    ///     "en",
+    ///     VoiceUnits::Metric,
+    ///     options
+    /// );
+    /// ```
+    pub fn new<U: Into<String>, P: Into<String>, L: Into<String>>(
+        endpoint_url: U,
+        profile: P,
+        locale: L,
+        voice_units: VoiceUnits,
```
ianthetechie, line 89: With Valhalla, we leave locale and voice units unspecified, but you had them hard-coded, which makes me wonder if these are required and/or strongly encouraged for every request. I've made some changes to the interface assuming the answer is "yes" but please let me know your thoughts. For Valhalla, we currently just handle this with extra JSON options, but I'd like to move toward making the options easier to use with strong types, so here's a first pass at this.

cc @Archdoog in case you have any comments.

ianthetechie (commented):
Sorry this took a bit for me to get around to! This is really great! I've just done some cleanup + added docs. If you could review the questions (esp verifying the JSON output :D), we can get this merged pretty quickly + I'll handle the iOS side.

android/core/src/androidTest/java/com/stadiamaps/ferrostar/core/ValhallaCoreTest.kt
ianthetechie: Yeah, agreed. I can take another pass at that. This is one of those weird Kotlin things ;)

common/ferrostar/src/routing_adapters/snapshots/ferrostar__routing_adapters__graphhopper__tests__request_body_with_custom_profile.snap
karussell: Yes, this looks good.

What is the purpose of these snapshots?

> I couldn't find any obvious docs on the navigate API

This is on my list :) ... basically it is a mix of existing things. The request JSON format is identical to the POST request for `/route` and the response is compatible (tries to be identical) to mapbox/osrm.

android/demo-app/src/main/java/com/stadiamaps/ferrostar/AppModule.kt
```diff
@@ -68,6 +83,11 @@ object AppModule {
   val locationProvider: LocationProvider by lazy {
     // TODO: Make this configurable?
     FusedLocationProvider(appContext)
+    // SimulatedLocationProvider().apply {
+    //  warpFactor = 2u
+    //  lastLocation =
+    //      UserLocation(GeographicCoordinate(51.049315, 13.73552), 1.0, null, Instant.now(), null)
+    // }
```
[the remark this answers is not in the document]
karussell: Please have a look. I made it configurable with a boolean now.

android/core/src/androidTest/java/com/stadiamaps/ferrostar/core/ValhallaCoreTest.kt
karussell: Not sure if I fully understand this but I have added this approach in the last commit. Please see the change and let me know if this is what you meant. (I also included the routing profile as it is dependent on the routing engine)

common/ferrostar/src/routing_adapters/snapshots/ferrostar__routing_adapters__graphhopper__tests__request_body_with_custom_profile.snap
ianthetechie: > What is the purpose of these snapshots?

Basically a sanity check that will check that we're generating the API request correctly. If we change anything about how it's generated by accident, the snapshot test will fail until it's updated.

Sort of like a unit test that writes itself ;)

ianthetechie:
Apologies for the delay on this. I'll handle the merge conflicts on this and see what's needed to get this over the line.

android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
```diff
@@ -176,16 +174,27 @@ class FerrostarCore(
...
   ) : this(
       RouteProvider.RouteAdapter(
-          RouteAdapter.newValhallaHttp(valhallaEndpointURL.toString(), profile, options.toJson())),
+          when (routingEngine) {
+            is RoutingEngine.GraphHopper ->
+                RouteAdapter.newGraphhopperHttp(
+                    routingEngine.endpoint,
+                    routingEngine.profile,
+                    "en",
+                    VoiceUnits.METRIC,
```
ianthetechie: These props should probably not be hard-coded. I guess they maybe belong in the RoutingEngine.GraphHopper data class?

common/ferrostar/src/routing_adapters/graphhopper.rs
ianthetechie: I couldn't find this anywhere in the GraphHopper docs, but I did find it in the source code: https://github.com/graphhopper/graphhopper/blob/be761ae311580b64ff258319d2b30437f66982ca/navigation/src/main/java/com/graphhopper/navigation/NavigateResource.java#L93.

It looks like this defaults to imperial, which to be fair, is going to displease at least half of the users and the majority of potential users in the world ;) So yeah, this makes sense to require as is here. Maybe we should do the same for Valhalla 🤔

common/ferrostar/src/routing_adapters/graphhopper.rs
```diff
@@ -0,0 +1,323 @@
...
+
+use serde::Serialize;
+use serde_json::{json, Map, Value as JsonValue};
+#[cfg(feature = "std")]
+use std::collections::HashMap;
+
+#[derive(Serialize, Debug)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+#[serde(rename_all = "lowercase")]
+pub enum VoiceUnits {
```
ianthetechie: I realized after an unrelated PR that the UniFFI namespace is flat, so we should probably add `GraphHopper` in front of this to clarify the specific purpose 😅

common/ferrostar/src/routing_adapters/graphhopper.rs
karussell: > It looks like this defaults to imperial

The default should be `metric` but yes, either way, it needs to be configurable somehow. The voice_units and locale parameters are not required, but I thought I highlight them, as they are often what users want to change and pass e.g. device information.

common/ferrostar/src/routing_adapters/graphhopper.rs
karussell: Can you explain what you mean :) ? Is this something I have to change or you need information from me? (I've no knowledge with rust and UniFFI :) )

ianthetechie: I've already made the change, and am working on the rest now :)

common/ferrostar/src/routing_adapters/graphhopper.rs
ianthetechie: Oh oops sorry I conflated this with line 145:

```java
            DistanceUtils.Unit unit = voiceUnits.equals("metric") ? DistanceUtils.Unit.METRIC : DistanceUtils.Unit.IMPERIAL;
```

Upon some further reflection, I think I actually like your approach of surfacing the super common parameters like this which almost everyone has an opinion on, and will probably adapt Valhalla to do the same.

karussell: > Oh oops sorry I conflated this with line 145

No worries :) ... the default is defined via the annotation `@DefaultValue("metric")` in the line you linked. (We use dropwizard, but I think most Java web servers do it via annotations.)

> I think I actually like your approach of surfacing the super common parameters like this which almost everyone has an opinion on, and will probably adapt Valhalla to do the same.

:+1:

karussell:
Thanks a lot for your final work here!

PR #517, opened 2025-04-16
Show only navigation (personal repo)

tedi4t:


ianthetechie (approved):
Thanks, for your patience while I worked out what the best approach was for some of the last bits here!

I made a few more changes and it's in a place where I'm happy to merge it!

One point of possible discussion: after some thought, I believe the `configureMap` property and all of the other map viewport properties (`center`, `pitch`, `zoom`, which were only one-way bindings anyways!!) make even less sense now. So I propose that we drop these completely and keep all of the map stuff external.

Let me know if you can think of a reason why you'd prefer to keep them instead. Otherwise, I'll merge and release as soon as you confirm.

PR #521, opened 2025-04-28
Trip state media queries

tedi4t:


ianthetechie (approved):
Excellent; this makes things look much better on narrow screens!

PR #541, opened 2025-05-06
FerrorstarCore always creates and owns the SpokenInstructionObserver

bolsinga:
The goal is to have less parts and an unified interface to the state. If this is non-optional it will just be a part of FerrostarCore. FerrostarCore is Observable, so its spokenInstructionObserver is Observable too.


ianthetechie (commented):
Yeah, good call. I think we (namely me :P) probably overthought this the first time around.

Code looks good; can you also update the guide real quick? I think just this file needs some light edits to reflect that this is now optionally settable at constructor time, and that you no longer need to explicitly opt in: https://github.com/stadiamaps/ferrostar/blob/main/guide/src/ios-getting-started.md

ianthetechie (approved):

PR #542, opened 2025-05-06
DemoApp: Clean up and simplification

bolsinga:
- Also Initialize FerrostarCore via an extension to encapsulate details
    - Use `try!` at the last possible moment and let the intervening items `throw`.
    - use an extension to create the `SimulatedLocationProvider`
    - Depends upon #541


bolsinga:
going to see if this can be only for one commit.

PR #546, opened 2025-05-08
AppEnvironment is a global variable.

### LINKED ISSUE #544: iOS DemoApp: Application Delegate is instantiated twice
I logged the initializer for `DemoAppDelegate`. It is created when the iOS application launches. Once the system determines the DemoApp's `CarPlaySceneDelegate` should be instantiated, that class uses:

```
    // Get the AppDelegate associated with the SwiftUI App/@main as the type you defined it as.
    @UIApplicationDelegateAdaptor(DemoAppDelegate.self) var appDelegate

```

This will actually create a new `DemoAppDelegate`, and not get one if there is already one. Then there are two. This may be a source of oddities when first launching navigation in CarPlay without the CarPlay window already shown at launch of the iOS app.

Reading the docs for SwiftUI, if the application delegate conforms to `ObservableObject`, SwiftUI will put the application delegate into the `Environment` for you. The problem is that we need to share navigation state between SwiftUI in the application and with CarPlay. If the `Environment` is read in the DemoApp's `CarPlaySceneDelegate`, it will not be populated since this is not in the SwiftUI realm. We'd have to read it in the SwiftUI View. But down there we do not have access to the DemoApp's specific `UIApplicationDelegate`, and thus lose the ability to access the navigation state without a cast.

Out loud thought: If the DemoApp's `CarPlayDelegate` had a small SwiftUI view that was located the app, that app view could read the app delegate, and grab the navigation state since it knows the type. Then it could hand that navigation state to the library's CP SwiftUI views.

While this is an issue with the DemoApp, it also seems to affect the design of the CarPlay navigation SDK in Ferrostar.

bolsinga:
- The old code would instantiate the `DemoAppDelegate` twice via `UIApplicationDelegateAdaptor`. This would make starting up the app when the CarPlay window was open behave differently based upon the order of launching.
- The documentation for `UIApplicationDelegateAdaptor` indicates that the application delegate will be in the `Environment` if it conforms to `ObservableObject`. This attempt showed it was true in the application, but not in CarPlay.
- Then an attempt was made to put the `AppEnvironment` into SwiftUI's `Environment` to be read on the CarPlay side of things. The App's `CarPlaySceneDelegate` could not read the `Environment`, since it is not SwiftUI. An attempt was made to have a SwiftUI view in the application that hosted a Ferrostar `CarPlayNavigationView`, so that the app's `View` could access the `Environment` and pass it to the library `View`. However this `Environment` did not have the `AppEnvironment` either.
- The [most promising idea](https://stackoverflow.com/a/57089965), which a few attempts were close, but not close enough was to have the `application(_:configurationForConnecting:options:)` pass the `AppEnvironment` into the `UISceneSession.userInfo`. This would work but unfortunately, `CarPlaySceneDelegate` was still instantiated twice. Basically the code would configure the `UIScene` much like the `UISceneManifest` in the Info.plist would. If I removed data from the `UISceneManifest` in the Info.plist, hoping it would not instantiate twice, the runtime would crash because the information wasn't there. I'm sure this is "the way" to pass data from `UIApplication` to `UIScene`, but it isn't working out yet. Just a matter of getting that configuration right. [Interesting information about SwiftUI lifecycle and CarPlay](https://github.com/vanities/carplay-swiftui).
- So this is a quick fix so that launch of CarPlay works better and it is connected to CarPlay from the get go.

Fixes #544 and a better fix should be addressed in the future.

### DIFF
--- a/apple/DemoApp/Demo/CarPlaySceneDelegate.swift
+++ b/apple/DemoApp/Demo/CarPlaySceneDelegate.swift
@@ -11,9 +11,6 @@ private extension Logger {
 }
 
 class CarPlaySceneDelegate: UIResponder, UIWindowSceneDelegate, CPTemplateApplicationSceneDelegate {
-    // Get the AppDelegate associated with the SwiftUI App/@main as the type you defined it as.
-    @UIApplicationDelegateAdaptor(DemoAppDelegate.self) var appDelegate
-
     private weak var ferrostarCore: FerrostarCore?
     private var carPlayViewController: UIViewController?
 
@@ -55,14 +52,14 @@ class CarPlaySceneDelegate: UIResponder, UIWindowSceneDelegate, CPTemplateApplic
         guard carPlayManager == nil else { return }
 
         // IMPORTANT: This is your app's shared FerrostarCore
-        ferrostarCore = appDelegate.appEnvironment.ferrostarCore
+        ferrostarCore = appEnvironment.ferrostarCore
 
         let view = CarPlayNavigationView(
             ferrostarCore: ferrostarCore!,
             styleURL: AppDefaults.mapStyleURL,
             camera: Binding(
-                get: { self.appDelegate.appEnvironment.camera.camera },
-                set: { self.appDelegate.appEnvironment.camera.camera = $0 }
+                get: { appEnvironment.camera.camera },
+                set: { appEnvironment.camera.camera = $0 }
             )
         )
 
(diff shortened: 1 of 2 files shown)


bolsinga:
I'm going to try another thing

bolsinga:
I tried to do more with `UISceneSession.userData` but I still can't seem to get the cycle of going through the various `UISceneDelegate` methods to work. I'll keep that branch on the side and work on it later. It's definitely better than this global variable version.

However this global variable version works for now, and allows me to work more on other parts of CarPlay.

ianthetechie (approved):

PR #547, opened 2025-05-10
feat: preferred user location

Archdoog:
- Dynamically toggle snapped location versus raw location for route deviation off route.
- Enables consistent user location pass through for non-navigating and off route situations. 
- Removes custom android code that achieved this only one platform.


ianthetechie (approved):
This is a slight regression in flexibility introduced by #250 for Android, but I think that it's such an improvement for the majority use case that it's worth it. I've updated the original issue #209 with the current state of things and we can revisit adding a "never snapped" config when someone needs it.

Other than that, just a few things I'll clean up.

PR #548, opened 2025-05-11
Optional geolocate control

tedi4t:


ianthetechie (approved):
Just a minor change to the property name for clarity. Looks good otherwise. Thanks!

PR #549, opened 2025-05-12
Adds missing init for Speech on convenience inits

Archdoog:
Pretty self explanatory 👍


bolsinga:
Arg. Should they all funnel through one initializer that has the default? Sorry about #541

ianthetechie (approved):

PR #551, opened 2025-05-13
CarPlayNavigationView.ferrostarCore is just a reference

bolsinga:
The state is held elsewhere (a reference in a global in the DemoApp).


bolsinga:
@Archdoog Would you try this out with your application, which has a different way of referencing `FerrostarCore`? Thanks!

PR #556, opened 2025-05-13
reset uiState when trip ends

### LINKED ISSUE #552: iOS: FerrostarCarPlayAdapter.uiState does not change from .navigating
If a user start/stop/start navigation from the app, the CP UI will not show maneuvers for the 2nd start. This is because `uiState == .navigating`, so the `CPMapTemplate` is never told to start again.

Need to figure out the proper flow for this so the state is maintained properly.

bolsinga:
- Add a helper method where the adapater calls the templatehost to terminate a trip.
- Reset `uiState` within this helper function.
- Fixes #552


ianthetechie (approved):
LGTM so I'll approve. Will wait for @Archdoog to test this an #551 though since he's the main production user.

Archdoog (approved):
Looks good to me as well.

PR #576, opened 2025-05-30
Improve error handling in iOS Demo App.

bolsinga:
- Show UI when routes fail to load and when navigation fails to start.
- The point of this change is to get the application ready for a state that is "There is not a FerrostarCore instance, since the route to navigate has not been picked yet". This will allow an "idle state" to exist in this demo.


ianthetechie (commented):
I like the error handling improvements, but I think we should probably not nudge demo app users toward assuming only one route is useful. While we don't really showcase the fact that routing APIs can (and do!) produce multiple routes now, I think I'd prefer to leave that as-is.

bolsinga:
> I like the error handling improvements, but I think we should probably not nudge demo app users toward assuming only one route is useful. While we don't really showcase the fact that routing APIs can (and do!) produce multiple routes now, I think I'd prefer to leave that as-is.

I almost made these 2 PRs! 😀

bolsinga:
> I like the error handling improvements, but I think we should probably not nudge demo app users toward assuming only one route is useful. While we don't really showcase the fact that routing APIs can (and do!) produce multiple routes now, I think I'd prefer to leave that as-is.

OK. I remember why I did this now. The UI is only good for 1 route. The App and the UI requires 1 route, since there isn't a route picking UI in the Demo App.

AppEnvironment (the model in this case) already throws if there is not a route.

If I check for 1 route in the UI, then the UI gets into logic, which I'd prefer to avoid. What do you think?

ianthetechie (approved):
Just added a comment to the code. Will merge after CI passes.

PR #587, opened 2025-06-03
Android Http Client Abstraction

ben-burwood:
Following from RFD here https://github.com/stadiamaps/ferrostar/discussions/586

This PR aims to abstract some of the HTTP Code in Android from being tied to OkHTTP, allowing for Ktor to be used instead.


ianthetechie (commented):
Ooh this is great! Thanks for the PR! Definitely support the ability to choose a different HTTP client.

I'm OK merging this as-is if you're planning on following up with some additional changes soon-ish. Otherwise, we should probably add some docs about how it works. Probably a new chapter under Customization... which we should probably organize by platform soon, but we can handle that.

ben-burwood:
> Should probably be renamed `HttpProvider` or `HttpClientProvider` thinking about it...

Complete

ben-burwood:
> Ooh this is great! Thanks for the PR! Definitely support the ability to choose a different HTTP client.
> 
> I'm OK merging this as-is if you're planning on following up with some additional changes soon-ish. Otherwise, we should probably add some docs about how it works. Probably a new chapter under Customization... which we should probably organize by platform soon, but we can handle that.

Thanks for the review, have made a couple of small tweaks with renaming to Provider (feels more consistent with other Interfaces).

I have also briefly updated the docs so that the getting started guide is now correct, haven't added anything else (e.g. an example of how to use Ktor?).
Don't currently plan any further changes to this, switching to Ktor as the default isn't really necessary anymore - happy to help implement if this is desired though?

ianthetechie (approved):
Yeah, this looks good to me; thanks for those last edits + polish on things!

On Ktor, if you feel like it, great, but I don't think anyone's holding their breath waiting for it.

PR #595, opened 2025-06-11
Update comments for RouteStep fields

SiarheiFedartsou:
Actually these fields are in "official" spec, but in newer version :) 

https://project-osrm.org/docs/v5.24.0/api/#routestep-object


ianthetechie (approved):
Thanks! Will merge once CI checks pass.

Issue #591, opened 2025-06-11
Define Rust Data Models for Recording


Kandarp05:
Create Rust structs/enums to represent recorded session data based on the decided format.

Kandarp05:
All data-models have been defined as per the decided format.

PR #596, opened 2025-06-11
Define data models for recording session

Kandarp05:
#591
 
Defining data models required for the session recording system, based on the design discussion so far.

Current implementation is an extension to `navigation_controller` module, would it be cleaner to move these models to a separate file?

### DIFF
--- a/common/ferrostar/src/navigation_controller/models.rs
+++ b/common/ferrostar/src/navigation_controller/models.rs
@@ -265,10 +265,6 @@ pub enum NavigationRecordingEventData {
         /// Updated trip state.
         trip_state: TripState,
     },
-    UserInteraction {
-        /// User interaction data.
-        interaction_data: UserInteractionData,
-    },
     RouteUpdate {
         /// Updated route steps.
         route_steps: Vec<RouteStep>,
@@ -277,13 +273,4 @@ pub enum NavigationRecordingEventData {
         /// Error message.
         error_message: String,
     },
-}
-
-pub enum UserInteractionData {
-    /// User has requested to advance to the next step.
-    AdvanceToNextStep,
-    /// User has requested to stop the navigation session.
-    StopNavigation,
-    /// User has requested to toggle the mute state of the navigation instructions.
-    ToggleMute,
 }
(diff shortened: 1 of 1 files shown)


Kandarp05:
This looks good enough for now, let me know if any changes are required

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -226,3 +226,64 @@ pub struct NavigationControllerConfig {
     /// Configures how the heading component of the snapped location is reported in [`TripState`].
     pub snapped_location_course_filtering: CourseFiltering,
 }
+
+pub struct NavigationRecording {
+    /// Version of ferrostar that created this recording.
```
ianthetechie:
```suggestion
    /// Version of Ferrostar that created this recording.
```

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -226,3 +226,64 @@ pub struct NavigationControllerConfig {
...
+}
+
+pub enum UserInteractionData {
+    /// User has requested to advance to the next step.
+    AdvanceToNextStep,
+    /// User has requested to stop the navigation session.
+    StopNavigation,
+    /// User has requested to toggle the mute state of the navigation instructions.
+    ToggleMute,
+}
```
ianthetechie: Ooh, this is interesting... I had NOT actually anticipated us doing any recording of this type of action, since every UI is potentially different. But now that I give it a closer thought, I guess this isn't necessarily a hard disqualifier!

This _does_ necessarily complicate things quite a bit though, since it's not as cleanly self-contained. You actually need to have the frontends forward every such event, define what events are possible, and decide how this affects replay, if at all.

My initial opinion is that this is extra complexity that we don't actually need for GSoC. Anything I'm missing / differing opinions @Archdoog?

ianthetechie (commented):
One discussion point before we merge this. I think we can leave the data in the current file for now and refactor easily later.

common/ferrostar/src/navigation_controller/models.rs
Kandarp05: Yeah, thinking about it again, this might add complexity without much real benefit. Even if the user skips a waypoint, the trip state will still update which we do record, the final recording would essentially remain the same.

ianthetechie: Bingo ;)

ianthetechie (approved):

PR #597, opened 2025-06-11
SpokenInstructionObserver does not need to be ObservableObject

bolsinga:
- This is because `DynamicallyOrientingNavigationView` just takes a `Bool` to indicate muted state, and a closure to toggle the value. The value lives within `FerrorstarCore`'s reference to `SpokenInstructionObserver`.


apple/Sources/FerrostarCore/Speech/SpokenInstructionObserver.swift
```diff
@@ -1,11 +1,10 @@
 import AVFoundation
-import Combine
 import FerrostarCoreFFI
 import Foundation
 
 /// An Spoken instruction provider that takes a speech synthesizer.
-public class SpokenInstructionObserver: ObservableObject {
-    @Published public private(set) var isMuted: Bool
+public class SpokenInstructionObserver {
+    public private(set) var isMuted: Bool
```
Archdoog: I would still expect this to be `@Published` unless we upgrade it to a stream so that upstream (e.g. ViewModel or SwiftUI) can listen to changes of isMuted. 

### How

There's a bit of nuance here between `ObservableObject` and `@Published`. In SwiftUI, both are used together frequently. However, `@Published` is simply a property wrapper for combine that lets downstream observe changes to the value. Check out this doc that helps understand the difference: https://developer.apple.com/documentation/combine/observableobject. 

And note here: https://developer.apple.com/documentation/combine/published that the class containing this `@Published` propertyWrapper doesn't use `ObservableObject`.

Lastly, this is how it can be used

```swift
cancellable = spokenInstructionObserver.$isMuted
    .sink {
        print("Spoken instruction observer isMuted has change to: \($0)")
    }
```

### Example

In my app, I more or less do this:

```swift
@Observable
class MyViewModel: ViewModel {
    var isMuted: Bool

    init() {
        spokenInstructionObserver.$isMuted
             .receive(on: DispatchQueue.main)
            .assign(to: \.isMuted, on: self)
            .store(in: &cancellables)
    }
}
```

Which then allows my SwiftUI view to reactively listen to changes of isMuted because it's piped from the `@Published` through to the `@Observable` ViewModel (or whatever other scene/view architecture you use).

### Why

End goal is to let an app reactively listen to the value changing. Without Combine or AsyncStream, we'd need to build a delegate or something to reflect the updated value up to the UI.

bolsinga: I agree with what you wrote, but the library view this populates (DynamicallyOrientingNavigationView) does not need it to be Published. Does your app use DynamicallyOrientingNavigationView?

I also wonder is all this speech stuff should be in the app, and not the navigation library. The navigation library would just provide and interface for speech.

ianthetechie: > the library view this populates (DynamicallyOrientingNavigationView) does not need it to be Published

Yeah, I think that's correct, but I _think_ keeping it enables more use cases. I admittedly don't have a great one offhand, but on the other hand I don't feel any need to remove `@Published`.

> I also wonder is all this speech stuff should be in the app, and not the navigation library. The navigation library would just provide and interface for speech.

My opinion is that it should stay in the library, because:

1. The _majority_ (by number) of developers will probably use the default behavior, at least at some point. (I think the only reason they would _not_ want the default behavior is if they are bringing an external service like Amazon Polly, ElevenLabs, etc.)
2. An even larger majority of devs might end up wasting days or even weeks getting it set up correctly with audio ducking etc. (ducking specifically is on my very short term TODO list). So putting it in the library makes it a lot more valuable IMO.
3. I'm pretty sure every other navigation SDK I'm aware of also bundles TTS for various reasons. We are (AFAIK) the only one that makes it very extensible at all ;)

Archdoog: #### Part 1

My app does use the DynamicallyOrientingNavigationView, but it has a custom `@Observable class ViewModel` separating the core from the UI. Its purpose is to properly isolate the service layers (e.g. FerrostarCore) from triggering unnecessary SwiftUI view re-renders and resource burn.

It's a common situation where a vanilla variable appears to change a SwiftUI view. However, when this happens, the view is being redrawn based on a different trigger. One of the big problems with our original implementation of this library (and particularly the DemoApp) is exactly that. That our entire service layer stack is `@Published` in an `ObservableObject`. This means whenever there's a diff on FerrostarCore, the re-render occurs for whatever larger scope. Since ferrostar core contains the entire state of the library and the spoken instruction observer, it's easy to loose granularity.

The reason we want to use @Published or AsyncStream is you need an observable/reactive type for it to trigger a change _when just that value changes_. Without a reactive stream, something wrapping this value has to trigger the change.

Our demo app is a bit of a mess on this front, so no surprise it's working as expected. It's just not an effective solution for architectures that are efficiently isolating SwiftUI state.

#### Part 2

I agree with Ian that this should stay in the library for the same reasons. Most nav SDK's I've used provide a default voice implementation.

bolsinga:
Just removed `ObservableObject` conformance.

PR #598, opened 2025-06-11
Location is not used as ObservableObject

bolsinga:


bolsinga:
Just removed `ObservableObject` conformance.

Issue #592, opened 2025-06-11
Implement the Recording Mechanism


Kandarp05:
Build the core logic that captures and stores session events as they happen.

PR #599, opened 2025-06-13
Define methods to record navigation sessions.

Kandarp05:
#592 
This PR aims to define methods to record navigation sessions

### DIFF
--- a/common/ferrostar/src/navigation_controller/mod.rs
+++ b/common/ferrostar/src/navigation_controller/mod.rs
@@ -1,6 +1,7 @@
 //! The navigation state machine.
 
 pub mod models;
+pub mod recording;
 
 #[cfg(test)]
 pub(crate) mod test_helpers;
--- a/common/Cargo.lock
+++ b/common/Cargo.lock
@@ -453,7 +453,7 @@ checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
 
 [[package]]
 name = "ferrostar"
-version = "0.36.0"
+version = "0.37.0"
 dependencies = [
  "android_logger",
  "assert-json-diff",
--- a/common/ferrostar/Cargo.toml
+++ b/common/ferrostar/Cargo.toml
@@ -2,7 +2,7 @@ lints.workspace = true
 
 [package]
 name = "ferrostar"
-version = "0.36.0"
+version = "0.37.0"
 readme = "README.md"
 description = "The core of modern turn-by-turn navigation."
 keywords = ["navigation", "routing", "valhalla", "osrm"]
--- a/common/ferrostar/src/navigation_controller/models.rs
+++ b/common/ferrostar/src/navigation_controller/models.rs
@@ -282,19 +282,6 @@ pub struct NavigationControllerConfig {
     pub snapped_location_course_filtering: CourseFiltering,
 }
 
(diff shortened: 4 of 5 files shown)


Kandarp05:
This defines the core methods needed for recording, that’s all I can think of for now. I’ll have a better sense of what else is needed once I figure out how to integrate this into the navigation_controller.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -368,6 +373,55 @@ impl NavigationController {
     }
 }
 
+/// Functionality for the navigation controller that is exported. (or will be exported)
+impl NavigationRecording {
```
ianthetechie: Minor organizational nit, but I usually prefer to have any `impl`s right after the `struct` definition.

I think we can also *probably* move both of these into a `recording` module under `navigation_controller` at this point; there's enough code now, and `NavigationRecording` isn't really a model.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -368,6 +373,55 @@ impl NavigationController {
...
+    }
+
+    /// Records an error that occurred during navigation.
+    pub fn record_navigation_error(&mut self, error_message: String) {
+        self.add_event(NavigationRecordingEventData::Error { error_message });
+    }
+}
+
+/// Shared functionality for the navigation controller that is not exported.
+impl NavigationRecording {
```
ianthetechie: Another minor style nit: I don't think we need separate impls within the same file except to separate trait implementations. "Section comments" that mark off areas are enough.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -368,6 +373,55 @@ impl NavigationController {
...
+
+    /// Records a route update during navigation.
+    pub fn record_route_update(&mut self, route: Route) {
+        self.add_event(NavigationRecordingEventData::RouteUpdate { route });
+    }
+
+    /// Records an error that occurred during navigation.
+    pub fn record_navigation_error(&mut self, error_message: String) {
+        self.add_event(NavigationRecordingEventData::Error { error_message });
+    }
```
ianthetechie: This will _probably_ break once you try exporting it. UniFFI doesn't support mutable self references in exports, because there is no way to guarantee that foreign code (e.g. Swift, Kotlin, etc.) will uphold the assumptions Rust makes. That's why we'll need to either:

1. Work out an efficiently appendable data  structure where every call consumes the original value and returns a new one with each call (like we do for the rest of the pure core already), OR
2. Introduce some sort of interior mutability (obviously not ideal).

common/ferrostar/src/navigation_controller/mod.rs
Kandarp05: Refactored `impl` to be with `struct` defination

common/ferrostar/src/navigation_controller/mod.rs
Kandarp05: Removed separate impls

common/ferrostar/src/navigation_controller/mod.rs
Kandarp05: Removed mutability as discussed

ianthetechie:
Yup; you got it!

ianthetechie (approved):

PR #601, opened 2025-06-15
feat: trip summary/stats

Archdoog:
- Adds TripSummary object which contains trip metadata, useful for capturing stats during and after the trip.

### DIFF
--- a/common/ferrostar/src/algorithms.rs
+++ b/common/ferrostar/src/algorithms.rs
@@ -426,6 +426,16 @@ fn travel_distance_to_end_of_step(
         .map(|traversed| step_length - traversed)
 }
 
+/// Calculates the distance increment between two user locations.
+pub(crate) fn distance_between_locations(
+    previous_location: &UserLocation,
+    current_location: &UserLocation,
+) -> f64 {
+    let prev_point: Point = previous_location.coordinates.into();
+    let current_point: Point = current_location.coordinates.into();
+    Haversine.distance(prev_point, current_point)
+}
+
 /// Computes the user's progress along the current trip (distance to destination, ETA, etc.).
 ///
 /// NOTE to callers: `remaining_steps` includes the current step!
--- a/common/ferrostar/src/navigation_controller/mod.rs
+++ b/common/ferrostar/src/navigation_controller/mod.rs
@@ -7,11 +7,13 @@ pub(crate) mod test_helpers;
 
 use crate::{
     algorithms::{
-        advance_step, apply_snapped_course, calculate_trip_progress,
+        advance_step, apply_snapped_course, calculate_trip_progress, distance_between_locations,
         index_of_closest_segment_origin, should_advance_to_next_step, snap_user_location_to_line,
     },
     models::{Route, UserLocation},
+    navigation_controller::models::TripSummary,
 };
+use chrono::Utc;
 use geo::{
     algorithm::{Distance, Haversine},
     geometry::{LineString, Point},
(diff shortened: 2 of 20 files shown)


common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -202,6 +214,21 @@ impl NavigationController {
...
+                    &snapped_user_location,
+                ) as u32;
+
+                // Update trip summary with accumulated distance
+                let updated_summary = self.update_trip_summary(
+                    summary,
+                    distance_increment,
+                    snapped_distance_increment,
+                );
+
```
ianthetechie: I *think* this would be a bit cleaner (esp as this function is already a long one) to move the distance increment calculation inside `update_trip_summary`.

common/ferrostar/src/navigation_controller/test_helpers.rs
```diff
@@ -101,3 +103,35 @@ pub fn gen_route_from_steps(steps: Vec<RouteStep>) -> Route {
...
+}
+
+pub(crate) fn nav_controller_insta_settings() -> Settings {
+    let mut settings = Settings::new();
+    settings.add_redaction(
+        ".**.startedAt",
+        dynamic_redaction(create_timestamp_redaction()),
+    );
+    settings.add_redaction(
+        ".**.endedAt",
```
ianthetechie, line 133: 💯

guide/src/contributing-advanced.md
```diff
@@ -13,7 +13,7 @@ In both the Demo App and Maplibre UI build.gradle, replace libs.maplibre.compose
 
 ```groovy
 //    api libs.maplibre.compose
-    api 'org.maplibre.gl:android-sdk:10.3.0'
-    api 'org.maplibre.gl:android-plugin-annotation-v9:2.0.2'
+    api 'org.maplibre.gl:android-sdk:11.9.0'
```
ianthetechie, line 16: good catch ;)

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -366,6 +400,31 @@ impl NavigationController {
...
         }
     }
+
+    /// Updates the trip summary with accumulated distance and duration.
+    fn update_trip_summary(
+        &self,
+        current_summary: &TripSummary,
+        distance_increment: u32,
+        snapped_distance_increment: u32,
+    ) -> TripSummary {
```
ianthetechie: This doesn't really need to be a method on `NavigationController`, since it is a pure function referencing only its arguments.

I think this might actually make more sense to have as a method on `TripSummary` directly, which consumes `self` (owned `self`; not a reference) and returns a new value. Something like:

`fn update(self,  previous_location: &UserLocation, current_location: &UserLocation) -> Self`

common/ferrostar/src/algorithms.rs
```diff
@@ -426,6 +426,16 @@ fn travel_distance_to_end_of_step(
         .map(|traversed| step_length - traversed)
 }
 
+/// Calculates the distance increment between two user locations.
```
ianthetechie:
```suggestion
/// Calculates the distance (in meters) between two user locations.
```

Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 
 let binaryTarget: Target
 let maplibreSwiftUIDSLPackage: Package.Dependency
-let useLocalFramework = false
+let useLocalFramework = true
```
ianthetechie: Whoops ;)

Archdoog: 😭

ianthetechie (approved):

PR #603, opened 2025-06-18
Associate VisualInstruction with CPManeuver

bolsinga:
- This will allow future work to see if the current `VisualInstruction` was made to create a `CPManeuver`.
- It uses `CPManeuver.userInfo` to store this data.

### DIFF
--- a/apple/Sources/FerrostarCarPlayUI/CarPlayModels/CPManeuver.swift
+++ b/apple/Sources/FerrostarCarPlayUI/CarPlayModels/CPManeuver.swift
@@ -3,6 +3,30 @@ import FerrostarCore
 import FerrostarCoreFFI
 import FerrostarSwiftUI
 
+private let InstructionKey = "com.stadiamaps.ferrostar.instruction"
+
+extension CPManeuver {
+    private var userDictionary: [String: Any]? {
+        get {
+            userInfo as? [String: Any] ?? [:]
+        }
+        set {
+            userInfo = newValue
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


apple/Sources/FerrostarCarPlayUI/CarPlayModels/CPManeuver.swift
```diff
@@ -3,6 +3,30 @@ import FerrostarCore
...
+    }
+
+    public var visualInstruction: VisualInstruction? {
+        get {
+            userDictionary?[InstructionKey] as? VisualInstruction
+        }
+        set {
+            var info = userDictionary ?? [:]
+            info[InstructionKey] = newValue
+            userDictionary = info
```
ianthetechie: Is there a better way to do this nowadays? This looks like a data race waiting to happen.

bolsinga: This value is set once, read more often. The property is not a part of the initializer, unfortunately.

bolsinga: This type doesn't have a specific initializer, so it's a bunch of properties to set just once and get thereafter.

bolsinga: Thanks for the inspiration to make the patch better!

ianthetechie (approved):
Ooh yes; massive improvement! Thanks!

PR #604, opened 2025-06-18
Associate Route with CPRouteChoice

bolsinga:
- This will allow future work to get the `Route` used to create a `CPRouteChoice` which will arrive from `CPMapTemplateDelegate` functions.
- It uses `CPRouteChoice.userInfo` to store this data.

### DIFF
--- a/apple/Sources/FerrostarCarPlayUI/CarPlayModels/CPTrip.swift
+++ b/apple/Sources/FerrostarCarPlayUI/CarPlayModels/CPTrip.swift
@@ -5,16 +5,23 @@ import FerrostarSwiftUI
 private let RouteKey = "com.stadiamaps.ferrostar.route"
 
 extension CPRouteChoice {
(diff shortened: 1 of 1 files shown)


apple/Sources/FerrostarCarPlayUI/CarPlayModels/CPTrip.swift
```diff
@@ -2,6 +2,30 @@ import CarPlay
...
+    }
+
+    public var route: Route? {
+        get {
+            userDictionary?[RouteKey] as? Route
+        }
+        set {
+            var info = userDictionary ?? [:]
+            info[RouteKey] = newValue
+            userDictionary = info
```
ianthetechie: Same question as asked on #603.

bolsinga: This value is set once, read more often. The property is not a part of the initializer, unfortunately.

Issue #602, opened 2025-06-18
Make higher levels work off a trait rather than a concrete navigation controller


Kandarp05:

ianthetechie:
Link from the call today: https://github.com/stadiamaps/ferrostar/blob/3d15c7519eacf78598d7d77764a701803e1c1516/common/ferrostar/src/navigation_controller/models.rs#L17-L58.

Kandarp05:
>The best path I can see forward which preserves purity is to add an [associated type](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html) (preferred) or generic (if necessary) to the trait. I don't actually know what UniFFI is capable of expressing here, so it's possible we hit a limitation with associated types, but generics should work.

![Image](https://github.com/user-attachments/assets/15bc5683-6ba9-47d0-af33-207682df4ac6)

![Image](https://github.com/user-attachments/assets/1504830e-7782-47ab-a3fa-7ac47428d2c9)

UniFFI doesn't seem to like either of those.

PR #605, opened 2025-06-19
Change higher levels to use trait instead of concrete NavigationController

Kandarp05:
#602

### DIFF
--- a/common/ferrostar/src/navigation_controller/mod.rs
+++ b/common/ferrostar/src/navigation_controller/mod.rs
@@ -39,9 +39,9 @@ pub struct NavigationController {
 
 /// Core interface for navigation functionalities.
 ///
-/// This trait defines the essential operations for any navigation implementation
-/// allowing different navigation strategies (standard, recording)
-/// to be used interchangeably
+/// This trait defines the essential operations for a navigation state manager.
+/// This lets us build additional layers (e.g. event logging)
+/// around [`NavigationController`] a composable manner.
 #[cfg_attr(feature = "uniffi", uniffi::export)]
 pub trait Navigator: Send + Sync {
     fn get_initial_state(&self, location: UserLocation) -> TripState;
--- a/common/ferrostar/src/navigation_controller/mod.rs
+++ b/common/ferrostar/src/navigation_controller/mod.rs
@@ -78,7 +78,7 @@ impl NavigationController {
     }
 }
 
-/// Shared functionality for the navigation controller that is not exported by uniFFI.
+/// Shared functionality for the navigation controller that is not exported by UniFFI.
 impl NavigationController {
     /// Snaps the user's location to the route line and updates the user's course if necessary.
     ///
--- a/common/ferrostar/src/navigation_controller/mod.rs
+++ b/common/ferrostar/src/navigation_controller/mod.rs
@@ -21,7 +21,7 @@ use geo::{
 };
 use models::{NavigationControllerConfig, StepAdvanceStatus, TripState, WaypointAdvanceMode};
 use std::clone::Clone;
-
+use std::sync::Arc;
 #[cfg(feature = "wasm-bindgen")]
 use wasm_bindgen::{prelude::wasm_bindgen, JsValue};
 
@@ -35,6 +35,39 @@ use wasm_bindgen::{prelude::wasm_bindgen, JsValue};
 pub struct NavigationController {
     route: Route,
     config: NavigationControllerConfig,
+}
+
+/// Core interface for navigation functionalities.
+///
+/// This trait defines the essential operations for any navigation implementation
+/// allowing different navigation strategies (standard, recording)
+/// to be used interchangeably
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+pub trait Navigator: Send + Sync {
+    fn get_initial_state(&self, location: UserLocation) -> TripState;
+    fn advance_to_next_step(&self, state: &TripState) -> TripState;
+    fn update_user_location(&self, location: UserLocation, state: &TripState) -> TripState;
+}
+
+/// Creates a new navigation controller for the given route and configuration.
+///
+/// It returns an Arc-wrapped trait object implementing `Navigator`.
+/// If `should_record` is true, it creates a controller that can record navigation events.
+/// Else it creates a regular controller without recording capabilities.
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+pub fn create_navigation_controller(
+    route: Route,
+    config: NavigationControllerConfig,
+    should_record: bool,
+) -> Arc<dyn Navigator> {
+    if should_record {
+        // Creates a navigation controller with recording capabilities.
+        // For now, it just returns the regular controller
+        Arc::new(NavigationController { route, config })
+    } else {
+        // Creates a navigation controller without recording capabilities.
+        Arc::new(NavigationController { route, config })
+    }
 }
 
 #[cfg_attr(feature = "uniffi", uniffi::export)]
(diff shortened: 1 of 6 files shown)


Kandarp05:
@ianthetechie  @Archdoog  Am i going in the right direction?

ianthetechie:
Yup, this is going the right direction! Thanks for the draft + ping!

From here, we'll need to make a way to use the logging version (maybe a boolean parameter to the function that returns these?) and, one of the classic unsolvable problems of computer science, better names ;)

Kandarp05:
Added a boolean parameter should_record, also renamed `NavigationControllerTrait` to be `Navigator`, short and simple, let me know about your thoughts on this

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -37,6 +37,39 @@ pub struct NavigationController {
...
+    fn get_initial_state(&self, location: UserLocation) -> TripState;
+    fn advance_to_next_step(&self, state: &TripState) -> TripState;
+    fn update_user_location(&self, location: UserLocation, state: &TripState) -> TripState;
+}
+
+/// Creates a new navigation controller for the given route and configuration.
+///
+/// It returns an Arc-wrapped trait object implementing `Navigator`.
+/// If `should_record` is true, it creates a controller that can record navigation events.
+/// Else it creates a regular controller without recording capabilities.
```
ianthetechie: Minor edits:

```suggestion
/// If `should_record` is true, it creates a controller with event recording enabled.
```

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -37,6 +37,39 @@ pub struct NavigationController {
...
+    route: Route,
+    config: NavigationControllerConfig,
+    should_record: bool,
+) -> Arc<dyn Navigator> {
+    if should_record {
+        // Creates a navigation controller with recording capabilities.
+        // For now, it just returns the regular controller
+        Arc::new(NavigationController { route, config })
+    } else {
+        // Creates a navigation controller without recording capabilities.
```
ianthetechie: It's possible that other variations will be created later. The revised comment is a bit more future-proof.

```suggestion
        // Creates a normal navigation controller.
```

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -37,6 +37,39 @@ pub struct NavigationController {
...
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+pub fn create_navigator(
+    route: Route,
+    config: NavigationControllerConfig,
+    should_record: bool,
+) -> Arc<dyn Navigator> {
+    if should_record {
+        // Creates a navigation controller with recording capabilities.
+        // For now, it just returns the regular controller
+        Arc::new(NavigationController { route, config })
```
ianthetechie:
```suggestion
        // Creates a navigation controller with a wrapper that records events.
        // TODO: Currently just returns the regular controller
        Arc::new(NavigationController { route, config })
```

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -37,6 +37,39 @@ pub struct NavigationController {
     config: NavigationControllerConfig,
 }
 
+/// Core interface for navigation functionalities.
+///
+/// This trait defines the essential operations for any navigation implementation
+/// allowing different navigation strategies (standard, recording)
+/// to be used interchangeably
```
ianthetechie: Minor nits; the comment as written sounds very mechanical.

```suggestion
/// This trait defines the essential operations for a navigation state manager.
/// This lets us build additional layers (e.g. event logging)
/// around [`NavigationController`] a composable manner.
```

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -405,6 +438,20 @@ impl NavigationController {
...
+    }
+
+    fn advance_to_next_step(&self, state: &TripState) -> TripState {
+        self.advance_to_next_step(state)
+    }
+
+    fn update_user_location(&self, location: UserLocation, state: &TripState) -> TripState {
+        self.update_user_location(location, state)
+    }
+}
```
ianthetechie: Rather than making methods that just call other methods by the same name, moving the implementation under the trait is more idiomatic.

(I'm actually surprised that this is legal; the signatures are the same, but yeah Rust traits can get a bit weird...)

ianthetechie (commented):
Looking good; mostly a bunch of minor nitpicks now.

* No update for Swift? I know you don't have a Mac, but this should be a trivial change along the lines of what you already did for Kotlin. We can clean up any errors if needed.
* The documentation (in the `guide` folder) needs to be updated, since the `FerrostarCore` constructors now require an extra argument.

common/ferrostar/src/navigation_controller/mod.rs
Kandarp05: fixed

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -37,16 +37,114 @@ pub struct NavigationController {
...
 #[cfg_attr(feature = "uniffi", uniffi::export)]
 impl NavigationController {
     #[cfg_attr(feature = "uniffi", uniffi::constructor)]
     /// Create a navigation controller for a route and configuration.
     pub fn new(route: Route, config: NavigationControllerConfig) -> Self {
         Self { route, config }
     }
+}
+
+/// Shared functionality for the navigation controller that is not exported by uniFFI.
```
ianthetechie: Not your comment, but since it showed up in the diff... ;)

```suggestion
/// Shared functionality for the navigation controller that is not exported by UniFFI.
```

ianthetechie:
> Also the web is still using the old way, can you explain me how the JsNavigationController works and how I can update it to use the trait?

Yeah, we can walk through this on the call tomorrow as it's probably easier that way. (Probably not a bad side task for you to improve documentation around this too :D I don't think our GSoC student last year really documented this anywhere.)

ianthetechie (commented):
I think this is pretty good from my side. We can discuss JS tomorrow on the call. Until then, I'll leave this open so @Archdoog can have a quick look over too.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -37,16 +37,114 @@ pub struct NavigationController {
...
+    fn advance_to_next_step(&self, state: &TripState) -> TripState;
+    fn update_user_location(&self, location: UserLocation, state: &TripState) -> TripState;
+}
+
+/// Creates a new navigation controller for the given route and configuration.
+///
+/// It returns an Arc-wrapped trait object implementing `Navigator`.
+/// If `should_record` is true, it creates a controller with event recording enabled.
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+pub fn create_navigator(
```
Archdoog, line 57: I'm fine with this as is for the purposes of what we're building. As an enhancement, we could add a TODO to create a NavigationController builder. I feel like this would be a nice solution on the platform side where a developer could do something like:

```kt
NavigationController.Builder
    .enableRecording() // this function could even take the format/file. E.g. .enableRecording("some/folder/ferrostar.export")
    .build()
```


```kt
NavigationController.Builder
    .build() // Same as NavigationController()
```

Archdoog:
I'd agree. Looking fine to me. Some considerations, but none are super important at this point 👍

common/ferrostar/src/navigation_controller/mod.rs
ianthetechie: Ooh, that's a fantastic idea! Either I'm forgetting some drawback of UniFFI that makes this impossible, or we should apply this everywhere ;) Let's give this a shot and even if it doesn't work out to expose in the UniFFI layer, we can probably make some more ergonomic APIs at the Swift and Kotlin level for better discoverability.

PR #607, opened 2025-06-23
Fix nondeterminism in SwiftUI tests (use explicit locale distance formatters)

ianthetechie:
* Adds the `TestSupport` package to the CarPlay UI tests, and uses the formatter from there (FYI @bolsinga; the defaults are perfect for "real world" use, but unfortunately nondeterministic in tests, so these will fail for any developer without a US English locale; we have a set of test formatters that work around this by setting a specific locale)
* Rename ambiguous "american" distance formatter
* Make the test distance formatters match the behavior that you'd see from a default formatter on device with abbreviated units (thanks to the CarPlay test for catching this!)


bolsinga:
Nice Change!

PR #608, opened 2025-06-23
Duck audio from other apps when speaking voice instructions on iOS

ianthetechie:
Does what it says on the tin ;)

I can walk through any questions on the dev call tomorrow, as there are a bunch of ways you could get this wrong. I have done a moderate amount of testing on-device and am pretty sure this works as desired.

Suggestions for comments on anything that isn't obvious on a first read are very welcome.

### DIFF
--- a/apple/Sources/FerrostarCore/Speech/AudioSessionManager.swift
+++ b/apple/Sources/FerrostarCore/Speech/AudioSessionManager.swift
@@ -1,0 +1,45 @@
+import AVFoundation
+
+/// Manages the `AVAudioSession` for the app,
+/// allowing the ``SpokenInstructionObserver`` to request
+/// "focus." This will duck audio from most other apps (e.g. Music),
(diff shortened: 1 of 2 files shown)


apple/Sources/FerrostarCore/Speech/AudioSessionManager.swift
```diff
@@ -0,0 +1,45 @@
+import AVFoundation
+
+/// Manages the `AVAudioSession` for the app,
+/// allowing the ``SpokenInstructionObserver`` to request
+/// "focus." This will duck audio from most other apps (e.g. Music),
+/// and interrupt speech for apps playing back spoken audio (e.g. Podcasts).
+///
+/// Automatically releases focus on deinit.
+actor AudioSessionManager {
```
Archdoog, line 9: 👌

Archdoog (approved):
Nice to see a good use of an actor 👍

PR #609, opened 2025-06-23
Web JS package updates

ianthetechie:
- Updates to the latest formatter package release (https://github.com/maptimy/platform-formatters/pull/2/files; cc @tedi4t)
- Regular `npm update`

As far as I can tell, this package isn't actually used in React Native yet; can you confirm @bjtrounson?


bjtrounson:
Yeah I don't think it is used there yet

Archdoog (approved):

PR #610, opened 2025-06-23
Duck audio from other apps when speaking voice instructions on Android

ianthetechie:
Same as #608, but for Android, and using an almost identical approach.

Main differences:
* Kotlin doesn't have actor isolation
* Android makes it a bit more convenient to register a listener for TTS
* We still support Android API 25, so we need a bit of runtime checking... any Android devs have an updated opinion on whether API 26 as a baseline is feasible yet?


Archdoog:
Just for reference. This is a good example of synchronized access in android/kotlin: https://github.com/googlemaps/android-maps-compose/blob/7df3b17e4df49f45bf50a2feaaeac3860969e404/maps-compose/src/main/java/com/google/maps/android/compose/CameraPositionState.kt#L154.

Since there's no swift actor.

PR #612, opened 2025-06-24
Update UniFFI to 0.29.3

bjtrounson:
Title describes it. 

I've just bumped the version we will see what the CI\CD does, it will be nice if it just works.


bjtrounson:
Okay I seemed to have fixed the above issue with adding `remote` to the custom types since they are in separate crates. There is still some other stuff failing but we are making progress.

bjtrounson:
Got CI passing 🎉

PR #614, opened 2025-06-24
Update NavigationRecording struct

Kandarp05:
Since we are defining things like NavigatorState, a name like InitialNavigationState could give a wrong idea to someone, so just a small update


ianthetechie (approved):

Archdoog (approved):

ianthetechie:
Will merge this once we are in a good state to do a release (since this necessarily bumps the version).

ianthetechie:
Sorry for the delay on this; I think with #620 resolved we can merge this now and cut a release!

PR #615, opened 2025-06-24
iOS: Allow CPTrip to be created with origin / destination

bolsinga:
- This is because taking the Waypoints from the first Route may not be correct? It's confusing because I do not know how multiple routes in the response may or may not have differing origin/destinations. This way the app can decide what the origin/destination are of the CPTrip, which may be different than the multiple CPRouteChoices to use...


Archdoog (approved):
Looks good to me 👍

PR #617, opened 2025-06-25
iOS: Add CustomStringConvertible to FerrostarCore types

bolsinga:
This can be used for logging or for a UI that shows detailed textual state for debugging purposes.


apple/Sources/UniFFI/Descriptions.swift
Archdoog, line 1: I'd probably put this in the same folder as `apple/Sources/FerrostarCore/Descriptions.swift`. The UniFFI folder has historically been exclusively for generated swift content. There are already several extensions and wrappers for uniffi generated types in the FerrostarCore target.

Thoughts @ianthetechie?

bolsinga: I recall why I had to do that. Because if I declare conforming to a protocol outside the library (Ferrorstar vs. FerrorstarFFI) I have to add @retroactive . Which leads me to: why is FerrostarFFI just "in" FerrorStar? In iOS, why would anyone use one w/o the other?

Archdoog: Diving into our build process. This explains why: 

https://github.com/stadiamaps/ferrostar/blob/21f45e4d12da0b278101e7384a4e16eecfa89209/common/build-ios.sh#L42-L57

It looks like the swift phase of `uniffi-bindgen-swift` auto generates whatever files it decides in the specified output folder. Because of this we just use a clean folder and move whatever it created into there. Additional info: https://mozilla.github.io/uniffi-rs/next/swift/uniffi-bindgen-swift.html. 

While it'd be nice to avoid `@retroactive`, I think it's the simpler move in this case.

bolsinga: Aha. Makes sense. update coming

Archdoog (approved):

PR #618, opened 2025-06-27
iOS: Add Observable DemoModel

bolsinga:
- Naming things is hard: This adds `DemoAppState`. It is the big-picture states of a navigation application. It iterates through .idle, .chooseDestination, load .routes, .selectRoute, .startNavigation
- The Demo app doesn't have a way (right now) to pick a destination, nor choose from multiple routes. And the plan it loads only ever returns one route. However a Navigation App and library needs to support all of these states. So right now this adds a button to move the UI through these states. This will be *very useful* for having this state in CarPlay.
- This is odd UI, but the the improvements to the DemoApp have to start somehow. The Choose Destination & Select Route buttons just move the state through using the same hard coded values as before.
- This also separates the parts out into separate files.
- It removes the need for `AppEnvironment`, and `@EnvironmentObject` completely.
- The minimum OS version for the DemoApp (but not the Ferrostar library) is iOS17, so this uses the modern `@Observable` for the `DemoModel`.
- Since `FerrorstarCore` must remain `ObservableObject` and use Combine (continuing to support iOS16), use Combine to subscribe to the Published properties and then just have it set Observable properties in the `DemoModel`.


bolsinga:
Here is a recording.

https://github.com/user-attachments/assets/c62f855f-e93b-4bad-90ed-e5fe74650273

apple/DemoApp/Demo/DemoModel.swift
```diff
@@ -0,0 +1,193 @@
...
+            errorMessage = nil
+            appState = try await wrap()
+        } catch {
+            errorMessage = error.localizedDescription
+            appState = .idle
+        }
+    }
+
+    func chooseDestination() {
+        wrap {
```
ianthetechie, line 157: Very cool pattern 😎

Archdoog (approved):

apple/DemoApp/Demo/DemoNavigationView.swift
```diff
@@ -171,6 +172,6 @@ struct DemoNavigationView: View {
     }
 }
 
-#Preview {
-    DemoNavigationView()
-}
+// #Preview {
+//    DemoNavigationView()
+// }
```
ianthetechie:
```suggestion
```

apple/DemoApp/Demo/DemoModel.swift
bolsinga: 😀thanks!I imagine there's a more compact way to achieve this.

PR #619, opened 2025-06-29
Add recording events to NavState

Kandarp05:
#602


common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -44,9 +45,9 @@ pub struct NavigationController {
 /// around [`NavigationController`] a composable manner.
 #[cfg_attr(feature = "uniffi", uniffi::export)]
 pub trait Navigator: Send + Sync {
-    fn get_initial_state(&self, location: UserLocation) -> TripState;
-    fn advance_to_next_step(&self, state: &TripState) -> TripState;
-    fn update_user_location(&self, location: UserLocation, state: &TripState) -> TripState;
+    fn get_initial_state(&self, location: UserLocation) -> NavigatorState;
+    fn advance_to_next_step(&self, state: &TripState) -> NavigatorState;
+    fn update_user_location(&self, location: UserLocation, state: &TripState) -> NavigatorState;
```
Kandarp05: These still take TripState as argument, should they be changed to take NavigatorState instead??

Kandarp05:
Since neither generic or associated types work with UniFFI, I have created a `NavigatorState`, using which we can send anything extra along with `TripState`, similar to the `NavState` you guys were showing

Is this the right way to go??

android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
```diff
@@ -134,6 +135,7 @@ class FerrostarCore(
   private val _executor = Executors.newSingleThreadScheduledExecutor()
   private val _scope = CoroutineScope(Dispatchers.IO)
   private var _navigationController: Navigator? = null
+  private var _navState: MutableStateFlow<NavState?> = MutableStateFlow(null)
```
ianthetechie:
```suggestion
  // TODO: Come back and clean this up later so we don't have two confusingly named types
  private var _navState: MutableStateFlow<NavState?> = MutableStateFlow(null)
```

android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
```diff
@@ -247,21 +249,24 @@ class FerrostarCore(
     // Apply the new config if provided, otherwise use the original.
     _config = config ?: _config
 
+    // TODO: This is always false
```
ianthetechie:
```suggestion
    // TODO: This is always false for now while it's a work in progress
```

android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
```diff
@@ -281,11 +286,13 @@ class FerrostarCore(
...
```
ianthetechie:
```suggestion
    // TODO: Similar to the above (should probably be extracted to an instance variable or something later)
```

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -419,6 +425,7 @@ impl JsNavigationController {
         serde_wasm_bindgen::to_value(&result).map_err(|e| JsValue::from_str(&format!("{:?}", e)))
     }
 
+    #[wasm_bindgen(js_name = advanceToNextStep)]
```
ianthetechie, line 428: Good catch!

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -356,13 +363,21 @@ impl From<JsNavigationControllerConfig> for NavigationControllerConfig {
     }
 }
 
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[derive(Clone)]
+#[cfg_attr(any(feature = "wasm-bindgen", test), derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "wasm-bindgen", derive(Tsify))]
```
ianthetechie, line 369: I'm not sure we _need_ to expose these publicly (via either unqualified `pub`, UniFFI, or wasm-bindgen). These feel like implementation details. I'm fin to merge and correct in a future PR though

ianthetechie:
Okay, I'm now extremely confused why this had some leftover comments on files that... haven't even changed in this PR?

ianthetechie (approved):

PR #630, opened 2025-07-02
Add serialization for Recording

### LINKED ISSUE #594: Implement Serialization for Recorded Session Data
Implement code to serialize the recorded data to JSON

Kandarp05:
Fixes #594 
Adds a method to add serialization to pretty JSON format (we could also use normal JSON??)


common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -18,7 +22,35 @@ pub struct NavigationRecording {
...
+}
+
+/// Functionality for the navigation controller that is exported.
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+impl NavigationRecording {
+    /// Serialize the recording to a pretty JSON format.
+    /// Returns a Result with the JSON string or an error.
+    pub fn to_json(&self) -> Result<String, NavigationRecordingError> {
+        serde_json::to_string_pretty(self).map_err(|error| NavigationRecordingError::from(error))
+    }
```
ianthetechie: Can you explain how this would be used? I'm not sure anyone would ever be able to read this if they tried, so let's just keep with the default (condensed) formatting 😂

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -18,7 +22,35 @@ pub struct NavigationRecording {
...
+}
+
+/// Implement conversion from serde_json::Error to NavigationRecordingError.
+impl From<serde_json::Error> for NavigationRecordingError {
+    fn from(e: serde_json::Error) -> Self {
+        NavigationRecordingError::SerializationError {
+            error: e.to_string(),
+        }
+    }
+}
```
ianthetechie: I think that having a single "type-erased" string is probably not ideal. We should let `thiserror` handle the deriving and make specific "sub-error" variants.

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -3,6 +3,10 @@ use crate::navigation_controller::models::{
     NavigationControllerConfig, NavigationRecordingEvent, NavigationRecordingEventData, TripState,
 };
 use chrono::Utc;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Object))]
 pub struct NavigationRecording {
```
ianthetechie: Question: does this need to be a public struct + UniFFI object? I'm not really sure if the callers need to care about this implementation detail. I guess they would just want to get the JSON data.

common/ferrostar/src/navigation_controller/recording.rs
Kandarp05: UniFFI can't export types like `serde_json::Error`, same is mentioned here
https://github.com/stadiamaps/ferrostar/blob/04a713a6d746ca00887865a50b3ff79d333a5977/common/ferrostar/src/routing_adapters/error.rs#L4-L16

Kandarp05:
Tried to fix this
1. `NavigationRecording` now has `JsNavigationControllerConfig` instead of `NavigationControllerConfig` to make it serializable.
2. `to_json` is not longer exported via UniFFI.
3. Fixed the error type.

common/ferrostar/src/navigation_controller/recording.rs
Kandarp05: This is no longer a problem as we are not exporting it anymore

ianthetechie:
I'm not quite sure why `#[cfg(feature = "wasm-bindgen")]`was removed in several places across this PR. That annotation ensures that JS-specific stuff isn't compiled for other targets.

Maybe `StepAdvanceConditionJsConvertible` should be gated at the trait level too... not sure why it was just the method before (probably my code :D).

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -1,15 +1,21 @@
...
     /// Version of Ferrostar that created this recording.
     pub version: String,
     /// The timestamp when the navigation session started.
     pub initial_timestamp: i64,
     /// Configuration of the navigation session.
-    pub config: NavigationControllerConfig,
+    ///
+    /// NOTE: This has to be the `JsNavigationControllerConfig` since `NavigationControllerConfig` can't be Serialized
+    pub config: JsNavigationControllerConfig,
```
ianthetechie: Ahh, maybe this is why... that makes sense actually.

Let me think about this a bit. I think this is fine, but maybe we just need a better name for the type.

Kandarp05: Yup this is exactly why.

I agree that we need to come up with better names for these, maybe something like `SerializableNavigationControllerConfig`?? (honestly sounds too long)

Maybe we can rename `NavigationControllerConfig` to `NavigatorConfig`, that would shorten it??

ianthetechie: Actually `SerializableNavigationControllerConfig` is exactly what I thought of too 😂  I think it's fine to go with that for now.

ianthetechie (approved):
Ok; I think this is good now; thanks!

Issue #164, opened 2024-07-23
Swift 6 warning: Main actor-isolated method cannot satisfy nonisolated protocol requirement


CatMe0w:
https://github.com/stadiamaps/ferrostar/blob/b8d925a991e04d529f275a36e1162441da07ef4a/apple/Sources/FerrostarCore/FerrostarCore.swift#L349

Xcode 16 gives this warning for this line in FerrostarCore package:

`Main actor-isolated instance method 'locationManager(_:didUpdateLocations:)' cannot be used to satisfy nonisolated protocol requirement; this is an error in the Swift 6 language mode`

ianthetechie:
That's correct. It's a known limitation. As far as we are aware, the current code is safe, but due to a variety of reasons, it's not possible for us to satisfy the Swift 6 strict concurrency checker ;) Long-term we want to transition this to an actor type, but it's not high enough priority quite yet.

ianthetechie:
We should probably coordinate on a timing of a full Swift 6 transition at some point. I don't think it's an urgent item as we can get pretty far with Swift 5.10, but we'll eventually need to switch.

Any opinions on this @Archdoog @hactar?

ianthetechie:
(Also of note: I haven't looked into how far along the path UniFFI is, so there will still probably be a 5.10 module for a while)

ianthetechie:
One more thing to note: the types in UniFFI are in fact marked `Sendable`, but only when using the Swift 6 compiler. So for the moment, the demo app marks these imports as `@preconcurrency`. We can revisit this later.

Import hacks added in #632.

PR #632, opened 2025-07-04
iOS: Mark FerrostarCoreFFI as @preconcurrency

bolsinga:
This will then allow DemoAppState to be interpreted as Sendable, thus removing the warning from DemoError. 

Route in FerrorstarCore is non-Sendable at the moment.

```
/Users/bolsinga/Documents/code/git/ferrostar/apple/DemoApp/Demo/DemoError.swift:9:10: error: associated value 'invalidState' of 'Sendable'-conforming enum 'DemoError' has non-sendable type 'DemoAppState'
    case invalidState(DemoAppState)
         ^
/Users/bolsinga/Documents/code/git/ferrostar/apple/DemoApp/Demo/DemoAppState.swift:5:6: note: consider making enum 'DemoAppState' conform to the 'Sendable' protocol
enum DemoAppState: CustomStringConvertible {
     ^
                                          , Sendable
```

### DIFF
--- a/apple/DemoApp/Demo/DemoAppState.swift
+++ b/apple/DemoApp/Demo/DemoAppState.swift
@@ -1,4 +1,5 @@
 import CoreLocation
+
 // Temporary measure until we transition the demo app to Swift 6.
 // See https://github.com/stadiamaps/ferrostar/issues/164#issuecomment-3037767025
 @preconcurrency import FerrostarCoreFFI
--- a/apple/DemoApp/Demo/DemoAppState.swift
+++ b/apple/DemoApp/Demo/DemoAppState.swift
@@ -1,4 +1,6 @@
 import CoreLocation
+// Temporary measure until we transition the demo app to Swift 6.
+// See https://github.com/stadiamaps/ferrostar/issues/164#issuecomment-3037767025
 @preconcurrency import FerrostarCoreFFI
 import Foundation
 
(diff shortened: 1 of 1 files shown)


ianthetechie:
Gotcha. Yeah, I guess this is because we currently use the Swift 5 compiler for the demo app. The code is already there in the auto-generated ferrostar.swift to add conformances so this is safe, but shouldn't be necessary long term.

apple/DemoApp/Demo/DemoAppState.swift
```diff
@@ -1,5 +1,5 @@
 import CoreLocation
-import FerrostarCoreFFI
+@preconcurrency import FerrostarCoreFFI
```
ianthetechie, line 5:
```suggestion
// Temporary measure until we transition the demo app to Swift 6.
// See https://github.com/stadiamaps/ferrostar/issues/164#issuecomment-3037767025
@preconcurrency import FerrostarCoreFFI
```

ianthetechie (approved):

bolsinga:
> Gotcha. Yeah, I guess this is because we currently use the Swift 5 compiler for the demo app. The code is already there in the auto-generated ferrostar.swift to add conformances so this is safe, but shouldn't be necessary long term.

OK. However, this is what I had to add in order to build with the Demo App set to Swift 6. The library still builds with Swift 5. I wonder if the library built with Swift 6, that the `Route` wouldn't have to be declared `Sendable` to compile properly?

ianthetechie:
> I wonder if the library built with Swift 6, that the Route wouldn't have to be declared Sendable to compile properly?

That's my understanding. We can revisit again soon. It's not a huge priority for me since they are interoperable and my read is that most devs haven't completely adopted Swift 6 yet. But I _think_ that we are pretty close to being able to do so. Feel free to leave any thoughts in the #164 issue.

PR #638, opened 2025-07-07
iOS: Update DemoApp CarPlay Support

bolsinga:
- Add `DemoCarPlayModel`. It has a `DemoModel` and the CarPlay types. 
-- It is created from the shared `DemoModel` in the iOS application. The code must be able to exist w/o a `UIApplication`, so this global data is necessary.
- Update the `CarPlayDelegate` to use the `DemoCarPlayModel`.
- The SwiftUI Views in CarPlay use `@Observable` properties to populate the SwiftUI views.
- No longer uses the old code path; but that is currently kept for backwards compatibility. 
-- Basically the library has helpers, since a library should not implement an application level delegate. The client of the library has to do work to make the proper decisions for their application's use cases. 
-- The `DemoAppState` is tied up in what CarPlay UI to show. It's been a challenge to determine how to library-ify this code.

### DIFF
--- a/apple/DemoApp/Demo/CarPlaySceneDelegate.swift
+++ b/apple/DemoApp/Demo/CarPlaySceneDelegate.swift
@@ -10,16 +10,16 @@ private extension Logger {
     static let carPlay = Logger(subsystem: "ferrostar", category: "carplaydelegate")
 }
 
-private let CarPlaySceneDelegateKey = "ferrostar"
+private let ModelKey = "com.stadiamaps.ferrostar.model"
 
 private extension UISceneSession {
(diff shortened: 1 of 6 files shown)


bolsinga:
https://github.com/user-attachments/assets/49a792b8-4932-4248-a2be-f64ec06cdc78

Putting this PR up for discussion with @Archdoog and @ianthetechie

ianthetechie (approved):
Nice! Coming along :)

Agreed on the approach; let's build first and figure out how to extract components into a library as we get a fuller demo app experience.

Archdoog (approved):
Reviewed on ferrostar call. Looking sweet 👍

PR #642, opened 2025-07-08
Feature/navigation state trait

Archdoog:


common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -141,7 +141,7 @@ impl Navigator for NavigationController {
             annotation_json,
         };
         let next_advance = Arc::clone(&self.config.step_advance_condition);
-        NavState::new(trip_state, next_advance)
+        NavState::new(trip_state, next_advance, None)
```
ianthetechie, line 144: Question for this (and maybe the one above?) as they are both in `get_initial_state`:

Do we actually want/need the ability to set recording events with `NavState::new`? I guess this might be for cases where we may carry over a session across multiple controller instances (a new "trip" due to recalculating)?

ianthetechie:
I've resolved all the merge conflicts, so despite this now being a silly number of commits, I think this is good for discussion. Can you look over this before the meeting tomorrow @Kandarp05?

common/ferrostar/src/navigation_controller/mod.rs
Kandarp05: Yup I have to agree with this, we don't need to set `recording_events` with `new` (or the other ones) and just default to `None`, and we can have something like `NavState::add_recording` which can set it??

PR #643, opened 2025-07-08
iOS: Add Route Preview to Demo App and CarPlay

bolsinga:
- It's Green, because why not?
- In `DemoModel`, the `MapViewCamera` is overridden if there is a `selectedRoute`. The `MapViewCamera` will then be the bounds of the `Route`. 
-- This works in the iOS application via `DynamicallyOrientingNavigationView`, but does **not** work in `CarPlayNavigationView`. It's not clear to me why yet.


bolsinga:
Hmm. I noticed CarPlay doing the "right thing" if I run navigation by tapping in the iOS app. Let me look at this more before reviewing.

bolsinga:
Also misbehaves in iOS when zooming the map while the route preview is showing....

PR #652, opened 2025-07-14
iOS: Demo Models funnel Errors through a single helper func

bolsinga:
- reduces redundant code so mistakes are less likely...

### DIFF
--- a/apple/DemoApp/Demo/DemoModel.swift
+++ b/apple/DemoApp/Demo/DemoModel.swift
@@ -140,9 +140,9 @@ extension DemoModel {
         return .idle
     }
 
-    func handleError(_ error: Error) {
+    func handleError(_ error: Error, newAppState: DemoAppState = .idle) {
         errorMessage = error.localizedDescription
-        appState = .idle
+        appState = newAppState
     }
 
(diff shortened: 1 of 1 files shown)


apple/DemoApp/Demo/DemoModel.swift
```diff
@@ -140,13 +140,17 @@ extension DemoModel {
         return .idle
     }
 
+    func handleError(_ error: Error) {
+        errorMessage = error.localizedDescription
+        appState = .idle
+    }
```
Archdoog: Something like this may be useful _if_ this handle error function is potentially valuable during navigation as well. Ok if it's not.

```suggestion
    func handleError(_ error: Error, newAppState: DemoAppState = .idle) {
        errorMessage = error.localizedDescription
        appState = newAppState
    }
```

Archdoog (approved):
One idea suggestion, otherwise lgtm

apple/DemoApp/Demo/DemoModel.swift
bolsinga: Oh! the DemoApp just bails to `.idle` when things go wrong. It definitely could get better.

Issue #640, opened 2025-07-08
Recording in the web app (first pass)


ianthetechie:
This can be really simple to start using `console.log` or similar.

PR #653, opened 2025-07-14
Define wrapper of NavigationController for recording

Kandarp05:
#640 

Tried to `console.log` events after every `NavState` update and it works :) 

But this still need some cleanup


ianthetechie:
> Tried to console.log events after every NavState update and it works :)

Nice!! Looking forward to trying it out!

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -391,6 +398,65 @@ impl NavigationController {
...
+
+    fn update_user_location(&self, location: UserLocation, state: NavState) -> NavState {
+        let new_state = self
+            .controller
+            .update_user_location(location, state.clone());
+        let new_recording = self
+            .recording
+            .record_nav_state_update(state, new_state.clone());
+
+        NavState::add_recording(new_state, new_recording)
```
ianthetechie: `new_state.add_recording(...)` is the more typical pattern I think. But let's also discuss this on the call today since I think there's a few related things we might rework.

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -1,33 +1,42 @@
...
     pub initial_route: Route,
-    /// Initial trip state.
-    pub initial_trip_state: Option<TripState>,
-    /// Collection of events that occurred during the navigation session.
     pub events: Vec<NavigationRecordingEvent>,
 }
 
 /// Custom error type for navigation recording operations.
-#[derive(Debug)]
+/// Note: Due to UniFFI limitations, we cannot include the underlying error details.
```
ianthetechie: I think you're referring to the fact that it can't send an arbitrary error over the FFI (which is indeed correct). The pattern we use elsewhere is to include a string description of the error rather than a rich object. We cleaned this up on our review call this morning.

web/index.html
```diff
@@ -40,13 +40,15 @@
...
 
     <!--
-	This should not require any API key if run via `npm run dev`.
-	Learn how to get a free prod API key here: https://docs.stadiamaps.com/authentication/.
-	See https://stadiamaps.github.io/ferrostar/vendors.html for more info on basemap vendors.
-      -->
+This should not require any API key if run via `npm run dev`.
+Learn how to get a free prod API key here: https://docs.stadiamaps.com/authentication/.
+See https://stadiamaps.github.io/ferrostar/vendors.html for more info on basemap vendors.
+-->
```
ianthetechie, line 51: Weird formatting change...

web/src/ferrostar-map.ts
```diff
@@ -336,9 +358,26 @@ export class FerrostarMap extends LitElement {
...
     this._navState = newState;
     this.onTripStateChange?.(newState?.tripState || null);
+
+    // Stop navigation automatically when it is finished.
+    if (
+      newState?.tripState &&
+      Object.keys(newState.tripState)[0] == "Complete"
+    ) {
+      this.stopNavigation();
+    }
```
ianthetechie: I don't think this is something we want to do automatically. There are reasons a developer may want to keep navigation "active" (own exit flow etc.)

ianthetechie (commented):
There's a bit of a split brain problem developing. The codebase now has:

* A `RecordingNavigationController` which stores a `NavigationRecording`.
* The storage of the actual recording inside `NavState`.
* The `NavigationRecording` struct, which isn't actually a recording; it's more of a Java style utility class with methods that accept + return an unrelated type (the event vector).

This makes the usage difficult to follow.

Let's discuss the following today:

1. Where should we hold the events?
2. What is the role of the `NavigationRecording` struct? It's in this funny middle ground right now since most of its functionality doesn't really need to exist on a separate struct. There are (at least?) two options:
  a. Make it an actual _recording_ struct, which stores the events internally. I _think_ we may have shot this idea down though due to issues making the interior mutability work. That's probably why `NavState` currently holds the vector of events (which is fine, to be clear :D).
  b. Its functionality moves largely to `NavState` / `RecordingNavigationController`. In this case, it might not exist at all.
  c. It becomes a "builder" instead, which is initialized with the initial variables, but doesn't produce an actual recording until you give it an event vector. I think that's what it functionally is right now, but it just doesn't implement the pattern cleanly. I have some ideas for how to clean this up which we can discuss in the next hour on our call.

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -368,30 +392,50 @@ impl From<NavigationControllerConfig> for SerializableNavigationControllerConfig
...
+            trip_state: serializable_nav_state.trip_state,
+            step_advance_condition: serializable_nav_state.step_advance_condition,
+        })
+    }
+
+    #[cfg_attr(feature = "uniffi", uniffi::constructor)]
+    pub fn error(error_message: String) -> Self {
+        Self::new(NavigationRecordingEventData::Error { error_message })
+    }
 }
```
Kandarp05: Do we need to export this via UniFFI, since we are not exporting `append_recording_event`?

Exporting this was somehow causing this error while building FFI bindings
```
Task :core:generateDebugUniFFIBindings FAILED
Caused by:

    0: extracting metadata for 'UNIFFI_META_FERROSTAR_CONSTRUCTOR_NAVIGATIONRECORDINGEVENT_ERROR'
    1: Constructor return type must be Self or Arc<Self>

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':core:generateDebugUniFFIBindings'.
```

ianthetechie: Ahh heh ;) Yeah, that's a known limitation of UniFFI.

I don't think these need to be exported though so all good.

PR #654, opened 2025-07-15
Add CarPlay Stadia Search

bolsinga:
- `DemoCarPlaySearch` -> this can actually go into the api-search repository. It is designed not to have external dependencies.


apple/DemoApp/Demo/DemoCarPlaySearch.swift
```diff
@@ -0,0 +1,193 @@
...
+            pushError = error
+        }
+        if let pushError { handleError(pushError) }
+    }
+
+    func searchTemplate(_: CPSearchTemplate, updatedSearchText searchText: String) async -> [CPListItem] {
+        do {
+            if self.searchText != searchText {
+                self.searchText = searchText
+                try await search(query: searchText, autocomplete: false)
```
ianthetechie, line 98: This should only ever be `false` when the user indicates they are finished searching (i.e. pressing the search button on the on-screen keyboard).

It looks like it's `false` only to get the full place details with every search, but in doing so, the results will actually be quite a bit worse on partial input (all of the input on mobile :P).

Ideally we'd fetch the details on selection. If that's not possible, lemme know and I've got some other ideas.

bolsinga: > This should only ever be `false` when the user indicates they are finished searching (i.e. pressing the search button on the on-screen keyboard).
> 
> It looks like it's `false` only to get the full place details with every search, but in doing so, the results will actually be quite a bit worse on partial input (all of the input on mobile :P).
> 
> Ideally we'd fetch the details on selection. If that's not possible, lemme know and I've got some other ideas.

I don't have an `async` context to get the details once Search is pressed. So it cannot request the details then...

bolsinga: FYI, if it helps, the only detail missing is the lat/lng. The rest are superfluous for "I need a search destination".

ianthetechie: Sorry for the delay in responding to this, but I just had a deeper look and it seems like, while there isn't a "swifty" async context, you can indeed do async operations (in this case, calling the place details endpoint with the selected item ID).

> CarPlay executes your handler on the main queue. You must call the completion closure, or completionBlock in Objective-C, after you finish processing the selection. If you need to perform asynchronous tasks, dispatch them to a background queue and call the completion closure or completionBlock when they complete. CarPlay displays an asynchronous progress indicator until you call the completion closure.

PR #656, opened 2025-07-16
Initial proptest setup for step advance conditions

ianthetechie:


common/ferrostar/src/navigation_controller/step_advance/conditions.rs
```diff
@@ -462,15 +444,9 @@ mod tests {
...
-        let route_step =
-            crate::navigation_controller::test_helpers::gen_route_step_with_coords(vec![
-                (0.0, 0.0),   // Origin
-                (0.001, 0.0), // 111 meters east at the equator
-            ]);
-
+        // FIXME: This comment isn't even correct?
         // Create a location that's on the line (so deviation = 0)
-        let user_location = user_location(0.0, 0.0005, 5.0); // Point directly on the line
+        let user_location = make_user_location(coord!(x: 0.0, y: 0.0005), 5.0); // Point directly on the line
```
ianthetechie: Do you happen to know anything about this comment?

common/ferrostar/src/test_utils.rs
```diff
@@ -0,0 +1,34 @@
...
+        },
+        horizontal_accuracy,
+        course_over_ground: None,
+        timestamp: SystemTime::now(),
+        speed: None,
+    }
+}
+
+prop_compose! {
+    pub fn arb_coord()(x in -180f64..180f64, y in -90f64..90f64) -> Coord {
```
ianthetechie, line 25: If you're in the mood to do a bit of cleanup, you can search the codebase for other uses of `in -180f64..180f64` and replace them with arb_coord. Probably a few other patterns too that we could unify like I think there's still one more way of generating randomized route steps that we could clean up.

(Also, yes, this is a bit of a mind trick to parse; it's a macro with its own special syntax for declaring these sorts of pseudo-higher-order functions for proptest).

common/ferrostar/src/navigation_controller/step_advance/conditions.rs
Archdoog: We can delete

Archdoog (approved):

PR #661, opened 2025-07-24
new distance system & maximum fraction digits

tedi4t:


web/src/ferrostar-map.ts
```diff
@@ -77,6 +98,12 @@ export class FerrostarMap extends LitElement {
   @property({ type: Boolean })
   addGeolocateControl: boolean = true;
 
+  @property({ converter: distanceSystemConverter })
+  system?: DistanceSystem;
+
+  @property({ type: Number })
+  maximumFractionDigits?: number;
```
ianthetechie: Probably can do a bit better with naming + maybe add docstrings.

web/src/trip-progress-view.ts
```diff
@@ -13,6 +14,12 @@ export class TripProgressView extends LitElement {
   @property()
   tripState: any = null;
 
+  @property()
+  system: DistanceSystem = "metric";
+
+  @property()
+  maximumFractionDigits = 2;
```
ianthetechie: ditto

web/index.html
```diff
@@ -53,6 +53,8 @@
         valhallaEndpointUrl="https://api.stadiamaps.com/route/v1"
         styleUrl="https://tiles.stadiamaps.com/styles/outdoors.json"
         profile="bicycle"
+        system="imperial"
```
ianthetechie: It looks like the display at the bottom updates, but the navigation banners are still metric ;)
image

ianthetechie:
Sorry for the delay getting back to this @tedi4t! I was traveling for the past week or so and this got buried.

Another contributor had been working on further splitting the responsibilities between the core and the map widget in #6622, which created a few conflicts. I think I've resolved all of the conflicts now.

The only outstanding issue I have is that it seems to show decimal places for the "smaller" units. While it's quite reasonable to display something like 0.9 mi or 1.2km, I don't think anyone cares about fractional meters, feet, or yards ;) GPS isn't even that accurate anyways.

image

ianthetechie (commented):
The latest changes address some of the issues. For example, feet are no longer shown with a decimal value. However, there are a few problems outstanding.

1. There is a very long stretch where "1 mi" is displayed. For very long distances, yes, we should hide the precision. But typically between 1 and 10 miles (or kilometers), you'd want a slightly more granular display (1 - 2 digits?).

image

2. The thresholds do not seem to be aligned between when the library switches to feet, and when we do the cutoff here. For example, at this point in the route, there's a significant period of time where we display "0 mi" which isn't very helpful. These should be aligned. Perhaps some of this code belongs in a PR in the formatters library instead, so we can have a holistic view of which unit we're showing and make decisions based on that? (We should never show fractional feet, yards, or meters, but we should SOMETIMES show fractional miles and kilometers).

image

ianthetechie:
Hmm... I'm using the PR branch with no changes locally. I just click "Simulate navigation" and for the longest time it just shows "1 mi" (and even eventually "0 mi" for a short bit) with no further detail. Then it eventually switches to feet.

> These should be aligned. Perhaps some of this code belongs in a PR in the formatters library instead, so we can have a holistic view of which unit we're showing and make decisions based on that? (We should never show fractional feet, yards, or meters, but we should SOMETIMES show fractional miles and kilometers).

My point here is that right now there's code within the formatters library which determines the distance at which we switch between, say, miles and feet, or km and meters. This is problematic for a few reasons.

The user of the library code has no idea where the switch (e.g. miles -> feet) happens. Since all we pass is a maximum number of fraction digits, and we don't know the switch-over point, it's essentially impossible to guarantee it behaves as desired. Looking at the code over there, it looks sane enough for metric, but doesn't really make any sense for either imperial setting.

I think we can just hard code this in the formatters library since it's pretty narrow in purpose. Here's the reference logic from Android: https://github.com/stadiamaps/ferrostar/blob/0b981e6802e1a030d02f1ef1d8039a0337e2edae/android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/formatting/DistanceFormatter.kt#L74. It looks like you've ported _some_ of the logic, but not all, and having to set an explicit number of (max) fraction digits when calling it is causing the issues.

ianthetechie:
Yeah, I think that sounds like a good plan @tedi4t. Thanks!

igalgh:
@ianthetechie we'd like to finalize this PR.
From your POV - is it sufficient to resolve the conflicts in order to merge it?

ianthetechie:
@igalgh I would prefer to have this implemented directly in Rust at some point so that we don't have to keep juggling around separate formatting libraries, but it hasn't been a priority for me 😅 

I would accept this PR if after the merge conflicts are resolved, it passes my testing. I recall having strange formatting bugs along the way with previous iterations (noted in the old comments) but if we can resolve those I'd merge this.

eminmapme:
Hi @ianthetechie  — I merged stadiamaps/main and fixed the "0 mi" / "1 mi" rounding. I can’t push to this branch (eminmapme has no write access on tedi4t/ferrostar). Could you add me as a collaborator so I can update the PR?

ianthetechie:
Superseded by #958.

PR #662, opened 2025-07-27
Separate Navigation Logic from Map Rendering in Web

Kandarp05:
This PR aims to separate the core navigation logic from map in web.

This introduces `ferrostar-core` which handles the whole navigation logic, location management and route computation. While `ferrostar-map` focuses mainly on map rendering and UI.

Core emits `navstate-change` events which is consumed by the map.

`ferrostar-map` is setup such that you just need to give the route to start with, then just give it new navStates and it will render it automatically, this will allow us to reuse this in Navigation Replay later on.


web/index.html
```diff
@@ -122,43 +133,46 @@
...
             const location =
-              await ferrostar.locationProvider.getCurrentLocation(30_000);
+              await ferrostarCore.locationProvider.getCurrentLocation(30_000);
 
             // Use the acquired user location to request the route
-            const routes = await ferrostar.getRoutes(location, waypoints);
+            const routes = await ferrostarCore.getRoutes(location, waypoints);
             const route = routes[0];
+            ferrostarMap.route = route;
+            ferrostarMap.isSimulation = false;
```
ianthetechie: Why does the map have a simulation property; isn't that entirely in the core?

web/src/ferrostar-map.ts
```diff
@@ -310,89 +259,95 @@ export class FerrostarMap extends LitElement {
...
-  }
+  /**
+   * Renders or updates the user's location marker on the map. If a marker already exists, its position is updated.
+   * Otherwise, a new marker is created and added to the map.
+   */
+  private renderUserLocationMarker() {
+    if (
+      !this.map ||
+      !this.navState?.tripState ||
+      !("Navigating" in this.navState.tripState)
```
ianthetechie: It seems like there are use cases where the user's location should still be visible after arrival at the destination 🤔

web/index.html
```diff
@@ -106,8 +107,18 @@
         });
 
         // Get a reference to the Ferrostar map component
-        const ferrostar = document.getElementById("ferrostar");
-        ferrostar.map = mapInstance;
+        const ferrostarMap = document.getElementById("ferrostar");
+        const ferrostarCore = document.getElementById("ferrostar-core");
+        ferrostarMap.map = mapInstance;
+        ferrostarMap.showNavigationUI = true;
```
ianthetechie: Should this be internally managed? Seems a bit odd to make the user control this themselves.

web/index.html
```diff
@@ -106,8 +107,18 @@
...
+
+        ferrostarMap.onStopNavigation = () => {
+          ferrostarCore.stopNavigation();
+        };
```
ianthetechie: This can probably be simplified like so. Overall though it seems like the user has to do a lot of work to hook these two up. Maybe we need a convenience method like `ferrostarMap.linkWith(ferrostarCore)` or vice versa?

```suggestion
        ferrostarMap.onStopNavigation = ferrostarCore.stopNavigation;
```

web/src/ferrostar-map.ts
```diff
@@ -1,63 +1,49 @@
...
    */
-  @property({ type: Object, attribute: false })
-  customStyles?: object | null;
+  @property({ type: Boolean })
+  showNavigationUI: boolean = false;
 
   /**
-   * Enables voice guidance via the web speech synthesis API.
-   * Defaults to false.
+   * Determines whether the navigation is a simulation
```
ianthetechie: Let's just make this comment a bit clearer.

web/index.html
ianthetechie: Let's also add a comment here letting the reader know this setup is for the demo purposes and that this enables the simulated marker display.

web/src/ferrostar-map.ts
ianthetechie: I see now... this is also just for simulation :) We can probably make this clearer with a comment and/or better name.

Kandarp05:
What why did that fail?

guide/src/web-customization.md
```diff
@@ -48,7 +48,7 @@ const waypoints = [{ coordinate: { lat: parseFloat(lat), lng: parseFloat(lon) },
 
 The asynchronous `getRoutes` method on `FerrostarCore`
 will fetch routes from your route provider (ex: a Valhalla server).
-Here’s an example:
+Here’s an example:1
```
ianthetechie:
```suggestion
Here’s an example:
```

web/index.html
```diff
@@ -50,18 +50,19 @@
...
-      <ferrostar-map
-        id="ferrostar"
+      <ferrostar-map id="ferrostar"></ferrostar-map>
+      <ferrostar-core
+        id="ferrostar-core"
         valhallaEndpointUrl="https://api.stadiamaps.com/route/v1"
-        styleUrl="https://tiles.stadiamaps.com/styles/outdoors.json"
         profile="bicycle"
-      ></ferrostar-map>
+      ></ferrostar-core>
```
ianthetechie: It *might* actually make sense to move this to JS now that it's completely abstract and not related to the DOM in any way.

web/src/ferrostar-core.ts
```diff
@@ -0,0 +1,222 @@
...
+          new SpeechSynthesisUtterance(
+            tripState.Navigating?.spokenInstruction.text,
+          ),
+        );
+      }
+    }
+  }
+
+  // This component doesn't render anything - it's purely for logic
+  render() {
```
ianthetechie: Discussed on the call: let's try using `ReactiveElement` instead so we don't need a render method.

ianthetechie (commented):
Just a few minor things :)

ianthetechie (approved):
Looks good now; thanks!

PR #675, opened 2025-08-19
Improve lint

haysmike:
Pulling in a couple of my favorite ESLint setup bits:

- Ignore unused arguments and rest siblings in `no-unused-vars` rule
- Pass the `--fix` flag to the existing `lint` command instead of duplicating the whole thing
- Add the `--max-warnings` flag to the lint command so we don't add warnings without being intentional about it (ideally this should be 0)


ianthetechie (approved):
Thanks, Mike!

PR #677, opened 2025-08-19
Write documentation for recording module

Kandarp05:
Started writing rust docs for recording/replay

### DIFF
--- a/common/ferrostar/src/navigation_controller/mod.rs
+++ b/common/ferrostar/src/navigation_controller/mod.rs
@@ -1,4 +1,7 @@
-//! The navigation state machine.
+//! # Documentation for Navigation Module
+//!
+//! The navigation module provides tools for managing route navigation and tracking user progress through a series of steps.
+//! Below is the documentation of the code provided, including its core functionality, traits, and implementations.
 
 pub mod models;
 pub mod recording;
@@ -43,6 +46,8 @@ pub trait Navigator: Send + Sync {
     fn get_initial_state(&self, location: UserLocation) -> NavState;
     fn advance_to_next_step(&self, state: NavState) -> NavState;
     fn update_user_location(&self, location: UserLocation, state: NavState) -> NavState;
+    /// Attempts to retrieve a recording based on the supplied navigation events.
+    /// Only controllers with recording capabilities should override this implementation.
     fn get_recording(
         &self,
         #[allow(unused_variables)] events: Vec<NavigationRecordingEvent>,
@@ -455,6 +460,8 @@ impl Navigator for RecordingNavigationController {
 }
 
 /// JavaScript wrapper for `NavigationController`.
+/// This wrapper is required because `NavigationController` cannot be directly converted to a JavaScript object
+/// and requires serialization/deserialization of its methods' inputs and outputs.
 #[cfg(feature = "wasm-bindgen")]
 #[wasm_bindgen(js_name = NavigationController)]
 pub struct JsNavigationController(Arc<dyn Navigator>);
--- a/common/ferrostar/src/navigation_controller/models.rs
+++ b/common/ferrostar/src/navigation_controller/models.rs
@@ -392,6 +392,7 @@ impl From<NavigationControllerConfig> for SerializableNavigationControllerConfig
     }
 }
 
+/// Represents a navigation recording event which contains a timestamp and associated event data.
 #[derive(Clone, Serialize, Deserialize, Debug)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 #[cfg_attr(feature = "wasm-bindgen", derive(Tsify))]
@@ -421,6 +422,9 @@ impl NavigationRecordingEvent {
     }
 }
 
+/// The `NavigationRecordingEventData` enum represents different types of events
+/// that occur during a navigation recording process. It holds data related
+/// to state updates, route updates, or errors encountered.
 #[derive(Clone, Serialize, Deserialize, Debug)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
 #[cfg_attr(feature = "wasm-bindgen", derive(Tsify))]
--- a/common/ferrostar/src/navigation_controller/recording.rs
+++ b/common/ferrostar/src/navigation_controller/recording.rs
@@ -1,3 +1,5 @@
+//! Module for recording and replaying navigation sessions.
+
 use crate::models::Route;
 use crate::navigation_controller::models::{
     NavigationControllerConfig, NavigationRecordingEvent, SerializableNavigationControllerConfig,
@@ -8,16 +10,10 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "wasm-bindgen")]
 use wasm_bindgen::prelude::*;
 
-/// Represents a recorded navigation session with its configuration and events.
-///
-/// # Fields
-///
-/// * `version` - The version of Ferrostar that created this recording
-/// * `initial_timestamp` - When the navigation session started (in milliseconds)
-/// * `config` - Configuration settings used for the navigation session
-/// * `initial_route` - The route that was initially assigned for navigation
-/// * `events` - A chronological list of all navigation events that occurred
-#[derive(Clone, Serialize, Deserialize)]
+/// A builder struct used to construct a recording of navigation events.
+/// This struct encapsulates necessary data for initializing and configuring
+/// a navigation recording process.
+#[derive(Serialize, Deserialize, Clone)]
 pub struct NavigationRecordingBuilder {
     pub version: String,
     pub initial_timestamp: i64,
@@ -25,7 +21,6 @@ pub struct NavigationRecordingBuilder {
     pub initial_route: Route,
 }
 
-/// Functionality for the navigation controller that is not exported.
 impl NavigationRecordingBuilder {
     /// Creates a new navigation recorder with route configuration and initial state.
     pub fn new(config: NavigationControllerConfig, initial_route: Route) -> Self {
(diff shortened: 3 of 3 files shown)


common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -88,6 +90,7 @@ pub enum RecordingError {
     RecordingNotEnabled,
 }
 
+/// A struct that wraps around `NavigationRecording` to provide replay functionality.
```
ianthetechie:
```suggestion
/// A wrapper around `NavigationRecording` to facilitate replaying the event stream.
```

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -71,7 +64,16 @@ impl NavigationRecording {
...
+/// Enum `RecordingError` represents the possible errors that can occur during
+/// the recording process in a navigation recording system.
+///
+/// # Variants
+///
+/// - `SerializationError`
+///   Indicates an error occurred during the serialization of a recording.
+///
+/// - `RecordingNotEnabled`
+///   Indicates that recording is not enabled for a specific controller.
```
ianthetechie: You can actually document these directly on the enum variant! GitHub doesn't let me suggest changes there, but you just add a `/// ` comment over each variant and it'll work. The enum comment can get a lot shorter.

```suggestion
/// A session recording error.
```

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -62,6 +54,7 @@ pub struct NavigationRecording {
 }
 
 impl NavigationRecording {
+    /// Converts a JSON string slice into an instance of the type implementing this method.
```
ianthetechie:
```suggestion
    /// Deserializes a previously saved navigation recording from a JSON string.
```

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -54,6 +44,8 @@ impl NavigationRecordingBuilder {
     }
 }
 
+/// Represents a navigation recording, which consists of a builder containing metadata
```
ianthetechie: A docstring doesn't normally need to describe implementation details unless the user really needs to be aware of them. In this case they don't (in fact, those details are private!).

```suggestion
/// A navigation session recording.
///
/// Internally this contains the full event stream.
/// TODO: Hints for how you would typically use / interact with this? You can link to other types (and functions) by the way like [`NavigationReplay`] :)
```

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -8,24 +10,17 @@ use serde::{Deserialize, Serialize};
...
-///
-/// * `version` - The version of Ferrostar that created this recording
-/// * `initial_timestamp` - When the navigation session started (in milliseconds)
-/// * `config` - Configuration settings used for the navigation session
-/// * `initial_route` - The route that was initially assigned for navigation
-/// * `events` - A chronological list of all navigation events that occurred
-#[derive(Clone, Serialize, Deserialize)]
+/// A builder struct used to construct a recording of navigation events.
+/// This struct encapsulates necessary data for initializing and configuring
+/// a navigation recording process.
```
ianthetechie:
```suggestion
/// A builder for serializing a navigation recording.
```

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -8,24 +10,17 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Serialize, Deserialize, Clone)]
 pub struct NavigationRecordingBuilder {
```
ianthetechie: I wonder if this type deserves a different name. It doesn't look anything like a normal builder pattern.

- All the fields are `pub`
- It doesn't have a `build` method
- It isn't ever "consumed" into another type... it just returns a string for you to write somewhere...

I'm not saying the API is wrong ;) Just that it doesn't look like a builder, so maybe we should call it something else if this is the API we want.

```suggestion
pub struct NavigationRecordingSerializer {
```

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -8,24 +10,17 @@ use serde::{Deserialize, Serialize};
...
     pub version: String,
     pub initial_timestamp: i64,
     pub config: SerializableNavigationControllerConfig,
     pub initial_route: Route,
```
ianthetechie: Docstrings here?  Also I'm not sure if they need to all be pub.

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -392,6 +392,7 @@ impl From<NavigationControllerConfig> for SerializableNavigationControllerConfig
     }
 }
 
+/// Represents a navigation recording event which contains a timestamp and associated event data.
```
ianthetechie: I wonder if this struct should go in the `recording` module instead?

```suggestion
/// An event that occurs during navigation.
///
/// This is used for the optional session recording / telemetry.
```

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -421,6 +422,9 @@ impl NavigationRecordingEvent {
     }
 }
 
+/// The `NavigationRecordingEventData` enum represents different types of events
+/// that occur during a navigation recording process. It holds data related
+/// to state updates, route updates, or errors encountered.
```
ianthetechie:
```suggestion
/// The event type.
///
/// For full replayability, we record things like rerouting, and not just location updates.
```

common/ferrostar/src/navigation_controller/models.rs
```diff
@@ -421,6 +422,9 @@ impl NavigationRecordingEvent {
...
 #[derive(Clone, Serialize, Deserialize, Debug)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
 #[cfg_attr(feature = "wasm-bindgen", derive(Tsify))]
```
ianthetechie, line 430: GitHub is stupid and won't let me comment a few lines down, but the `Error` variant here is strange enough that it needs more documentation of what it's recording, and why.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -43,6 +43,8 @@ pub trait Navigator: Send + Sync {
     fn get_initial_state(&self, location: UserLocation) -> NavState;
     fn advance_to_next_step(&self, state: NavState) -> NavState;
     fn update_user_location(&self, location: UserLocation, state: NavState) -> NavState;
+    /// Attempts to retrieve a recording based on the supplied navigation events.
+    /// Only controllers with recording capabilities should override this implementation.
```
ianthetechie:
```suggestion
    /// Attempts to retrieve a recording based on the supplied navigation events.
    ///
    /// The default implementation returns an error indicating that recording is not enabled.
    /// Navigation controllers which support recording can provide their own implementation.
```

common/ferrostar/src/navigation_controller/models.rs
Kandarp05: Yup I think these can now be moved to the recording module.

common/ferrostar/src/navigation_controller/models.rs
Kandarp05: I mean to be honest, we don't even use this enum, I defined it before thinking we are gonna need this, but never found a good use for it. Same is for the `RotueUpdate`. 

For now I am just adding a TODO here.

common/ferrostar/src/navigation_controller/recording.rs
Kandarp05: `NavigationRecordingSerializer` doesn't really sound correct either, what about `NavigationRecordingMetadata`??

common/ferrostar/src/navigation_controller/models.rs
Kandarp05: Or maybe, it will be nice to refactor the whole recording/replay, as right now everything is kind of spread everywhere (maybe making a recording module, instead of a sub-module??)

common/ferrostar/src/navigation_controller/recording.rs
ianthetechie: Works for me; thanks!

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -10,18 +10,20 @@ use serde::{Deserialize, Serialize};
...
-/// a navigation recording process.
+/// A builder for serializing a navigation recording.
 #[derive(Serialize, Deserialize, Clone)]
-pub struct NavigationRecordingBuilder {
-    pub version: String,
-    pub initial_timestamp: i64,
-    pub config: SerializableNavigationControllerConfig,
-    pub initial_route: Route,
+pub struct NavigationRecordingMetadata {
+    /// Version of ferrostar used.
```
ianthetechie:
```suggestion
    /// Version of Ferrostar used.
```

common/ferrostar/src/navigation_controller/models.rs
ianthetechie: Yeah, a bit of cleanup is probably in order now that it's built. Happens every time :)

common/ferrostar/src/navigation_controller/models.rs
ianthetechie: Okay. Let's discuss today on the call.

ianthetechie: Discussed; we can remove it.

ianthetechie (approved):

PR #679, opened 2025-08-25
Add controls for replay in web.

Kandarp05:
#641 
This PR adds controls like play/pause, timer, and slider for replays in web.

### DIFF
--- a/common/ferrostar/src/navigation_controller/recording.rs
+++ b/common/ferrostar/src/navigation_controller/recording.rs
@@ -96,11 +96,24 @@ impl NavigationReplay {
         Self(NavigationRecording::from_json(json))
     }
 
-    pub fn get_next_event(&self, current_index: u64) -> Option<NavigationRecordingEvent> {
+    pub fn get_event_by_index(&self, current_index: u64) -> Option<NavigationRecordingEvent> {
         match self.0.events.get(current_index as usize) {
             Some(event) => Some(event.clone()),
             None => None,
         }
+    }
+
+    pub fn get_all_events(&self) -> Vec<NavigationRecordingEvent> {
+        self.0.events.clone()
+    }
+
+    pub fn get_total_duration(&self) -> i64 {
+        if self.0.events.is_empty() {
+            return 0;
+        }
+        let first_event = self.0.events.first().unwrap();
+        let last_event = self.0.events.last().unwrap();
+        last_event.timestamp() - first_event.timestamp()
     }
 
     pub fn get_initial_timestamp(&self) -> i64 {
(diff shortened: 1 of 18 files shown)


common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -96,13 +96,26 @@ impl NavigationReplay {
         Self(NavigationRecording::from_json(json))
     }
 
-    pub fn get_next_event(&self, current_index: u64) -> Option<NavigationRecordingEvent> {
+    pub fn get_event_by_index(&self, current_index: u64) -> Option<NavigationRecordingEvent> {
```
ianthetechie: That is a much better method name 🎉

common/ferrostar/src/navigation_controller/recording.rs
```diff
@@ -96,13 +96,26 @@ impl NavigationReplay {
...
         match self.0.events.get(current_index as usize) {
             Some(event) => Some(event.clone()),
             None => None,
         }
     }
 
+    pub fn get_all_events(&self) -> Vec<NavigationRecordingEvent> {
```
ianthetechie: I'm not sure where this is used, but it's somewhat surprising to return a cloned value from a getter like this when it'd result in a deep copy. I think I'd normally have this return a borrowed slice (`&[NavigationRecordingEvent]`), but I guess it depends how you're using it.

web/eslint.config.js
```diff
@@ -35,7 +35,7 @@ export default [
       ...wcPlugin.configs.recommended.rules,
       "wc/no-constructor-params": "off",
       "lit/no-template-arrow": "off",
-      "@typescript-eslint/no-explicit-any": "warn",
+      "@typescript-eslint/no-explicit-any": "off",
```
ianthetechie: Hmmm..... I'm not quie sure why this is off now 🤔

web/index.html
```diff
@@ -246,9 +410,15 @@
...
+            // Reset the controls
+            controls.style.display = "none";
+            if (currentProgressInterval) {
+              clearInterval(currentProgressInterval);
+              currentProgressInterval = null;
+            }
           };
 
-          await replay.startReplay();
+          replay.play();
```
ianthetechie: Investigate

web/src/ferrostar-map.ts
```diff
@@ -373,7 +373,7 @@ export class FerrostarMap extends LitElement {
    * Handles the logic for stopping the navigation process. It clears the navigation state,
    * resets the simulation flag, and invokes the optional onStopNavigation callback if provided.
    */
-  private handleStopNavigation() {
+  handleStopNavigation() {
```
ianthetechie: Why did the visibility change?

ianthetechie (approved):
Great work; thanks!

PR #680, opened 2025-08-29
Feat/apple/enhanced car play view

Archdoog:


Package.swift
```diff
@@ -31,7 +31,7 @@ if useLocalMapLibreSwiftUIDSL {
 } else {
     maplibreSwiftUIDSLPackage = .package(
         url: "https://github.com/maplibre/swiftui-dsl",
-        from: "0.12.0"
+        branch: "Rallista:feat/some-stylelayercollection"
```
Archdoog: Flagged for revert & upgrade: just using this to test out a real build.

Issue #682, opened 2025-09-02
Android - `TripState.preferredUserLocation()` returns wrong location value


iboalali:
I don't know if it affects iOS too but in the Kotlin code there might be a bug.
See this extension function https://github.com/stadiamaps/ferrostar/blob/5ce10904a48145a8e8d92852bda75a19736fb18d/android/core/src/main/java/com/stadiamaps/ferrostar/core/extensions/TripStateExtensions.kt#L102-L121

The code does not match the docs and I think the docs are correct. The snapped location should be returned when there is **no deviation** and the users raw location when there is.

The function should be
```
fun TripState.preferredUserLocation() =
    when (this) {
      is TripState.Navigating -> {
        when (this.deviation) {
          is RouteDeviation.NoDeviation -> this.snappedUserLocation
          is RouteDeviation.OffRoute -> this.userLocation
        }
      }
      is TripState.Idle -> this.userLocation
      is TripState.Complete -> this.userLocation
    }
```

Issue #414, opened 2025-01-07
Android (and maybe iOS) - When user is not on route, we should present the un-snapped location.


Archdoog:
Currently, if you go off course without a re-route handler (e.g. no cellular or it's disabled). The location puck will be stuck at the closest part of the route line until a re-route succeeds.

I've only tested this extensively on android, so I'm not sure if it's a core issue that affects iOS & android or just android.

ben-burwood:
Hi!  We're having this same problem where a user can be way off route but still show snapped to the Route. 

Would it not be possible to just update `snapUserLocationToRoute` reactively, monitoring the `core.state.tripState.deviation()` , where we snap to the Route if on-route, else unsnapped (this could be either any deviastion or >x distance from the Route). Not sure what else you had in mind for how this would function?

ianthetechie:
@Archdoog to verify but I think this is fixed.

PR #683, opened 2025-09-03
fix: returns raw user location when there is no deviation

iboalali:
The snapped location should be returned when there is no deviation and the users raw location when there is. This fixes https://github.com/stadiamaps/ferrostar/issues/682 and maybe even https://github.com/stadiamaps/ferrostar/issues/414


ianthetechie (approved):
Oh my 😅 That's kinda funny in retrospect.

Issue #685, opened 2025-09-08
Starting navigation in the middle of the route tries to snap to the first Step and is immediately off route


iboalali:
I work on a custom navigation solution with very specific needs. It is a pre-calculated route which can be started without internet connection. Sometimes the driver is not at the beginning of the route, but somewhere on it.

But this navigation engine always assumes the navigation session always starts on the first _Step_, but we need it to start anywhere on the route. _MapboxNavigation_ (even the old one) can handle it.

I think this is the part that is causing this behavior. It gets the first _Step_ and then snaps to it:
https://github.com/stadiamaps/ferrostar/blob/abb3bde981634335726d5ffb3aac453576cdcda1/common/ferrostar/src/navigation_controller/mod.rs#L110-L118

If it is modified to check which _Step_ is the closes and than snap to it, it should handle both cases: starting at the beginning or in the middle.

Sadly, I don't know Rust to modify it

PR #692, opened 2025-09-12
feat: allow starting navigation at any point in the route

iboalali:
This is for the issue https://github.com/stadiamaps/ferrostar/issues/685.
It is a draft because it is just for demonstration purposes and discussion. And honestly I tried to implement it but failed 😅


Archdoog:
Was going to post this here. But it was better served in the slack thread: https://osmus.slack.com/archives/C0553U433FX/p1757887198821429?thread_ts=1757448967.522969&cid=C0553U433FX

PR #693, opened 2025-09-14
fix: correct clone bug

Archdoog:
- Removes clone from `AndCondition` and `OrCondition` that was causing ARC ref state weirdness across navigation states.

### DIFF
--- a/common/ferrostar/src/navigation_controller/step_advance/mod.rs
+++ b/common/ferrostar/src/navigation_controller/step_advance/mod.rs
@@ -23,16 +23,47 @@ pub struct StepAdvanceResult {
     /// The next iteration of the step advance condition.
     ///
     /// This is what the navigation controller passes to the next instance of [`NavState`](super::NavState) on the completion of
-    /// an update (e.g. a user location update). Usually, this value is one of the following:
+    /// an update (e.g. a user location update). Use the helper methods on [`StepAdvanceResult`] to ensure proper state handling:
     ///
-    /// 1. When `should_advance` is true, this should typically be a clean/new instance of the condition.
-    /// 2. When the condition is not advancing, but the condition maintains no state, this should be a
-    ///    clean/new instance of the condition.
-    /// 3. When the condition is not advancing and maintains state, this should be a new
-    ///    instance including the current state of the condition. See [`DistanceEntryAndExitCondition`]
+    /// - [`StepAdvanceResult::next()`] - Unified method that takes a boolean and handles state appropriately
+    /// - [`StepAdvanceResult::advance()`] - Creates an advancing result with automatically reset state
+    /// - [`StepAdvanceResult::no_advance()`] - Creates a non-advancing result preserving current state
     ///
-    /// IMPORTANT! If the condition advances. This **must** be the clean/default state.
+    /// The trait method [`StepAdvanceCondition::new_instance()`] ensures that composite conditions
+    /// (Or/And) properly reset all nested conditions when any condition triggers advancement.
+    ///
+    /// **CRITICAL**: When advancing, this must be a clean/reset state to prevent state leakage between steps.
     pub next_iteration: Arc<dyn StepAdvanceCondition>,
+}
+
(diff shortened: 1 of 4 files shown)


common/ferrostar/src/navigation_controller/step_advance/mod.rs
```diff
@@ -23,18 +23,49 @@ pub struct StepAdvanceResult {
...
+    /// Uses the provided next_iteration as-is for preserving stateful progress.
+    pub fn no_advance(next_iteration: Arc<dyn StepAdvanceCondition>) -> Self {
+        Self {
+            should_advance: false,
+            next_iteration,
+        }
+    }
+
+    /// Creates a step advance result that advances to the next step.
+    /// Automatically creates a fresh instance of the condition to ensure proper state isolation.
```
ianthetechie: 👍

common/ferrostar/src/navigation_controller/step_advance/mod.rs
```diff
@@ -23,18 +23,49 @@ pub struct StepAdvanceResult {
...
+    pub fn advance(condition: &dyn StepAdvanceCondition) -> Self {
+        Self {
+            should_advance: true,
+            next_iteration: condition.new_instance(),
+        }
+    }
+
+    /// Creates a step advance result based on the should_advance boolean.
+    /// If advancing, creates a fresh instance. If not advancing, uses the provided next_iteration.
+    pub fn next(should_advance: bool, condition: &dyn StepAdvanceCondition) -> Self {
```
ianthetechie: Can we make the API simpler here? It sems like, based on the descriptions, this is completely covered by `advance` and `no_advance`, and it's just a more confusing interface.

It also doesn't do what the docstring says; it creates the new condition instance only if we stay put (`no_advance`), which seems strange.

common/ferrostar/src/navigation_controller/step_advance/mod.rs
```diff
@@ -23,18 +23,49 @@ pub struct StepAdvanceResult {
     /// The next iteration of the step advance condition.
     ///
     /// This is what the navigation controller passes to the next instance of [`NavState`](super::NavState) on the completion of
-    /// an update (e.g. a user location update). Usually, this value is one of the following:
+    /// an update (e.g. a user location update). Use the helper methods on [`StepAdvanceResult`] to ensure proper state handling:
```
ianthetechie, line 26: Can we make `should_advance` private prevent accidental footguns?

common/ferrostar/src/navigation_controller/step_advance/mod.rs
Archdoog: Cool. Will remove.

common/ferrostar/src/navigation_controller/step_advance/mod.rs
```diff
@@ -19,22 +19,52 @@ pub mod conditions;
...
+    ///
+    /// This is used by the navigation controller to determine whether to process the step advance.
+    pub fn should_advance(&self) -> bool {
+        self.should_advance
+    }
+
+    /// Creates a step advance result that does not advance to the next step.
+    /// Uses the provided next_iteration as-is for preserving stateful progress.
+    ///
+    /// Note: it's up to the caller to determine whether next_interation should reset.
```
Archdoog:
```suggestion
    /// Note: it's up to the caller to determine whether next_iteration should reset.
```

Archdoog (commented):
Spelling fix

common/ferrostar/src/navigation_controller/step_advance/mod.rs
Archdoog: Fixed 👍

common/ferrostar/src/navigation_controller/step_advance/mod.rs
Archdoog: Removed 👍. I also renamed the functions for a bit easier understanding.

PR #695, opened 2025-09-16
fix: hide road label on iOS when off route

Archdoog:
- Hides the current road name view when the user is not on route by requiring `.noDeviation`.


ianthetechie (approved):
Ha!

This also incidentally seems like it could theoretically support "free drive" mode whenever someone cares about that.

PR #712, opened 2025-10-06
feat: fix deviation before arrive on demo app

Archdoog:
- Fixes the demo app to prevent a forced offRoute at arrival. (Step before arrival advances at 30 meters, but deviation was 20 meters).


apple/DemoApp/Demo/DemoModel.swift
```diff
@@ -34,7 +34,7 @@ private extension FerrostarCore {
                 minimumHorizontalAccuracy: 32
             ),
             arrivalStepAdvanceCondition: stepAdvanceDistanceToEndOfStep(
-                distance: 30,
+                distance: 10,
```
ianthetechie, line 37: Double check Android too ;)

Archdoog: Good call, synced Android and iOS up and gave both a simulated run.

PR #714, opened 2025-10-08
feat: added deviation to step advance for short circuit tooling

Archdoog:
- Adds `RouteDeviation` to step advance condition check, enabling short circuit type step advance usage to require on route. This allows fine tuning a short circuit step advance to only succeed when the user re-joins the route.
- `RouteDeviationDetection::StaticThreshold` now first checks the current step, then falls back to the full route. This allows us to fall back to the full route for the updated `StepAdvanceConditions`.

### DIFF
--- a/android/core/src/androidTest/java/com/stadiamaps/ferrostar/core/FerrostarCoreTest.kt
+++ b/android/core/src/androidTest/java/com/stadiamaps/ferrostar/core/FerrostarCoreTest.kt
@@ -427,7 +427,7 @@ class FerrostarCoreTest {
         routes.first(),
         NavigationControllerConfig(
             WaypointAdvanceMode.WaypointWithinRange(100.0),
-            stepAdvanceDistanceFromStep(16u, 32u),
+            stepAdvanceDistanceFromStep(16u, 32u, true),
             stepAdvanceDistanceToEndOfStep(16u, 32u),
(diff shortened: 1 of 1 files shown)


Archdoog:
> @Archdoog, @ianthetechie,
> 
> > RouteDeviationDetection::StaticThreshold now first checks the current step, then falls back to the full route. This allows us to fall back to the full route for the updated StepAdvanceConditions.
> 
> Shouldn't it fall back to the remaining steps instead of the full route? Assuming that Ferrostart only skips steps forward.
> 
> Checking against the full route allows the user to go back to a past step in the route. This changes the state to on-route, and step skipping can then race to the end.
> 
> If we instead checked the off-route status only against the current or remaining steps, going back to past steps wouldn't trigger the on-route state and unintended step advancement (this is the issue I'm currently dealing with when testing the `main` branch).

That's a good catch. You're correct this behavior should only consider remaining steps to better restrict cases & configurations where racing to the end is possible. While this merged behavior is definitely an improvement over the last iteration where a user would just get stuck on the route, I think we should create a bug/improvement ticket to capture this.

https://github.com/stadiamaps/ferrostar/issues/720

I'll take a pass this week trying to write up how extensive down this road it makes sense to go. Since most of the systems in ferrostar tend to keep needed a bit richer understand of the state.

PR #715, opened 2025-10-09
"Speed run" step advance until we can't advance anymore

ianthetechie:
This is a work in progress due to the following outstanding issues:

- [x] We probably need special handling for the final step, since right now it basically skips over that (this causes ~half a dozen snapshot failures)
- [x] Documentation around manual step advance

Part of the solution to #685

### DIFF
--- a/common/ferrostar/src/navigation_controller/mod.rs
+++ b/common/ferrostar/src/navigation_controller/mod.rs
@@ -285,7 +285,7 @@ impl Navigator for NavigationController {
                     let updated_state = self.advance_to_next_step(intermediate_nav_state);
 
                     if is_arriving {
-                        return updated_state
+                        return updated_state;
                     } else {
                         // Recurse ("speed run" behavior)
                         return self.update_user_location(location, updated_state);


common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -257,7 +257,12 @@ impl Navigator for NavigationController {
 
                 if should_advance {
                     // Advance to the next step
-                    return self.advance_to_next_step(intermediate_nav_state);
+                    let updated_state = self.advance_to_next_step(intermediate_nav_state);
+                    // TODO: Special handling for the final step?
+                    // TODO: Manual step advance may not expect this "speed running" behavior!
```
Archdoog: Since Manual Step Advance always returns false for `should_advance`. We should be good.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -257,7 +257,12 @@ impl Navigator for NavigationController {
...
+                    // TODO: Ensure that we are careful about calling this when the user is off the route. Probably obsoleted by Jacob's PR
```
Archdoog: This is definitely sorted by the route deviation behavior inside the step advance condition. Using `calculateWhileOffRoute: false` with the `stepAdvanceDistanceFromStep` where you'd create a short circuit is working well with some considerations that we'll put in the guide.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -257,7 +257,12 @@ impl Navigator for NavigationController {
 
                 if should_advance {
                     // Advance to the next step
-                    return self.advance_to_next_step(intermediate_nav_state);
+                    let updated_state = self.advance_to_next_step(intermediate_nav_state);
+                    // TODO: Special handling for the final step?
```
Archdoog: Yeah, we probably just want to exit early or simply move this `remaining_steps.len() <= 2` into a param for the last step condition. Extending this to also block the looping exit: 

```rs                
                // Get the step advance condition result.
                let next_step = remaining_steps.get(1).cloned();
                let step_advance_result = if remaining_steps.len() <= 2 {
                    self.config
                        .arrival_step_advance_condition
                        .should_advance_step(
                            location,
                            current_step.clone(),
                            next_step,
                            deviation.clone(),
                        )
                } else {
                    state.step_advance_condition().should_advance_step(
                        location,
                        current_step.clone(),
                        next_step,
                        deviation.clone(),
                    )
                };
```

This code is just above the new looping behavior, so should be pretty easy to involve it.

PR #716, opened 2025-10-12
Feat/common/waypoint on step

Archdoog:
- Detaches WaypointAdvance logic from NavigationController
- adds WaypointAlongAdvancingStep mode for use with skip forward-like step advance modes.
- **Bugfix**: Changes waypoint behavior to always keep last waypoint, to ensure route re-calculation can occur until the trip has actually been completed.

### DIFF
--- a/common/ferrostar/src/navigation_controller/waypoint_advance.rs
+++ b/common/ferrostar/src/navigation_controller/waypoint_advance.rs
@@ -41,61 +41,59 @@ impl WaypointAdvanceChecker {
             } => {
                 match self.mode {
                     WaypointAdvanceMode::WaypointWithinRange(range) => {
-                        if event == WaypointCheckEvent::LocationUpdated {
-                            // Only advance waypoints if there are more than 1 remaining
-                            // (never remove the final destination waypoint)
-                            if remaining_waypoints.len() <= 1 {
-                                return WaypointAdvanceResult::Unchanged;
-                            }
-
-                            remaining_waypoints.first().map_or(
-                                WaypointAdvanceResult::Unchanged,
-                                |waypoint| {
-                                    let current_location: Point = user_location.coordinates.into();
-                                    let next_waypoint: Point = waypoint.coordinate.into();
-                                    let distance =
-                                        Haversine.distance(current_location, next_waypoint);
-                                    if distance < range {
-                                        // Slice the remaining waypoints starting from the second element
-                                        WaypointAdvanceResult::Changed(
-                                            remaining_waypoints[1..].to_vec(),
-                                        )
-                                    } else {
-                                        WaypointAdvanceResult::Unchanged
-                                    }
-                                },
-                            )
-                        } else {
-                            WaypointAdvanceResult::Unchanged
+                        if event != WaypointCheckEvent::LocationUpdated {
+                            return WaypointAdvanceResult::Unchanged;
                         }
+
+                        // Only advance waypoints if there are more than 1 remaining
+                        // (never remove the final destination waypoint)
+                        if remaining_waypoints.len() <= 1 {
+                            return WaypointAdvanceResult::Unchanged;
+                        }
+
+                        remaining_waypoints.first().map_or(
+                            WaypointAdvanceResult::Unchanged,
+                            |waypoint| {
+                                let current_location: Point = user_location.coordinates.into();
+                                let next_waypoint: Point = waypoint.coordinate.into();
+                                let distance = Haversine.distance(current_location, next_waypoint);
+                                if distance < range {
+                                    // Slice the remaining waypoints starting from the second element
+                                    WaypointAdvanceResult::Changed(
+                                        remaining_waypoints[1..].to_vec(),
+                                    )
+                                } else {
+                                    WaypointAdvanceResult::Unchanged
+                                }
+                            },
+                        )
                     }
                     WaypointAdvanceMode::WaypointAlongAdvancingStep(range) => {
-                        if let WaypointCheckEvent::StepAdvanced(current_step) = event {
-                            let step_linestring = current_step.get_linestring();
-                            let mut filtered_waypoints: Vec<Waypoint> = remaining_waypoints
-                                .iter()
-                                .filter(|waypoint| {
-                                    let waypoint_point: Point = waypoint.coordinate.into();
-                                    !self.is_waypoint_within_range_of_linestring(
-                                        &waypoint_point,
-                                        &step_linestring,
-                                        range,
-                                    )
-                                })
-                                .cloned()
-                                .collect();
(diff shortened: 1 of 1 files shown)


common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -179,6 +176,15 @@ impl Navigator for NavigationController {
                 let update = advance_step(remaining_steps);
                 match update {
                     StepAdvanceStatus::Advanced { step: current_step } => {
+                        // Trim the remaining waypoints if needed.
+                        let remaining_waypoints = self
+                            .has_new_waypoints_from_advance(
+                                &state.trip_state(),
+                                &current_step,
+                                WaypointCheckEvent::LocationUpdated,
```
ianthetechie: FIXME

common/ferrostar/src/navigation_controller/waypoint_advance.rs
```diff
@@ -0,0 +1,458 @@
...
+
+impl WaypointAdvanceChecker {
+    /// Returns a new set of waypoints if they are changed.
+    ///
+    /// # Parameters
+    ///
+    /// * `state` - The current trip state.
+    /// * `current_step` - The current route step being navigated.
+    /// * `event` - The event that triggered the waypoint check.
+    pub fn has_new_waypoints(
```
ianthetechie: Maybe a better name here.... `has_` sounds like it should return a boolean. 

```suggestion
    pub fn get_new_waypoints(
```

common/ferrostar/src/navigation_controller/waypoint_advance.rs
```diff
@@ -0,0 +1,458 @@
...
+        &self,
+        state: &TripState,
+        current_step: &RouteStep,
+        event: WaypointCheckEvent,
+    ) -> Option<Vec<Waypoint>> {
```
ianthetechie: Probably return an enum with variants like `Unchanged` and `Changed(Vec<Waypoint>)`

common/ferrostar/src/navigation_controller/waypoint_advance.rs
```diff
@@ -0,0 +1,458 @@
...
+        match state {
+            TripState::Navigating {
+                ref user_location,
+                ref remaining_waypoints,
+                ..
+            } => {
+                match self.mode {
+                    WaypointAdvanceMode::WaypointWithinRange(range) => {
+                        // Only advance waypoints if there are more than 1 remaining
```
ianthetechie: Probably add a guard here to make sure it's a location update event, because this method can actually be called twice! (Leave a comment to this effect too, since it's not necessarily obvious).

common/ferrostar/src/navigation_controller/waypoint_advance.rs
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    /// Helper function to check if a waypoint is within range of any point on a linestring
+    pub(crate) fn is_waypoint_within_range_of_linestring(
+        &self,
+        waypoint: &Point,
+        linestring: &geo::LineString,
+        range: f64,
+    ) -> bool {
+        match linestring.haversine_closest_point(waypoint) {
```
ianthetechie: Refactor to use `algorithms::deviation_from_line`

common/ferrostar/src/navigation_controller/waypoint_advance.rs
```diff
@@ -0,0 +1,458 @@
...
+    ///
+    /// # Parameters
+    ///
+    /// * `state` - The current trip state.
+    /// * `current_step` - The current route step being navigated.
+    /// * `event` - The event that triggered the waypoint check.
+    pub fn has_new_waypoints(
+        &self,
+        state: &TripState,
+        current_step: &RouteStep,
```
ianthetechie: This is only accessed from one match arm right now. We could either put it in the event enum or pull it from trip state (need to check that trip state has or does not have any modifications from prev call chains :D).

common/ferrostar/src/navigation_controller/waypoint_advance.rs
```diff
@@ -0,0 +1,458 @@
...
+
+        waypoints
+    }
+
+    proptest! {
+        /// Test WaypointWithinRange mode within range
+        #[test]
+        fn test_waypoint_within_range_always_advances(
+            user_lng in -180.0..180.0,
+            user_lat in -85.0..85.0,
```
ianthetechie: Let's update the ranges on all tests to be `-180..180` and `-90..90` respectively.

common/ferrostar/src/navigation_controller/waypoint_advance.rs
```diff
@@ -0,0 +1,458 @@
...
+
+            let user_location = create_user_location(step_start_lng, step_start_lat);
+
+            // Create a step with a line
+            let current_step = gen_route_step_with_coords(vec![
+                coord! { x: step_start_lng, y: step_start_lat },
+                coord! { x: step_end_lng, y: step_end_lat },
+            ]);
+
+            // Create waypoints far from the step line (guaranteed out of range)
```
ianthetechie: Move this comment down

common/ferrostar/src/navigation_controller/waypoint_advance.rs
```diff
@@ -0,0 +1,458 @@
...
+        }
+
+        /// Test that single waypoints (destinations) are never advanced in WaypointAlongAdvancingStep mode
+        #[test]
+        fn test_single_waypoint_never_advanced_along_step(
+            step_start_lng in -180.0..180.0,
+            step_start_lat in -85.0..85.0,
+            step_end_lng in -180.0..180.0,
+            step_end_lat in -85.0..85.0,
+            range_meters in 50.0..1000.0, // Detection range in meters
```
ianthetechie: Maybe make this a bit more like the above for better coverage?

common/ferrostar/src/navigation_controller/waypoint_advance.rs
```diff
@@ -98,11 +113,10 @@ impl WaypointAdvanceChecker {
...
     ) -> bool {
-        match linestring.haversine_closest_point(waypoint) {
-            Closest::Intersection(closest_point) | Closest::SinglePoint(closest_point) => {
-                Haversine.distance(*waypoint, closest_point) < range
-            }
-            Closest::Indeterminate => false,
+        if let Some(diff) = deviation_from_line(waypoint, linestring) {
+            diff <= range
+        } else {
+            false
```
ianthetechie: I only recently discovered this (thanks to `cargo clippy`), but this little combinator exists :)

```suggestion
        deviation_from_line(waypoint, linestring).is_some_and(|diff| diff <= range)
```

PR #718, opened 2025-10-19
React Native Expo example

bjtrounson:
This PR is nothing crazy, just switches the example app to expo. Also upgrades the uniffi-bindgen-react-native package to the latest. 

Can confirm that iOS now runs under this version.

### DIFF
--- a/react-native/example/package.json
+++ b/react-native/example/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "ferrostar-example-react-native",
+  "name": "@stadiamaps/ferrostar-example-react-native",
   "main": "expo-router/entry",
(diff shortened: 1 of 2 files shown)


react-native/example/assets/images/splash-icon.png
ianthetechie, line 1: Question from @Archdoog: Do these need to be checked in or will expo auto regen / use defaults? Concern is that SPM clones the whole repo 😅

ianthetechie (commented):
LGTM; just one question on assets

ianthetechie:
I say let's merge this for now and figure out later if we can remove stuff.

PR #721, opened 2025-10-21
Arbitrary extended waypoint properties

ianthetechie:
Our current `Waypoint` model is essentially the "least common denominator" of what all routing engines support. However, this limits some advanced use cases which use additional rich location properties. The Valhalla API for example supports like a dozen extra properties.

The goal of this PR is to enable extension of the `Waypoint` model with any number of additional properties, and ideally in a way that's type safe without too many paper cuts for users.

This necessarily makes a few trade-offs:

* We can no longer make very strong guarantees about what information is carried back to the `Route`'s list of `Waypoint`s. The example at hand is how Valhalla-based APIs often operate in OSRM mode, but the OSRM response drops all of those interesting waypoint properties.
* There are no cross-language generics in the bindings, nor is there support for constructors on record types in UniFFI. So, we fall back to the lowest common denominator of passing serialized bytes around with JSON as an interchange layer. It works 🤷  On the plus side, we did manage to keep it type safe.

Opening as a draft initially since there are some TODOs in the code still to address, and the initial commit doesn't have any mobile examples yet, but want to get feedback going.

### DIFF
--- a/apple/Sources/UniFFI/ferrostar.swift
+++ b/apple/Sources/UniFFI/ferrostar.swift
@@ -894,13 +894,13 @@ public protocol DistanceEntryAndSnappedExitConditionProtocol: AnyObject, Sendabl
  * The exit distance is measured from the current step to the route-snapped position.
  */
 open class DistanceEntryAndSnappedExitCondition: DistanceEntryAndSnappedExitConditionProtocol, @unchecked Sendable {
-    fileprivate let pointer: UnsafeMutableRawPointer!
+    fileprivate let handle: UInt64
 
-    /// Used to instantiate a [FFIObject] without an actual pointer, for fakes in tests, mostly.
+    /// Used to instantiate a [FFIObject] without an actual handle, for fakes in tests, mostly.
 #if swift(>=5.8)
     @_documentation(visibility: private)
 #endif
-    public struct NoPointer {
+    public struct NoHandle {
         public init() {}
     }
 
@@ -910,42 +910,39 @@ open class DistanceEntryAndSnappedExitCondition: DistanceEntryAndSnappedExitCond
 #if swift(>=5.8)
     @_documentation(visibility: private)
 #endif
-    required public init(unsafeFromRawPointer pointer: UnsafeMutableRawPointer) {
-        self.pointer = pointer
+    required public init(unsafeFromHandle handle: UInt64) {
+        self.handle = handle
     }
 
     // This constructor can be used to instantiate a fake object.
-    // - Parameter noPointer: Placeholder value so we can have a constructor separate from the default empty one that may be implemented for classes extending [FFIObject].
+    // - Parameter noHandle: Placeholder value so we can have a constructor separate from the default empty one that may be implemented for classes extending [FFIObject].
     //
     // - Warning:
-    //     Any object instantiated with this constructor cannot be passed to an actual Rust-backed object. Since there isn't a backing [Pointer] the FFI lower functions will crash.
+    //     Any object instantiated with this constructor cannot be passed to an actual Rust-backed object. Since there isn't a backing handle the FFI lower functions will crash.
 #if swift(>=5.8)
     @_documentation(visibility: private)
 #endif
-    public init(noPointer: NoPointer) {
-        self.pointer = nil
+    public init(noHandle: NoHandle) {
+        self.handle = 0
     }
 
 #if swift(>=5.8)
     @_documentation(visibility: private)
 #endif
-    public func uniffiClonePointer() -> UnsafeMutableRawPointer {
-        return try! rustCall { uniffi_ferrostar_fn_clone_distanceentryandsnappedexitcondition(self.pointer, $0) }
+    public func uniffiCloneHandle() -> UInt64 {
+        return try! rustCall { uniffi_ferrostar_fn_clone_distanceentryandsnappedexitcondition(self.handle, $0) }
     }
     // No primary constructor declared for this class.
 
     deinit {
-        guard let pointer = pointer else {
-            return
-        }
-
-        try! rustCall { uniffi_ferrostar_fn_free_distanceentryandsnappedexitcondition(pointer, $0) }
+        try! rustCall { uniffi_ferrostar_fn_free_distanceentryandsnappedexitcondition(handle, $0) }
     }
 
     
 
     
 
+    
 }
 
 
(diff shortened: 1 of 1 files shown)


common/ferrostar/src/routing_adapters/valhalla.rs
```diff
@@ -15,16 +15,131 @@ use alloc::{
...
+    /// NOTE: If the location is not offset from the road centerline,
+    /// or is very close to an intersection, this option has no effect!
+    pub preferred_side: Option<ValhallaWaypointPreferredSide>,
+    // TODO: display_lat + display_lon
+    // TODO: search_cutoff
+    // TODO: node_snap_tolerance
+    // TODO: street_side_tolerance
+    // TODO: street_side_max_distance
+    // TODO: street_side_cutoff
+    // TODO: search_filter
```
ianthetechie: These should be added before merging.

common/ferrostar/src/routing_adapters/valhalla.rs
```diff
@@ -155,6 +299,7 @@ impl RouteRequestGenerator for ValhallaHttpRequestGenerator {
         } else {
             let headers =
                 HashMap::from([("Content-Type".to_string(), "application/json".to_string())]);
+            // TODO: Figure out how / if we want waypoint properties for the initial location
```
ianthetechie, line 398: Not required before merging. I don't know if anyone specifically needs these. We already try to set good defaults like heading and location accuracy from the device GNSS sensors.

apple/Tests/FerrostarCoreTests/__Snapshots__/FerrostarCoreTests/test200MockGETRouteResponse.1.txt
ianthetechie, line 1: Most of these will have no properties since they use a mock response parser that returns a specifically constructed value which doesn't include any waypoint properties.

apple/Tests/FerrostarCoreTests/__Snapshots__/FerrostarCoreTests/testValhallaCostingOptionsJSON.1.txt
```diff
@@ -45,11 +45,15 @@
           - lat: 60.534715
           - lng: -149.543469
         - kind: break
+        ▿ properties: Optional<Data>
+          - some: 30 bytes
```
ianthetechie, line 49: These last 2 tests actually parse "real" responses. I'm not stoked about these just being bytes in the fixture, but on the other hand we have plenty of tests on the Rust side so I'm not inclined to put any more time into this unless someone already has a fix for decoding this as UTF-8 for snapshotting.

common/ferrostar/Cargo.toml
```diff
@@ -35,6 +35,7 @@ polyline = "0.11.0"
 rand = "0.9.2"
 serde = { version = "1.0.219", features = ["derive"] }
 serde_json = { version = "1.0.145", default-features = false }
+serde_with = "3.15.1"
```
ianthetechie, line 38: New dependency; this extends some of the classic serde attributes. The one I use in this PR skips serialization for optional fields with a value of `None`. It's possible to do this with `#[serde(skip_serializing_if = "Option::is_none")]` for every single field, but this gets tedious and error prone.

common/ferrostar/src/models.rs
```diff
@@ -97,13 +97,34 @@ impl From<GeographicCoordinate> for Point {
...
+    /// MUST use JSON.
+    ///
+    /// We selected JSON is selected not because it is good,
+    /// but because generics (i.e., becoming `Waypoint<T>`, where an example `T` is `ValhallaProperties`)
+    /// would be way too painful, particularly for foreign code.
+    /// Especially JavaScript.
+    ///
+    /// In any case, [`crate::routing_adapters::RouteRequestGenerator`] and [`crate::routing_adapters::RouteResponseParser`]
+    /// implementations SHOULD document their level support for this,
+    /// ideally with an exportable record type.
```
ianthetechie, line 125: See here for the main documentation of the design.

common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -145,7 +145,7 @@ impl Navigator for NavigationController {
             snapped_user_location,
             remaining_steps,
             // Skip the first waypoint, as it is the current one
-            remaining_waypoints: self.route.waypoints.iter().skip(1).copied().collect(),
+            remaining_waypoints: self.route.waypoints.iter().skip(1).cloned().collect(),
```
ianthetechie, line 148: Had to drop `Copy` since we added properties.

common/ferrostar/src/navigation_session/caching/mod.rs
```diff
@@ -1,8 +1,10 @@
+#[cfg(feature = "uniffi")]
 use std::sync::{Arc, Mutex};
 
+#[cfg(feature = "uniffi")]
 use chrono::{Duration, Utc};
 
-#[cfg(all(feature = "uniffi", not(feature = "wasm_js")))]
+#[cfg(feature = "uniffi")]
```
ianthetechie, line 7: Bunch of minor cleanup here. Just the uniffi gate is enough. We can't target wasm with uniffi (yet?), so I changed this to match the rest of the codebase.

common/ferrostar/src/navigation_session/caching/mod.rs
```diff
@@ -76,29 +78,13 @@ impl NavigationSessionCache {
     }
 }
 
-#[cfg(all(feature = "uniffi", not(feature = "wasm_js")))]
+#[cfg(feature = "uniffi")]
 #[cfg_attr(feature = "uniffi", uniffi::export)]
 impl NavigationObserver for NavigationSessionCache {
-    fn on_route_available(&self, route: Route) {
```
ianthetechie, line 82: Lint: usually impls should use the same order as the trait for consistency.

common/ferrostar/src/routing_adapters/osrm/mod.rs
```diff
@@ -87,6 +96,17 @@ impl Route {
                 } else {
                     WaypointKind::Break
                 },
+                properties: if waypoint.name.is_some() || waypoint.distance.is_some() {
+                    Some(
+                        serde_json::to_vec(&OsrmWaypointProperties {
+                            name: waypoint.name.clone(),
+                            distance: waypoint.distance,
+                        })
+                        .expect("Infallible JSON serialization"),
```
ianthetechie, line 105: This can't fail unless the data structure itself (`OsrmWaypointProperties`) does something that's illegal in JSON like non-string map keys, so this is reasonable to panic on.

common/ferrostar/src/models.rs
```diff
@@ -97,13 +97,34 @@ impl From<GeographicCoordinate> for Point {
...
+    #[cfg_attr(feature = "uniffi", uniffi(default))]
```
ianthetechie, line 126: I discovered a new UniFFI feature 🎉

This lets us mark a property of a record type (struct) as optional. At the FFI level, we get much easier constructors. This is especially noticeable in Swift.

ianthetechie:
I should see if we can add a JavaScript unit test for the waypoint properties 🤔

common/ferrostar/src/routing_adapters/valhalla.rs
```diff
@@ -15,16 +15,227 @@ use alloc::{
...
+
+/// Waypoint properties supported by Valhalla servers.
+///
+/// Our docstrings are short here, since Valhalla is the final authority.
+/// Refer to <https://valhalla.github.io/valhalla/api/turn-by-turn/api-reference/#locations>
+/// for more details, including default values.
+/// Other Valhalla-based APIs such as Stadia Maps or Mapbox may have slightly different defaults.
+/// Refer to your vendor's documentation when in doubt.
+///
+/// NOTE: Waypoint properties will NOT currently be echoed back by Valhalla-based servers,
```
ianthetechie:
```suggestion
/// NOTE: Waypoint properties will NOT currently be echoed back in OSRM format,
```

Issue #717, opened 2025-10-16
Could `DistanceEntryAndExitCondition` use route-snapped position for exit check?


devnull133:
Hi there, 

First, thanks for the great work and clean framework!

I've been testing Ferrostar for pedestrian navigation and noticed that `DistanceEntryAndExitCondition` uses raw GPS position for the exit distance check. I think this causes premature step advancement in urban environments when the user might have a constant distance from the step polyline (e.g., simply walking on the opposite sidewalk). In my testing, we correctly trigger the entry condition being 20-30m from the step end, but then almost immediately trigger the exit condition (being a few meters away from the step polyline).

I can't use a larger exit threshold (like 20m) because that breaks tight urban sequences where multiple turns happen within 5-15m of each other.

I could imagine a solution where we don't use raw GPS position for the exit condition, but instead snap to the current step + next step polyline combined, and only then measure distance to the current step polyline. That would make the exit condition trigger on a real exit rather than being triggered by some small off-route condition (common in pedestrian navigation or when GPS accuracy is poor).

Am I missing something, or would this be a reasonable enhancement?

PR #722, opened 2025-10-21
Add DistanceEntryAndSnappedExitCondition

devnull133:
Hi,

this is a proposal to fix issue described here: https://github.com/stadiamaps/ferrostar/issues/717

I believe we could simply update the existing `DistanceEntryAndExitCondition`, but I'm adding a new variant called `DistanceEntryAndSnappedExitCondition` so that it's easier to compare their behaviors, run real-world tests, and then decide whether the swap makes sense.

Let's assume following config:
```
stepAdvanceDistanceEntryAndExit(
  distanceToEndOfStep: 25,
  distanceAfterEndOfStep: 5,
  minimumHorizontalAccuracy: NavigationConstants.minimumHorizontalAccuracy
)
```
and the situation:
Screenshot 2025-10-21 at 13 06 11

The large circle has a radius of 25 m. The distance between the snapped and real user positions is greater than 5 m. The user is moving north and is expected to make a right turn toward the east.

The difference is that the current `DistanceEntryAndExitCondition` immediately advances the step in the state shown above. This behavior isn’t ideal - it’s very noticeable during pedestrian mode navigation testing.

The new `DistanceEntryAndSnappedExitCondition` does not behave this way. The user position is still snapped to both the current and next step, but the distance is calculated only from the current step. This means the distance remains zero until the user actually reaches the turn. Then, as the snapping transitions to the next step, the user starts increasing the distance from the previous step - eventually triggering the advance condition.

There are still some quirks with very short next steps or unusual orientations, but my gut feeling is that this approach shouldn’t be any worse than what we currently have. I tested that pretty extensively in the real word, while hiking and navigating in the city, and it worked well. But probably it'd be good to consider some edge cases.

### DIFF
--- a/common/ferrostar/src/navigation_controller/step_advance/conditions.rs
+++ b/common/ferrostar/src/navigation_controller/step_advance/conditions.rs
@@ -495,10 +495,7 @@ impl StepAdvanceCondition for DistanceEntryAndSnappedExitCondition {
                 // Check GPS accuracy first
                 if user_location.horizontal_accuracy > self.minimum_horizontal_accuracy.into() {
                     return StepAdvanceResult::continue_with_state(Arc::new(
-                        DistanceEntryAndSnappedExitCondition {
-                            has_reached_end_of_current_step: true,
-                            ..*self
-                        },
+                        DistanceEntryAndSnappedExitCondition { ..*self },
                     ));
                 }
 
@@ -530,10 +527,7 @@ impl StepAdvanceCondition for DistanceEntryAndSnappedExitCondition {
                     StepAdvanceResult::advance_to_new_instance(self)
                 } else {
                     StepAdvanceResult::continue_with_state(Arc::new(
-                        DistanceEntryAndSnappedExitCondition {
-                            has_reached_end_of_current_step: true,
-                            ..*self
-                        },
+                        DistanceEntryAndSnappedExitCondition { ..*self },
                     ))
                 }
             } else {


Archdoog:
I haven't written this up (likely in an RFD). But I'm thinking we should just update the `StepAdvanceCondition`'s `fn should_advance_step` function to:

```rust
    fn should_advance_step(
        &self,
        trip_state: TripState
    ) -> StepAdvanceResult;
```

This would make it really to easy use whatever params are best for each step advance condition. To keep things clean, we'd just extend `TripState` itself with functions like `fn next_step() -> Option<RouteStep>` instead of passing them in as individual params.

## Reasoning behind this:

We're having the same thing happen with RouteDeviation, where certain conditions and scenarios can be fixed or further refined with various params (e.g. remaining steps for skipping forward, the snapped location for this, etc). TripState has all of the information, so we might as well just pass use that consistently in everything.

common/ferrostar/src/navigation_controller/step_advance/conditions.rs
```diff
@@ -428,6 +428,158 @@ impl StepAdvanceConditionSerializable for DistanceEntryAndExitCondition {
...
+    ) -> StepAdvanceResult {
+        if self.has_reached_end_of_current_step {
+            // EXIT CHECK: Use route-snapped position
+
+            if let Some(next) = next_step {
+                // Check GPS accuracy first
+                if user_location.horizontal_accuracy > self.minimum_horizontal_accuracy.into() {
+                    return StepAdvanceResult::continue_with_state(Arc::new(
+                        DistanceEntryAndSnappedExitCondition {
+                            has_reached_end_of_current_step: true,
```
ianthetechie: Quick sanity check... if I'm reading this correctly, we are recording that the user has reached the end of the current step automatically whenever they have poor GPS accuracy?

common/ferrostar/src/navigation_controller/step_advance/conditions.rs
```diff
@@ -428,6 +428,158 @@ impl StepAdvanceConditionSerializable for DistanceEntryAndExitCondition {
...
+            // ENTRY CHECK: Use existing distance to end logic
+            let distance_to_end = DistanceToEndOfStepCondition {
+                minimum_horizontal_accuracy: self.minimum_horizontal_accuracy,
+                distance: self.distance_to_end_of_step,
+            };
+
+            let next_iteration = DistanceEntryAndSnappedExitCondition {
+                minimum_horizontal_accuracy: self.minimum_horizontal_accuracy,
+                distance_to_end_of_step: self.distance_to_end_of_step,
+                distance_after_end_of_step: self.distance_after_end_of_step,
```
ianthetechie: Minor nit: i think these can be replaced with `..self`.

common/ferrostar/src/navigation_controller/step_advance/conditions.rs
```diff
@@ -428,6 +428,158 @@ impl StepAdvanceConditionSerializable for DistanceEntryAndExitCondition {
...
+
+            StepAdvanceResult::continue_with_state(Arc::new(next_iteration))
+        }
+    }
+
+    fn new_instance(&self) -> Arc<dyn StepAdvanceCondition> {
+        Arc::new(DistanceEntryAndSnappedExitCondition {
+            distance_to_end_of_step: self.distance_to_end_of_step,
+            distance_after_end_of_step: self.distance_after_end_of_step,
+            minimum_horizontal_accuracy: self.minimum_horizontal_accuracy,
```
ianthetechie: Ditto `..self` pattern

ianthetechie (commented):
Thanks for this PR and the excellent comments + tests!

I think your analysis of (one of the) issues for pedestrian, hiking, cycling, etc. is spot on. And it's a pretty cool leveraging of the flexibility we have by splitting up step advance and off route detection! Reading your comments on the method, it seems to match what my favorite cycling GPS does, precisely to account for the fact that you may be off on some unmapped sidewalk on the "wrong" side of the road, etc.

I left a few comments, after which I'm good to merge this.

ianthetechie:
> I believe we could simply update the existing DistanceEntryAndExitCondition, but I'm adding a new variant called DistanceEntryAndSnappedExitCondition so that it's easier to compare their behaviors, run real-world tests, and then decide whether the swap makes sense.

I also had the same thought on whether we should combine these. I think this is fine for now too. If it seems to work out well, I think it's worth a quick revisit to combine them. Though there were enough differences in the logical pattern that maybe it still ends up being separate 🤷

common/ferrostar/src/navigation_controller/step_advance/conditions.rs
devnull133: Thanks! Note that we're already in the if block:
`if self.has_reached_end_of_current_step {`
so this was redundant - but hopefully harmless. I removed it in the last commit.

PR #723, opened 2025-10-22
Feat/common/step advance trip state

Archdoog:
- Adds getters to TripState for uniform access to params.
- Converts StepAdvanceConditon's to full `TripState` to enable better behaviors based on ideal params.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -16,19 +16,9 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install latest mdbook
-        run: |
-          tag=$(curl 'https://api.github.com/repos/rust-lang/mdbook/releases/latest' | jq -r '.tag_name')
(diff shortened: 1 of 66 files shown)


common/ferrostar/src/navigation_controller/step_advance/conditions.rs
```diff
@@ -560,6 +480,70 @@ impl StepAdvanceCondition for DistanceEntryAndSnappedExitCondition {
     }
 }
 
+impl DistanceEntryAndSnappedExitCondition {
```
Archdoog, line 483: @devnull133 This is your new condition, revised to accept a `TripState`. Feel free to provide any feedback on any restructuring that makes it more confusing.

devnull133: @Archdoog, thanks, it looks very good to me 👍

PR #724, opened 2025-10-28
Feat: converted route devation to trip state pattern

Archdoog:
- Converts the RouteDeviation tooling to use TripState instead of UserLocation & Current Step.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,7 @@ import PackageDescription
 
 let binaryTarget: Target
 let maplibreSwiftUIDSLPackage: Package.Dependency
-let useLocalFramework = true
+let useLocalFramework = false
 let useLocalMapLibreSwiftUIDSL = false
 
 if useLocalFramework {


common/ferrostar/src/navigation_controller/mod.rs
```diff
@@ -148,11 +144,48 @@ impl Navigator for NavigationController {
...
+            deviation: RouteDeviation::NoDeviation,
             visual_instruction,
             spoken_instruction,
             annotation_json,
         };
+
+        let deviation = self
+            .config
+            .route_deviation_tracking
+            .check_route_deviation(&self.route, &initial_trip_state);
```
ianthetechie, line 156: I found this to be a bit confusing during a first read. I think it probably makes sense to calculate the deviation first and assign and match it below.

common/ferrostar/src/deviation_detection.rs
```diff
@@ -69,43 +72,48 @@ impl RouteDeviationTracking {
...
+                        }
+
+                        if matches!(step_deviation, RouteDeviation::NoDeviation) {
+                            return RouteDeviation::NoDeviation;
+                        }
                     }
-                } else {
-                    RouteDeviation::NoDeviation
+
+                    first_step_deviation.unwrap_or(RouteDeviation::NoDeviation)
```
ianthetechie, line 112: I *think* the logic here boils down to only looking at the first step deviation, so looking at any of the others doesn't do anything?

If that's what you intend, then you could just do a `map_or` over `remaining_steps.first()`, where the default value is `NoDeviation`.

ianthetechie: Discussed on the call: we'll add a comment here explaining what it does (my reading was incorrect; we actually do loop through all in order to find if any step seems to be a good match for the user's current position!).

Jacob said this needs a bit of real world testing and just wanted to get something down first :)

common/ferrostar/src/navigation_controller/mod.rs
ianthetechie: Okay, so this DOES actually depend on the value of `initial_trip_state`... Because deviation depends on trip state ;)

I guess we just need another comment here explaining the flow.

devnull133:
I'm holding my breath for this one :) I'm going to merge this branch into my local build and will let you know soon how it works for me.

devnull133:
Actually, @Archdoog, would you mind resolving merge conflicts in this branch? I could do it myself, but if I'm about to test, I'd rather test what's going to be in `main` branch later. :)

devnull133:
I mostly ran synthetic tests with SimGenie, but it seems to work well. If I move back to the beginning of my route, it's now considered an off-route state, and Ferrostar won't advance all the steps to the end. Thanks!

PR #740, opened 2025-11-16
Add hysteresis to route deviation tracking

devnull133:
Adds `on_route_threshold` parameter to `StaticThreshold` route deviation tracking. When set lower than `max_acceptable_deviation`, this creates a buffer zone that prevents users from oscillating between on-route and off-route states near the threshold boundary.

For example, with max_acceptable_deviation=50m and on_route_threshold=40m:
- User goes off-route when deviation exceeds 50m
- User returns on-route only when deviation drops to 40m or less
- This 10m buffer prevents rapid state changes when GPS accuracy fluctuates

This is especially useful for slow-paced navigation, such as when walking. Otherwise, when the user is at the max_acceptable_deviation, on- and off-route alerts keep triggering constantly, which can be quite annoying.

### DIFF
--- a/common/ferrostar/src/deviation_detection.rs
+++ b/common/ferrostar/src/deviation_detection.rs
@@ -24,6 +24,120 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "wasm-bindgen")]
 use tsify::Tsify;
 
+/// Errors that can occur when creating a [`StaticThresholdConfig`].
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Error))]
+pub enum StaticThresholdError {
+    /// The maximum acceptable deviation must be non-negative.
+    #[cfg_attr(
+        feature = "std",
+        error("max_acceptable_deviation must be non-negative, got {value}")
+    )]
+    NegativeMaxDeviation { value: f64 },
+    /// The return buffer must be non-negative.
+    #[cfg_attr(
+        feature = "std",
+        error("return_buffer must be non-negative, got {value}")
+    )]
+    NegativeReturnBuffer { value: f64 },
+    /// The return buffer must not exceed the maximum acceptable deviation.
+    #[cfg_attr(
+        feature = "std",
+        error(
+            "return_buffer ({return_buffer}) must not exceed max_acceptable_deviation ({max_acceptable_deviation})"
+        )
+    )]
+    ReturnBufferTooLarge {
+        return_buffer: f64,
+        max_acceptable_deviation: f64,
+    },
+}
+
+/// Configuration for static threshold route deviation detection with hysteresis support.
+///
+/// This struct ensures valid configuration through a failable constructor,
+/// making it impossible to create invalid threshold configurations.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm-bindgen", derive(Tsify))]
+#[cfg_attr(feature = "wasm-bindgen", tsify(from_wasm_abi))]
+pub struct StaticThresholdConfig {
+    /// The minimum required horizontal accuracy of the user location, in meters.
+    /// Values larger than this will not trigger route deviation warnings.
+    pub minimum_horizontal_accuracy: u16,
+    /// The maximum acceptable deviation from the route line, in meters.
+    ///
+    /// If the distance between the reported location and the expected route line
+    /// is greater than this threshold, it will be flagged as an off route condition.
+    pub max_acceptable_deviation: f64,
+    /// The buffer distance used for hysteresis when returning to on-route state, in meters.
+    ///
+    /// The actual threshold for returning to on-route is calculated as:
+    /// `max_acceptable_deviation - return_buffer`
+    ///
+    /// For example, if `max_acceptable_deviation` is 50m and `return_buffer` is 10m,
+    /// the user must deviate more than 50m to trigger off-route, but must return within
+    /// 40m to be considered back on route.
+    ///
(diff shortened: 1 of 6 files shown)


common/ferrostar/src/deviation_detection.rs
```diff
@@ -58,6 +58,16 @@ pub enum RouteDeviationTracking {
         /// If the distance between the reported location and the expected route line
         /// is greater than this threshold, it will be flagged as an off route condition.
         max_acceptable_deviation: f64,
+        /// The threshold for returning to on-route state, in meters.
+        ///
+        /// Must be less than or equal to `max_acceptable_deviation`.
+        /// This creates hysteresis to prevent oscillation between on/off route states.
```
ianthetechie: Nice $100 word usage 😂

common/ferrostar/src/deviation_detection.rs
```diff
@@ -58,6 +58,16 @@ pub enum RouteDeviationTracking {
...
+        /// For example, if `max_acceptable_deviation` is 50m and `on_route_threshold` is 40m,
+        /// the user must deviate more than 50m to trigger off-route, but must return within
+        /// 40m to be considered back on route.
+        ///
+        /// If not specified or equal to `max_acceptable_deviation`, no hysteresis is applied.
+        on_route_threshold: f64,
```
ianthetechie: I think the API on the Swift side is great. But at the lower level, it's currently too easy for a user to accidentally violate the invariants (e.g. you have a note about behavior when it's equal, but what about when it's greater?). See https://cliffle.com/blog/rust-typestate/.

I think we should probably create a new struct with a failable constructor that captures the desired behavior, since we can't do this with enums. Basically we should make it so it's not possible to accidentally construct a nonsensical value :)

common/ferrostar/src/deviation_detection.rs
```diff
@@ -80,24 +90,32 @@ impl RouteDeviationTracking {
...
+                    };
+
                     let mut first_step_deviation = None;
 
                     for (index, step) in remaining_steps.iter().enumerate() {
                         let step_deviation = self.static_threshold_deviation_from_line(
                             &Point::from(*user_location),
                             &step.get_linestring(),
-                            max_acceptable_deviation.clone(),
+                            threshold,
```
ianthetechie, line 210:
```suggestion
                            // Note: This is always <= max_acceptable_deviation
                            threshold,
```

common/ferrostar/src/deviation_detection.rs
devnull133: Haha, indeed $100 - straight from my $100/month Claude MAX plan 😄
I'm happy to reword it, however "hysteresis" is a pretty accurate term, even though this particular implementation is dead simple. Probably it could have skipped "oscillation" not to scare humans away.

common/ferrostar/src/deviation_detection.rs
devnull133: You basically mean the scenario when user provides on_route_threshold > max_acceptable_deviation, right?

Another approach could be to specify max_acceptable_deviation and `buffer_width` (well, or come up with some better name). Then, it'd be rather impossible (or at least more difficult) to specify incorrect values. But I wasn't sure which one is easier to understand, and finally decided to go with more explicitly named "on_route_threshold".

common/ferrostar/src/deviation_detection.rs
ianthetechie: Haha nah I'm fine leaving it. It's a great word. Dictionaries and LLMs exist. Excuse my reviewing style which includes plenty of humor and snide remarks when not actually asking for a change; I'll make that obvious otherwise 😂

common/ferrostar/src/deviation_detection.rs
ianthetechie: > You basically mean the scenario when user provides on_route_threshold > max_acceptable_deviation, right?

Yup. Additionally, you say (at least in the comment) that it can be "not specified" but that isn't actually possible at the moment, since it's a (non-optional) float 😂  

I don't actually have a strong opinion on this and I think you've done more thinking about the use case than me, but from an API perspective I'd like to see it designed so it's either impossible or at least difficult to misuse. The `buffer_width` (or better name) idea does seem to satisfy this pretty well, but I don't really have a strong opinion.

One more bit that may not have been obvious in my hasty comment: the reason I suggest a separate struct is because that lets you do validation work in the constructor. A public enum in Rust is public _all the way down_ by necessity, so to introduce something like a "failable" constructor you need to add another type (e.g. a struct) in the middle.

common/ferrostar/src/deviation_detection.rs
```diff
@@ -24,6 +24,120 @@ use serde::{Deserialize, Serialize};
...
+    NegativeMaxDeviation { value: f64 },
+    /// The return buffer must be non-negative.
+    #[cfg_attr(
+        feature = "std",
+        error("return_buffer must be non-negative, got {value}")
+    )]
+    NegativeReturnBuffer { value: f64 },
+    /// The return buffer must not exceed the maximum acceptable deviation.
+    #[cfg_attr(
+        feature = "std",
```
ianthetechie, line 46: Unrelated to this PR, I just realized we don't need to gate this anymore. Tracked separately in #747.

common/ferrostar/src/deviation_detection.rs
```diff
@@ -24,6 +24,120 @@ use serde::{Deserialize, Serialize};
...
+    ///
+    /// The actual threshold for returning to on-route is calculated as:
+    /// `max_acceptable_deviation - return_buffer`
+    ///
+    /// For example, if `max_acceptable_deviation` is 50m and `return_buffer` is 10m,
+    /// the user must deviate more than 50m to trigger off-route, but must return within
+    /// 40m to be considered back on route.
+    ///
+    /// Set to 0 for no hysteresis (same threshold for going off-route and returning).
+    pub return_buffer: f64,
```
ianthetechie, line 84: We have a constructor, but it looks like you can bypass this all with the `pub` fields. I _think_ we can just hide these and expose accessors, but I'll have to check; @Archdoog noted that maybe UniFFI records require this but I'll check.

PR #752, opened 2025-12-06
newbie

jenilcodes25:


jenilcodes25:
@ianthetechie Thanks for the clarification! That makes sense—handling this at the platform layer seems more appropriate given the application-specific nature of “paused.” I’ll draft an RFD to explore this further. Appreciate the guidance!

PR #755, opened 2025-12-17
[Bug] : Fix isMinimumTimeBeforeRecalcExceeded

radityagumay:
## Description

Earlier the time comparison is using nanoseconds instead of seconds, the PR essentially to use TimeUnit.SECONDS to fix aggressive nano time usage.

In this PR i also extract the functionality into a function so that i can add a unit test to verify the logic.

## Safety net
Providing unit tests to validate the logic < and > of 5 seconds

### DIFF
--- a/android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
+++ b/android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
@@ -73,13 +73,13 @@ class FerrostarCore(
   }
 
   /**
-   * The minimum time (in seconds) to wait before initiating another route recalculation.
+   * The minimum duration to wait before initiating another route recalculation.
    *
    * This matters in the case that a user is off route, the framework calculates a new route, and
    * the user is determined to still be off the new route. This adds a minimum delay (default 5
    * seconds).
    */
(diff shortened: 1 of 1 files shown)


ianthetechie:
For future PRs, note that we use `ktFmtFormat` to avoid this sort of formatting noise 😅 https://stadiamaps.github.io/ferrostar/dev-env-setup.html#pr-checklist-3

android/core/src/main/java/com/stadiamaps/ferrostar/core/FerrostarCore.kt
```diff
@@ -424,6 +423,18 @@ class FerrostarCore(
...
+  /**
+   * Has the user moved enough since the time we recalculated the route?
+   *
+   * This predicate avoids rapid recomputation when the route is unlikely to change.
+   */
+  private fun hasUserMovedSignificantlySinceLastRecalc(location: UserLocation): Boolean {
+    return _lastRecalculationLocation?.let {
+      it.toAndroidLocation().distanceTo(location.toAndroidLocation()) >
+          minimumMovementBeforeRecalculation
+    } ?: true // Default to true if no prior automatic recalculation
```
radityagumay, line 435: this is make sense, i am about to make a changes of this. thank you @ianthetechie

PR #802, opened 2026-02-11
[Android] Chunk of dependency upgrades

ianthetechie:
This addresses the issue @Archdoog flagged when upgrading the Compose BOM + getting a runtime crash from the autocomplete search widget.

Gradle 9 is a slightly larger upgrade so I'll do that separately.


ianthetechie:
@Archdoog converting this back to draft status since the scope had to balloon to JDK 21 and AGP 9.

1. There is some UI breakage as a result of the upgrade. I've committed snapshots so that the breakage is visible in the PR.
2. The AGP / Gradle 9 upgrade currently uses some compatibility flags to defer changes. This is fine IMO for now, but we'll need to fix this eventually.

PR #806, opened 2026-02-24
chore: remove std feature gates from thiserror derives (fixes #747)

### LINKED ISSUE #747: Clean up error feature flags
Currently we feature flag a lot of stuff (thiserror derives mostly) behind a `std` gate. The `error` module was moved from `std` to `core` though in 1.81, so this is no longer something we need to gate. Our MSRV is now long since past that so we can clean this up and un-gate.

Praful0306:
Closes #747

Since Rust 1.81, `core::error::Error` is stable and `thiserror` uses it 
unconditionally. The `std` feature gate on all `thiserror` derives is no 
longer needed.

## Changes
- Removed all `#[cfg_attr(feature = "std", derive(thiserror::Error))]` guards
- Removed all `#[cfg_attr(feature = "std", error("..."))]` guards on variants
- `thiserror::Error` is now derived unconditionally on all error enums

## Files changed
- `common/ferrostar/src/routing_adapters/error.rs`
- `common/ferrostar/src/models.rs`
- `common/ferrostar/src/simulation.rs`
- `common/ferrostar/src/navigation_session/recording/models.rs`

### DIFF
--- a/common/ferrostar/src/simulation.rs
+++ b/common/ferrostar/src/simulation.rs
@@ -10,7 +10,7 @@
 //!
 //! ```
 //! use ferrostar::simulation::{advance_location_simulation, location_simulation_from_polyline, LocationBias};
-//! # use std::error::Error;
+//! # use core::error::Error;
 //! # fn main() -> Result<(), Box<dyn Error>> {
 //!
 //! let polyline_precision = 6;


ianthetechie (approved):
Thanks for the initiative in cleaning this up!

PR #814, opened 2026-03-09
Add support for Valhalla waypoint U-turn restrictions

ianthetechie:
This PR effectively adds support for the `through` and `break_through` Valhalla waypoint types.

I personally don't like the Valhalla API that much, since it's a lot of mental overhead just to figure out that these are the same as `via` and `break`, except that they explicitly disallow U-turns.

So, in the interest of making this both clearer to the user and exploding the `WaypointKind` enum with values that don't make sense outside Valhalla, I've added an `allow_uturns` flag to `ValhallaWaypointProperties`. To make this visible to users with more experience around Valhalla than Ferrostar, I've documented the translation in the guide.

NOTE: Since we're using the properties object, this will _not_ be immediately visible to JavaScript users.

### DIFF
--- a/guide/src/route-providers.md
+++ b/guide/src/route-providers.md
@@ -109,7 +109,8 @@ Here's an example in Swift:
 let waypoint = createWaypointWithValhallaProperties(
     coordinate: GeographicCoordinate(lat: 60.5349908, lng: -149.5485806),
     kind: .break,
-    properties: ValhallaWaypointProperties(preferredSide: .same, allowUturns: false))
+    properties: ValhallaWaypointProperties(preferredSide: .same,
+	allowUturns: false))
 ```
 
 > [!TIP]


guide/src/route-providers.md
```diff
@@ -106,11 +106,18 @@ This is exposed via a top-level helper function named `create_waypoint_with_valh
 Here's an example in Swift:
 
 ```swift
-let waypoint = createWaypointWithValhallaProperties(coordinate: GeographicCoordinate(lat: 60.5349908, lng: -149.5485806),
-                                                    kind: .break,
-													properties: ValhallaWaypointProperties(preferredSide: .same))
+let waypoint = createWaypointWithValhallaProperties(
+    coordinate: GeographicCoordinate(lat: 60.5349908, lng: -149.5485806),
+    kind: .break,
+    properties: ValhallaWaypointProperties(preferredSide: .same, allowUturns: false))
```
ianthetechie: Make this go down to a separate line

PR #822, opened 2026-03-15
feat: shared ui for items that apply to compose-ui and car app

Archdoog:
- Create standalone shared ui package.
- Moves drawable icons to `ui.shared`
- Creates maneuver icon type,

### DIFF
--- a/android/ui-shared/build.gradle
+++ b/android/ui-shared/build.gradle
@@ -27,8 +27,8 @@ android {
     }
     compileOptions {
         coreLibraryDesugaringEnabled = true
-        sourceCompatibility JavaVersion.VERSION_11
-        targetCompatibility JavaVersion.VERSION_11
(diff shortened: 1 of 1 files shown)


android/ui-shared/build.gradle
```diff
@@ -0,0 +1,70 @@
...
+        signAllPublications()
+    }
+
+    configure(new AndroidSingleVariantLibrary("release", true, true))
+
+    apply from: "${rootProject.projectDir}/common-pom.gradle"
+
+    pom {
+        name = "Ferrostar Formatters"
+        description = "Distance, duration and arrival formatters for Ferrostar"
```
Archdoog:
```suggestion
        name = "Ferrostar Shared UI"
        description = "Shared UI that applies to both Compose & CarApp"
```

Issue #165, opened 2024-07-26
Android - Create default MapViewCamera for other navigation modes


Archdoog:
Currently the `navigating()` MapViewCamera extension is tailored for automotive navigation. We should upgrade this methodology to allow specifying the a default navigation camera for the current activity (walking, cycling, driving).

Duplicate this ticket if needed (e.g. if iOS only is completed).

PR #824, opened 2026-03-16
Other Nav Modes Camera and Non Navigation UI Improvements 

hactar:
This PR tries to bring iOS closer to Android by porting parts of the commits:

- 0d3c0882 (Android route overview / isNavigating overlay gating)
- 0ee36b82 (adds NavigationActivity camera modes tied to #165)

This PR should therefore also address #196 

Concrete changes:
- Camera not tilted in non navigation mode
- buttons hidden in non navigation mode
- demo button for choosing destination moved outside the .navigation grid, cause its not a navigation button (this is different to the android set up)
- adjustments to the demo relative path of where to find the ferrostar lib (no longer requires the root folder to be called ferrostar)
- function and test adjustments as iOS was adding the current road label even if it was nil, which android didn't

Video of the demo app before changes (notice the camera tilt at the beginning and the nav buttons visible on the right side) :
https://github.com/user-attachments/assets/74b97f55-98bf-4798-8d8a-cc3d6642704c

Video of the demo app after changes (no tilt, nav buttons only visible during nav):
https://github.com/user-attachments/assets/9e8532c4-7595-4d2e-8932-3ad7a8e705a7

Please note that things go wrong if you try to start a second navigation, she search field no longer disappears. But this happens before these changes too, so I didn't address it to not further inflate this PR.


apple/Sources/FerrostarMapLibreUI/Extensions/MapViewCamera.swift
```diff
@@ -21,4 +59,28 @@ public extension MapViewCamera {
...
+    ///
+    /// - Parameters:
+    ///   - zoom: The zoom value (default is 20.0)
+    ///   - pitch: The pitch (default is 10.0)
+    /// - Returns: The configured MapViewCamera
+    static func pedestrianNavigation(zoom: Double = 20.0, pitch: Double = 10.0) -> MapViewCamera {
+        MapViewCamera.trackUserLocationWithCourse(zoom: zoom,
+                                                  pitch: pitch,
+                                                  pitchRange: .fixed(pitch))
+    }
```
ianthetechie, line 94: I'm guessing these are mostly to allow overriding the defaults, right? Maybe there's a cleaner way we can express that. As-is, we have the same constant values both in the enum and here in the default arguments, so there's room for them to drift by accident.

apple/Tests/FerrostarMapLibreUITests/Views/DynamicallyOrientingNavigationViewTests.swift
```diff
@@ -40,9 +40,11 @@ final class DynamicallyOrientingViewTests: XCTestCase {
                     .padding()
             }
             .navigationViewCurrentRoadView { state in
-                Text("Current Road: \(state?.currentRoadName ?? "unknown")")
-                    .background(Color.yellow)
-                    .padding()
+                if let currentRoadName = state?.currentRoadName, !currentRoadName.isEmpty {
```
ianthetechie: I feel like we lost something with this test since the view no longer renders :)

apple/DemoApp/Ferrostar Demo.xcodeproj/project.pbxproj
```diff
@@ -290,7 +290,7 @@
 			mainGroup = E9505FAE2AD449700016BF0A;
 			packageReferences = (
 				FE0519072E2051EF0084240B /* XCRemoteSwiftPackageReference "swiftui-autocomplete-search" */,
-				16731E072E380B9500B3E2C9 /* XCLocalSwiftPackageReference "../../../ferrostar" */,
+				16731E072E380B9500B3E2C9 /* XCLocalSwiftPackageReference "../.." */,
```
ianthetechie, line 293: Ha! Will double check that this works but presumably the original path was a very roundabout way of going up too many levels and then descending once more 😂 Maybe that explains some of the Xcode finnicky-ness...

apple/Tests/FerrostarMapLibreUITests/Views/DynamicallyOrientingNavigationViewTests.swift
hactar: with "view no longer renders" you mean the Text view?

So for me, the tests were not passing on my computer. When I looked at the expected screenshots, they did not contain the yellow text box - yet the new snapshot did, like this:

download

Looking at the android logic, the road name is not displayed when empty there, so i used that logic in iOS instead of the "Unknown" label (https://github.com/hactar/ferrostar/blob/ios-sync-165-and-overlay/android/composeui/src/main/java/com/stadiamaps/ferrostar/composeui/config/NavigationViewComponentBuilder.kt#L63) - but this was just an assumption, if we need something else here I'll need more info :).

apple/Sources/FerrostarMapLibreUI/Extensions/MapViewCamera.swift
hactar: I didn't put too much thought into this, automotive had this type of initializer, so I applied the same pattern to the new types incoming from android (which as far as i can tell does not have this). I'm not familiar enough with the code yet to say which pattern is better for us here.

I have addressed your "drift" feedback by making the initializer use the default values too.

ianthetechie: Cool; makes sense. Yeah, that addresses my drift concern.

apple/Tests/FerrostarMapLibreUITests/Views/DynamicallyOrientingNavigationViewTests.swift
ianthetechie: Huh weird...

Yeah, it's test code so I don't have an issue with it differing between platforms. I've added it back so these render and we can get a sanity check on layout changes. Lemme know if this still causes issues for you developing locally.

Issue #754, opened 2025-12-16
Official maplibre-compose - Evaluate and upgrade


Archdoog:
Evaluate replacing [maplibre-compose-playground](https://github.com/Rallista/maplibre-compose-playground) with [maplibre-compose](https://github.com/maplibre/maplibre-compose) for our android target. If the migration looks promising and efficient, an MR is welcome.

### History

When ferrostar started, the only compose maplibre implementation was [Ramani Maps](https://github.com/ramani-maps/ramani-maps). It was relatively limited at that point, so we forked and created [maplibre-compose-playground](https://github.com/Rallista/maplibre-compose-playground) to add the camera behaviors (and some other features) required for ferrostar and the turn-by-turn navigation use case.

Since then, [maplibre-compose](https://github.com/maplibre/maplibre-compose) has become the official MapLibre compose library. It's got significantly more exposure, contribution, and should support most of our requirements.

### Why

- Ramani Maps is MPL-2.0. We'd prefer the more permissive BSD-3 of maplibre-compose.
- Performance issues. We've had several discussions on the ferrostar weekly call.
- Collaboration. Our Ramani fork has pretty limited usage, likely only ferrostar. As a result, we're missing out on a lot of collaboration and knowledge.
- None of the core ferrostar contributors here are android first by skill 😄

### Requirements

- [ ] All the basics: camera control, polyline, user puck.
- [ ] Android auto: https://github.com/Rallista/maplibre-compose-playground/pull/134. This isn't in ferrostar yet, but it's a short term goal.
- [ ] Demo app upgrade.
- [ ] Migration guide.

sargunv:
Hey, I'm the maintainer of maplibre-compose. Definitely happy to answer questions or take feedback and make changes. 

We're still developing the project; I'd say it's about 80-90% complete for Android and iOS, with most MapLibre functionality supported but some rough edges and small gaps. Desktop and web are far more experimental and incomplete right now. 

We're in a stage where we're still open to breaking changes to refine the API (and indeed each release makes at least some small ones) so it's a good time to try it out and make sure your use case is supported well.

PR #853, opened 2026-03-25
Migrate Android MapLibre Compose UI to the official SDK 

klemensz:
### Summary

This PR migrates `ui-maplibre` from the legacy Rallista MapLibre Compose integration to the official `org.maplibre.compose:maplibre-compose-android ` SDK. (#754)

The main scope is Android phone/tablet Compose navigation UI. Android Auto is not fully migrated here and should follow in a separate PR.

### Changes

- switch `ui-maplibre` to the official MapLibre Compose Android artifact
- replace the old `MapViewCamera` API with `NavigationMapState`
- add Ferrostar camera modes for follow, follow-with-bearing, overview, and free camera
- migrate route rendering to GeoJSON + `LineLayer`
- add configurable location puck styling
- add map tap and long-press callbacks
- update the demo app to the new API
- add unit tests for camera state, map state, GeoJSON generation, and location conversion

### Notes

- this PR is focused on Android phone/tablet Compose; `ui-maplibre-car-app` remains a compatibility path for now
- `NavigationMapView.onMapReadyCallback` has been removed in favor of `onMapLoadFinished` and `onMapLoadFailed`
  - this avoids the reflection-based workaround for accessing the native style and aligns the API with official `maplibre-compose` load callbacks
- Android Auto now uses the new map state/camera model, but the car host integration still depends on legacy Rallista components such as `ComposableScreen` and `SurfaceGestureCallback`-based host APIs
- Android Auto surface gestures are now forwarded without relying on Ramani internals
  - scroll, scale, and best-effort fling gestures are translated into camera/projection updates through `NavigationMapState` using the official `maplibre-compose` API, so behavior may still differ somewhat from the previous native/Ramani-based implementation
- the navigation puck shown during active navigation is a custom approximation of the old native GPS render mode
  - it is visually closer to the previous behavior, but not a 1:1 reproduction of the old `LocationComponent` `RenderMode.GPS`

### Testing

```
cd android
./gradlew :ui-maplibre:testDebugUnitTest :ui-maplibre-car-app:compileDebugKotlin :demo-app:testDebugUnitTest --no-daemon
```

### Screen recording

https://github.com/user-attachments/assets/3b38f609-266a-4d8d-8c7e-cb9cf1b9cea0


android/ui-maplibre/src/main/java/com/stadiamaps/ferrostar/maplibreui/NavigationMapView.kt
```diff
@@ -2,85 +2,194 @@ package com.stadiamaps.ferrostar.maplibreui
...
+      locationState = userLocationState,
+      enabled = navigationMapState.isTrackingUser,
+      trackBearing = navigationMapState.cameraMode == NavigationCameraMode.FOLLOW_USER_WITH_BEARING,
+  ) {
+    cameraState.position =
+        navigationMapState.trackingFollowingCameraPosition(
+            target = currentLocation.position,
+            bearing = currentLocation.bearing,
+        )
+  }
```
ianthetechie: Well... 😅 I guess this is why the video looked like a clock going tick tock advancing every second haha.

I assume this is a limitation inherent in the current MapLibre Compose API?

- Do they support animation (e.g. smooth the transition over 1 second)? That could work as a stop gap. Though humorously this would "solve" the periodic complaints of location lag if we hooked it up to a source that could generate new locations at 60Hz 😂 
- Ideally we could hook into the underlying MapLibre location engine (I think that's the type name) interface that alre ady exists and leverage the native camera modes (C++ level).

klemensz: The user location currently comes from `NavigationUiState.location`, which during active navigation is typically the snapped location from the trip state. That means the UI is not rendering the raw GPS position, but the map-matched/snapped position, and in practice this currently seems to update at roughly 1 Hz.

While navigating, the puck is rendered via a custom map overlay (`NavigationPuckOverlay`) rather than the built-in `maplibre-compose` location puck. I implemented it this way because the built-in puck is not flexible enough for the current navigation-style appearance (for example the arrow / bearing presentation). I am currently experimenting with interpolating / animating the overlay position itself, but that is not entirely straightforward.

For the camera, I am also trying to smooth tracking via `CameraState.animateTo()`.

Outside active navigation, Ferrostar uses the regular `maplibre-compose` location puck. I checked the `maplibre-compose` implementation as well, and it is structurally similar in that it also just renders the latest location value without doing any special native interpolation magic.

klemensz: I improved this further in the migration branch.

The key fix is that the navigation puck and tracking camera now use the same shared route-relative displayed location/bearing while on-route. Previously, the camera was following a smoothed route bearing while the puck still had its own extra bearing smoothing / raw bearing fallback, which caused visible mismatch and step-like rotation.

Now the only smoothing happens in the shared displayed navigation location:
- position is smoothed as forward progress along the route
- bearing comes from the route tangent

That makes the snapped on-route behavior noticeably more stable and keeps puck/camera rotation aligned.

I already tested it in real life while driving. When going off the planned route, the location puck stays on the old route for a short time, and then switches to the newly calculated route.

That said, I see this as more of a stopgap than the final solution as well. The current implementation is still a Compose-layer approximation using a custom puck layer plus `CameraState`, whereas the old Android path relies on native MapLibre tracking behavior with snapped locations as input.

So for the long-term solution, I think we would still want an Android-native integration path in `maplibre-compose` that lets us use the underlying native location/tracking behavior again, instead of rebuilding it manually at the Compose layer.

https://github.com/user-attachments/assets/75499f63-f3c4-4fae-b9ef-abe0b123183e

klemensz: Note: There is still a rendering issue with a "shaking" navigation arrow when zooming in more, but that's probably only solvable with some kind of access to the native map (which I would target as a follow-up PR).

ianthetechie: Yeah, that's a pretty good summary of the issues ! limitations in play. For the moment, I think you can assume that no consumer mobile device will generate GNSS updates faster than ~1Hz. It's just how all the major mobile OS's balance efficiency.

And yes, I agree this is more of a stop gap. We also feel that we should try to leverage the native-level features as much as possible, upstreaming where possible. I also agree with your assessment of it not being flexible enough as-is 😂 

Our medium-term solution is probably going to be combining a Kalman filter (or similar approximation of the user's location a very short time in the future) with the synchronous update parameter you mentioned. This way we can have an extremely fast updating symbol layer that matches the user's location more closely.

Question for you: do you think that layering one or more symbol layers in MapLibre to create the puck will be flexibile enough for all your needs? It certainly feels like the cleanest solution to me if we can get performance right, and sidesteps a lot of the current issues with the native implementation.

klemensz: I think layering one or more symbol layers should be flexible enough for our visual needs, yes.

For both two upcoming projects that I have in mind, the main question is probably not flexibility of appearance, but whether we can get the result stable and efficient enough for long-running navigation sessions.

My current view is:

- for appearance and customization, this approach should be fully sufficient
- for navigation robustness, the open questions are visual stability at high zoom / pitch and the runtime / battery cost of frequent source updates over longer sessions

So if synchronous source updates plus short-horizon prediction get the puck visually stable enough, this could very well be good enough for our needs in practice. I would still keep native tracked-location support in mind as a longer-term option, but I agree the layer-based puck feels like the cleanest medium-term solution if performance holds up.

ianthetechie:
I've just merged in main and downgraded Gradle to 9.2.1. We're caught in an unfortunate amount of turmoil with the snapshot testing library (Paparazzi) being somewhat in limbo before a big release and not quite having all the targets lined up.

klemensz:
> I've just merged in main and downgraded Gradle to 9.2.1.

OK, I think I upgraded it to 9.4.1 to be able to use AGP 9.1.1 because 9.0.1 was not compatible with my integration project. But I'll downgrade that now as well for now.

ianthetechie:
Ok; I've been playing around with this for a while in a real navigation scenario today and have a few bits of feedback. Most of them can be handled as separate issues.

At this point I've only found 2 that would block merging:

- The route overview seems to have incorrect padding. 
- Rotating the device puts the camera into a weird state. It's still tracking the user's location, but goes to north-up with no pitch instead of following the user's heading at an angle.

klemensz:
> * The route overview seems to have incorrect padding.

@ianthetechie Can you make this more specific or share a screenshot? Then I can look into it.

klemensz:
> * Rotating the device puts the camera into a weird state. It's still tracking the user's location, but goes to north-up with no pitch instead of following the user's heading at an angle.

I pushed a fix for this now. The problem was that NavigationCameraMode was not persisted across configuration changes.

ianthetechie:
https://github.com/user-attachments/assets/f6d15776-52d6-4dc0-a717-3d4526f76362

klemensz:
I tested both the main branch and this migration branch: From what I can see, the padding (or no padding) is the same on both branches when I run the demo app. So either 
- you can specify the expected padding behaviour in detail, or
- we leave this as a separate follow-up issue because it doesn't seem to be a regression

From my perspective it seems that there is padding missing in portrait mode on the left and right side of the screen - in my example where the route is more horizontally wide.

main branch:

https://github.com/user-attachments/assets/88fa0453-ee75-4b17-b652-d4b7f95b397b

PR branch:

https://github.com/user-attachments/assets/7ee981e5-9731-4600-b11a-7a6cc9d2a1dc

klemensz:
FYI

My PR "Implement synchronous updates for GeoJSON sources" (Android only currently) was merged into main at maplibre-compose: https://github.com/maplibre/maplibre-compose/pull/761

My other PR "Add app-provided native location tracking for Android and iOS" on the other hand was closed after some discussion, mainly because the route maplibre-compose wants to go is more away from MapLibre-Native Android/iOS (and eventually drop it at some point) towards implementing more of functionality in the library itself. https://github.com/maplibre/maplibre-compose/pull/763

ianthetechie (approved):
Thanks so much for your effort on this @klemensz! And it appears my testing did indeed surface old bugs, so yeah we can handle those in follow ups 😅

android/ui-maplibre/src/main/java/com/stadiamaps/ferrostar/maplibreui/NavigationMapView.kt
DestBro: > "For the camera, I am also trying to smooth tracking via CameraState.animateTo()."

 - it worked bad for me, I received jums, delays, each 3 sec. I implemented via another API.

PR #873, opened 2026-04-06
(WIP) fix: correction for car app

Archdoog:
This is just a running branch of revisions as I work through Google Play Review with Android Auto enabled.


Archdoog:
This comment is in progress/incomplete and will be updated with further testing.

So far only notes are:

1. The cluster view is showing the wrong template config. The distance is the full route distance, not the step distance.

Issue #875, opened 2026-04-10
0.44.0 deviation detection changes causing misbehaviour on loops and overlapping routes


freszu:
In our use case for ferrostar routes often end up overlapping and changes introduced in https://github.com/stadiamaps/ferrostar/pull/714 cause us a lot of problems and unexpected behaviors (the route deviation logic was further refactored between ferrostar .44~.49 - but the gist of it is the same - the steps can skip forward along the whole route).

## Repro
https://github.com/stadiamaps/ferrostar/compare/main...freszu:ferrostar:repro?expand=1
Lets say we have this route:
Our user leaves residential area (1) then he needs to complete 2,3,4 - but he takes a wrong turn - toward 3:
<img width="1334" height="682" alt="Image" src="https://github.com/user-attachments/assets/fa7b68e3-205f-4d7c-b4b2-c8643f7ac13d" />

Lets observe what will happen:
i've prepared a small repro based on demo app: https://github.com/stadiamaps/ferrostar/compare/main...freszu:ferrostar:repro?expand=1

### A) using default demo config NavigationControllerConfig

https://github.com/user-attachments/assets/ca99d5a9-3b80-45fc-9445-b6f90fe04742

User gets stuck instantly - he is not informed to take a uturn - why? because with new deviation logic - which takes whole route into account it never says user deviated - up until 0:50 - when user on his own decided to take u turn - but what happens afterward is even more problematic - he was able to U-turn - and wants to continue - but his position is stuck - i think i described here what might have happened in https://github.com/stadiamaps/ferrostar/issues/816

Edit: This variant has even worse edge case

<details>

When user starts with wrong heading - this can happen

<img width="1200" height="1920" alt="Image" src="https://github.com/user-attachments/assets/7a97a1be-c8b8-49df-a7f3-7561f72fdfab" />

</details>

## B) Our config - using `stepAdvanceDistanceFromStep`
line 24 in DemoConfigs 
In our app we use `stepAdvanceDistanceFromStep`
Lets run the same scenario:

https://github.com/user-attachments/assets/08ecae97-9412-4ad0-a421-422d3a61618c

User is not getting stuck - but also - he is never asked to U-turn - ferrostar happily skips forward - but waypoints are not completed - so when user deviates ~1:05 the route recalculation leads him once again to 1st point

## Same scenarios with Strict deviation behavior - pre 0.44:
[...]

PR #876, opened 2026-04-10
Feature/route deviation strict mode

freszu:
This is a followup to #875  read it first as it covers the problem well.
As explained in #875 - the whole behavior might require more thinking from maintainers - strict mode being true or false by default is one of such considerations.

<details>
<summary> Videos with strict mode true </summary>

Current demo step advance setup:

https://github.com/user-attachments/assets/9bb0cc05-a962-47d5-814c-0148fae794b3

Distance based step advance `stepAdvanceDistanceFromStep`:

https://github.com/user-attachments/assets/5b2f75a7-055a-43a2-88c7-1c95ae614652

</details>

### DIFF
--- a/android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoConfigs.kt
+++ b/android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoConfigs.kt
@@ -13,7 +13,7 @@ fun NavigationControllerConfig.Companion.demoConfig(): NavigationControllerConfi
       WaypointAdvanceMode.WaypointWithinRange(100.0),
       stepAdvanceDistanceEntryAndExit(30u, 5u, 32u),
       stepAdvanceDistanceToEndOfStep(10u, 32u),
-      RouteDeviationTracking.StaticThreshold(15U, 50.0),
+      RouteDeviationTracking.StaticThreshold(15U, 50.0, strictMode = false),
       CourseFiltering.SNAP_TO_ROUTE)
 }
 
--- a/apple/DemoApp/Demo/DemoModel.swift
+++ b/apple/DemoApp/Demo/DemoModel.swift
@@ -37,7 +37,7 @@ private extension FerrostarCore {
                 distance: 10,
                 minimumHorizontalAccuracy: 32
             ),
-            routeDeviationTracking: .staticThreshold(minimumHorizontalAccuracy: 15, maxAcceptableDeviation: 50),
+            routeDeviationTracking: .staticThreshold(minimumHorizontalAccuracy: 15, maxAcceptableDeviation: 50, strictMode: false),
             snappedLocationCourseFiltering: .snapToRoute
         )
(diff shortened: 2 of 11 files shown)


Archdoog:
Thanks for the PR!

Looking this and applying it conceptually to the deviation + step advance process, I think the right solution would be to add a 3rd case to our deviation state:

```rust
/// Status information that describes whether the user is proceeding according to the route or not.
///
/// Note that the name is intentionally a bit generic to allow for expansion of other states.
/// For example, we could conceivably add a "wrong way" status in the future.
#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
#[cfg_attr(feature = "wasm-bindgen", tsify(into_wasm_abi, from_wasm_abi))]
pub enum RouteDeviation {
    /// The user is proceeding on course within the expected tolerances; everything is normal.
    NoDeviation,
    /// The user is off the step, but still on the route polyline.
    OffStep {
        /// The deviation from the step line, in meters.
        deviation_from_step_line: f64,
    },
    /// The user is off the expected route.
    #[cfg_attr(feature = "wasm-bindgen", serde(rename_all = "camelCase"))]
    OffRoute {
        /// The deviation from the route line, in meters.
        deviation_from_route_line: f64,
    },
}
```

Once that's done, we'd be able to:

1. Differentiate within each step advance condition what's acceptable for advance. They could have configurations or we could wrap them for multiple types. (think `SomeAdvanceCondition` + `RejoinRouteSomeAdvanceCondition`).
2. Specific step advance modes could be configured to allow some fashion of your strict_mode. Named according to how the step advance mode's conditional behavior/how it would say I can advance.

If your core condition is distance from step, you very well might want it to be bound to OffStep, not OffRoute as it's `deviation_condition`.

Thoughts?

freszu:
It sounds like a smarter approach! - i was trying to think of edge cases described in #875 and it sounds like with proper step/waypoint advance conditions distances set they would be all covered - we could still use `WaypointWithinRange` for waypoint completion to force users visits on waypoints unless explicitly dropped (#875 video B timemark 1:00).

freszu:
@Archdoog i gave it a shot see https://github.com/stadiamaps/ferrostar/pull/879

i am keeping it as a draft to provide better description and videos but with quick tests

stepAdvanceDistanceFromStep - with new flag set to true - allows skipping forward - keeping behavior from ferro >=.44 - when false it behaves like <.44

and for 
> Current demo step advance setup:

Its also not getting stuck like in #875 and continues recalculating new route to do U-turn like on video presented in this PR.

Let me know WDYT, if i should push this further etc., does that fit ferrostar vision and changes that were made in .44 For now we will have to keep a fork on our side as we need access to Valhalla waypoint U-turn property - that was introduced in .48

ianthetechie:
Superseded by #879

PR #878, opened 2026-04-14
Freshen iOS CI targets

ianthetechie:
This should fix the random CI failures we're seeing re: the iOS Simulator target. I'm just jumping all the way to ~current so we don't have to do this for a while again.


apple/Tests/FerrostarSwiftUITests/Views/DynamicIslandViewTests.swift
```diff
@@ -7,6 +7,14 @@ import XCTest
 @testable import FerrostarSwiftUI
 
 final class DynamicIslandViewTests: XCTestCase {
+    /// A locale-pinned distance formatter for deterministic snapshot output.
+    private let testDistanceFormatter: MKDistanceFormatter = {
+        let formatter = MKDistanceFormatter()
+        formatter.locale = Locale(identifier: "en_US")
+        formatter.units = .imperial
+        return formatter
+    }()
```
ianthetechie, line 16: This test was nondeterministic before 😂

apple/Sources/FerrostarSwiftUI/DynamicIsland/Views/LiveActivityView.swift
```diff
@@ -2,11 +2,20 @@ import MapKit
...
-
+    let theme: InstructionRowTheme
+    let distanceFormatter: Formatter
     let state: TripActivityAttributes.ContentState
 
+    init(
+        theme: InstructionRowTheme = DefaultInstructionRowTheme(),
+        distanceFormatter: Formatter = MKDistanceFormatter(),
+        state: TripActivityAttributes.ContentState
+    ) {
```
ianthetechie, line 13: This should be fine from a public UI perspective (it's non-breaking) but was required to make the UI snapshot tests deterministic.

Archdoog (approved):

PR #882, opened 2026-04-24
feat: web replay tool

ppenicka:
Available at /tools/replay/, allows to replay recorded navigation from a JSON file while providing a list of steps and events in a sidebar.


ianthetechie (approved):
Just a few minor tweaks; looks good after I verify CI passes form my own changes to clean up the tsc warnings :D 

(There is a bit of internal cleanup to do here too at the binding generation layer eventually.)

PR #884, opened 2026-04-28
Android demo app improvements (bottom sheet, recenter button, etc.)

klemensz:
## Summary

This follow-up improves the Android demo app UX on top of the ongoing `maplibre-compose` migration in [#853](https://github.com/stadiamaps/ferrostar/pull/853).

Instead of starting navigation immediately from a long press or search result, the demo now uses an explicit destination-selection flow with a preview bottom sheet, camera preview, and a browse-mode recenter button.

## Why

This brings the demo app closer to a real navigation app and exercises more of the `maplibre-compose` integration in realistic user flows. It is split into a separate PR so the original migration PR does not grow even larger.

## Changes

- add destination selection state to the demo scene / view model
- show a bottom sheet before starting navigation
- preview the selected destination with camera updates and adjusted bottom padding
- add a recenter button while not navigating
- keep the dropped-pin overlay, now backed by typed `FeatureCollection` data instead of raw JSON strings
- include a small `MapControls` padding fix for top system insets

## Notes

This is primarily demo-app UX work. The only `ui-maplibre` change is the small `MapControls` inset adjustment.

## Screenshots

image
image


android/demo-app/build.gradle
```diff
@@ -63,6 +63,7 @@ dependencies {
     implementation libs.androidx.compose.ui.graphics
     implementation libs.androidx.compose.ui.tooling
     implementation libs.androidx.compose.material3
+    implementation libs.androidx.compose.material.icon.extended
```
ianthetechie, line 66: We can probably should replace this with SVGs (it's not a merge blocker for the demo app but does increase build times).

ianthetechie (approved):

android/demo-app/build.gradle
Archdoog: https://github.com/stadiamaps/ferrostar/issues/907

PR #885, opened 2026-04-28
Android: Use dynamic screen-based padding again for navigation camera 

klemensz:
During the maplibre-compose migration (#853) relative values (0.5) for padding were replaced with hard coded Dp value. I changed this to relative values again.

It could be observed in these screen captures for example, that the 3D navigation puck is too far away from the bottom elements (before and after migration).

Bildschirmfoto 2026-04-29 um 10 26 37

## Changes

Updates the Android MapLibre navigation camera padding to use dynamic, screen-based values again instead of fixed `Dp` offsets introduced during the maplibre-compose migration.

- Calculates navigation camera padding from the current screen width/height.
- Restores landscape behavior by shifting the camera target based on screen width.
- Adjusts bottom camera padding while navigating so the target remains visible above bottom UI insets.
- Applies the effective camera options across portrait, landscape, and dynamically orienting navigation views.


Archdoog (approved):
Looks good to me. Will let @ianthetechie decide how he wants to get all this merged 👍

klemensz:
Conflicts are resolved now.

ianthetechie:
Looks good; thanks! Probably worth a follow-up at some point to automatically take the navigation use case into account (e.g. bike vs pedestrian vs car).

We have a somewhat badly named issue #196 that hints at a broader project of optimizing the display per use.

PR #886, opened 2026-04-28
Suppress visual + spoken instructions while off-route

devnull133:
Hey! Small fix here that I ran into in a navigation setup where automatic recalculation is intentionally disabled. The bug exists regardless of that setting, but it's easiest to spot with rerouting disabled.
                                                                                                                                                                                                                                                     
## What's wrong                                                 

  Visual and spoken instructions are computed from the snapped distance to the next maneuver — i.e. from the user's projection onto the route polyline, not from the user's actual position. While the user is on-route, those two are close enough  
  that the countdown is meaningful. While the user is **off-route**, the snap is geometrically unsound: a user laterally far from the route still projects to *some* point along it, and that phantom projection's "distance to next maneuver" ticks
  down as if they were following the route.                                                                                                                                                                                                          
                                                                  
  The result is that Ferrostar happily emits `visual_instruction` / `spoken_instruction` updates from a phantom snap. A consumer that wires `SpokenInstructionObserver` to a TTS engine ends up announcing "in 100 m, turn left onto Main Street"    
  while the user is 200 m parallel to (and walking *away* from) Main Street — misleading at best.
                                                                                                                                                                                                                                                     
  ## The fix                                                      

  Two emission sites in `navigation_controller/mod.rs` now gate on deviation: when `deviation != NoDeviation`, both `visual_instruction` and `spoken_instruction` are set to `None` instead of being computed from the unsound snap.                 
   
  - `create_intermediate_trip_state` — the `update_user_location` path.                                                                                                                                                                              
  - `get_initial_state` — covers the case where the user opens the app already off-route.
                                                                                                                                                                                                                                                     
If maintainers feel this should be opt-out-able rather than hardcoded — e.g. for vehicle-nav scenarios where brief deviation flips during GPS dropouts are mostly false positives and choking instructions during them is undesirable — happy to add a suppress_instructions_while_off_route: bool flag to NavigationControllerConfig. Left it hardcoded in this PR because the snapped instruction is genuinely unsound while off-route, but I'm open to either direction.
                                                                                                                                                                                                                                                     
  ## How to reproduce                                             

  Easiest way to see it without writing any test scaffolding:                                                                                                                                                                                        
   
  1. Run your usual sample app.                                                                                                                                                                                                                      
  2. Pick any route (the bug isn't route-specific — any multi-step route with voice instructions will surface it).
  3. Disable automatic recalculation, or just deviate fast enough that recalculation hasn't kicked in yet.
  4. Use a simulated location that moves **parallel** to the route, ~100–200 m laterally offset, in roughly the route's direction.                                                                                                                   
  5. With voice instructions enabled, observe that maneuver announcements continue to fire as the parallel travel progresses, despite the deviation indicator being on. The countdown ticks down off the snapped (phantom) position.                 
                                                                                                                                                                                                                                                     
  After this PR, the off-route alert still fires (deviation handling is unchanged), but the maneuver instructions go silent until the user is back on-route.                                                                                         
                                                                                                                                                                                                                                                     
  ## Tests                                                                                                                                                                                                                                           
                                                                  
  Two integration tests in the existing `mod.rs` test module, using the `Valhalla` fixture:

  - `test_off_route_suppresses_instructions` drives the controller through an on-route → off-route → back-on-route sequence and asserts instructions go `Some → None → Some`, with the recovered values matching what was emitted before deviation.  
  - `test_off_route_initial_state_suppresses_instructions` covers the `get_initial_state` branch independently.
                                                                                                                                                                                                                                                     
  When the suppression is reverted, the test fails with concrete diagnostic output (literally "Turn left onto the walkway." surfaced from a phantom snap at a location ~55 km off the route).

### DIFF
--- a/common/ferrostar/src/navigation_controller/mod.rs
+++ b/common/ferrostar/src/navigation_controller/mod.rs
@@ -169,6 +169,16 @@ impl Navigator for NavigationController {
             ..
         } = initial_trip_state
         {
+            // If the user starts off-route, suppress instructions for the same reason as
+            // in `create_intermediate_trip_state`: the snap-derived distance to the next
+            // maneuver is geometrically unsound, so any countdown surfaced from it would
+            // mislead the user.
+            let (visual_instruction, spoken_instruction) =
+                if deviation == RouteDeviation::NoDeviation {
+                    (visual_instruction, spoken_instruction)
+                } else {
+                    (None, None)
+                };
             TripState::Navigating {
                 current_step_geometry_index,
                 user_location,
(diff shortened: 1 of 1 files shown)


ianthetechie (approved):
Solid; thanks for the detailed writeup, comments, and tests!

PR #887, opened 2026-05-01
Reset step advance condition state on every step advance

devnull133:
Hi, I noticed this while working on something else. I'm not sure how often the manual step advance is used, but it feels like it should be fixed anyway.

`NavigationController::advance_to_next_step` carried the existing step advance condition forward unchanged. Auto-advance was unaffected because `should_advance_step` already returns a fresh condition via `advance_to_new_instance`, but manual advance - `Navigator::advance_to_next_step` bypasses that path entirely. Stateful conditions (the entry-latch family, composites containing them, custom user implementations) leaked their per-step state into the next step.


Archdoog (approved):
Good catch. This looks good as it, thanks 👍

A note/question. Any thoughts on any further cleanup? 

E.g. we could extract new instance out to the NavigationController so that anytime we "recycle" a continuing or advanced condition it's clear at the call site. The way I built this, where the `new_instance` was hidden inside the conditional logic on _some_ conditions is definitely confusing looking from the top down. Even though it made sense from the state management process inside each one.

devnull133:
Sorry, I missed this comment. To be honest, I haven't taken a closer look at the overall architecture here, so I'd leave this one up to you.

PR #905, opened 2026-05-21
feat: add optional language parameter to Valhalla provider

NAKKA-POOJITHA:
"Adds an optional language field to the Valhalla request payload."

### DIFF
--- a/common/ferrostar/src/routing_adapters/valhalla.rs
+++ b/common/ferrostar/src/routing_adapters/valhalla.rs
@@ -302,7 +302,7 @@ pub struct ValhallaHttpRequestGenerator {
     endpoint_url: String,
     /// The Valhalla costing model to use.
     profile: String,
-    pub language: Option<String>,
+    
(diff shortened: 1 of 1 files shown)


ianthetechie (commented):
I think that a better direction for this would be an options builder which makes the details more discoverable. Unfortunately this is probably quite a lot of work (there are hundreds of options), so we'll probably want to always keep a way to add arbitrary keys like we have today.

I'm guessing you want this more immediately for use in your app, and the good news is that you can already pass options via an arbitrary map :) See the docstring a few lines above for an example. At the top level you can just pass `{"language": "de"}` etc.... a new struct field isn't necessary.

NAKKA-POOJITHA:
I've updated the PR to remove the top-level struct field. Instead, I added a chainable with_language convenience method that injects the key straight into the existing options map behind the scenes. This keeps the architecture clean while making the feature easily discoverable. Let me know if this looks good to merge!

ianthetechie (commented):
Thanks!

PR #924, opened 2026-06-04
iOS: Namespace includes to avoid polluting global name space

### LINKED ISSUE #912: iOS: ferrostar incompatible with similarly architected packages
I liked the rust-core + uniffi approach you used with ferrostar so much, that I wanted to copy your layout for another library I'm using in tandem with ferrostar.

Doing so produces this error:

```
Multiple commands produce '/Users/mkirk/Library/Developer/Xcode/DerivedData/DoubleUniffiDemo-fvpuxzaqemypbrarfclclmwlzqlm/Build/Products/Debug-iphonesimulator/include/module.modulemap'
```

It seems like there is some kind of global namespace ferrostar has claimed, which prevents other packages from working in this way.

I've got a repro, along with some more notes here:
https://github.com/michaelkirk/double_uniffi_error_demo

It includes a HelloWorld library that follows the rust-core + uniffi architecture to induce the error. Let me know if you have any ideas.

michaelkirk:
Fixes #912 

This is ready for review, but marked a draft because it includes a commit to use local frameworks, which is useful for testing, but should probably not be merged.

Note: There are actually two alternative proposals - one which requires renaming a build target. If that's too disruptive, take a look at my earlier (reverted) commit which solves the problem without renaming the target, but note that it relies on a yet-to-be-released version of uniffi.

### DIFF
--- a/common/build-ios.sh
+++ b/common/build-ios.sh
@@ -46,19 +46,20 @@ generate_ffi() {
     if [[ "$(uname)" == "Darwin" ]]; then
       lib_ext="dylib"
     fi
-    cargo run -p uniffi-bindgen-swift -- target/release/lib$1.$lib_ext target/uniffi-xcframework-staging --swift-sources --headers --modulemap --module-name $1FFI --modulemap-filename module.modulemap
+    cargo run -p uniffi-bindgen-swift -- target/release/lib$1.$lib_ext target/uniffi-xcframework-staging/$1FFI --swift-sources --headers --modulemap --module-name $1FFI --modulemap-filename module.modulemap
   else
     echo "Using iOS library for FFI generation"
     # NOTE: Convention requires the modulemap be named module.modulemap
-    cargo run -p uniffi-bindgen-swift -- target/aarch64-apple-ios/release/lib$1.a target/uniffi-xcframework-staging --swift-sources --headers --modulemap --module-name $1FFI --modulemap-filename module.modulemap
+    cargo run -p uniffi-bindgen-swift -- target/aarch64-apple-ios/release/lib$1.a target/uniffi-xcframework-staging/$1FFI --swift-sources --headers --modulemap --module-name $1FFI --modulemap-filename module.modulemap
   fi
   
   mkdir -p ../apple/Sources/UniFFI/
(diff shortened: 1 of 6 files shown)


Archdoog:
Beat me to it! 

I have another uniffi repo in my project and found this out as well.

One bonus, we should also remove the public library export of FerrostarCoreFFI from the library registration in Package.swift. We can do this by using `@_exported import FerrostarCoreFFI` inside `FerrostarCore`. Meaning the dev only has to `import FerrostarCore` to get both the code in there and the generated models. Because FerrostarCore will export the models through its own target.

michaelkirk:
Yes I'm using that re-export feature in my own framework. 

Note that the syntax is considered unstable. I'm not sure what the migration path will be if the syntax changes, but presumably it *could* break at any time with a future swift update.

common/build-ios.sh
```diff
@@ -46,19 +46,20 @@ generate_ffi() {
...
-    cargo run -p uniffi-bindgen-swift -- target/aarch64-apple-ios/release/lib$1.a target/uniffi-xcframework-staging --swift-sources --headers --modulemap --module-name $1FFI --modulemap-filename module.modulemap
+    cargo run -p uniffi-bindgen-swift -- target/aarch64-apple-ios/release/lib$1.a target/uniffi-xcframework-staging/$1FFI --swift-sources --headers --modulemap --module-name $1FFI --modulemap-filename module.modulemap
   fi
   
   mkdir -p ../apple/Sources/UniFFI/
-  mv target/uniffi-xcframework-staging/*.swift ../apple/Sources/UniFFI/
+  mv target/uniffi-xcframework-staging/$1FFI/*.swift ../apple/Sources/UniFFI/
   
   # Only move modulemap if not in ffi-only mode
+  # TO REVIEW: I've updated this code to be in sync with other changes, but it seems like it is and was a no-op. Is there some reason to mv a file to itself?
```
ianthetechie: Huh that's weird... clearly a mistake on my part 😅 Let's just remove this then.

ianthetechie:
This looks good to me @michaelkirk. Anything else you wanted to clean up since it's still a draft?

michaelkirk:
All good from my perspective. Thanks!

Issue #347, opened 2024-11-05
Submit to the Swift Package Index


ianthetechie:
This has a few sub-tasks built into it.

- [ ] Set up DocC (ideally bringing in the iOS getting started guide; maybe via CI script)
- [ ] Actually submit to SPI
- [ ] Point sections of the docs to SPI hosted reference when appropriate

ianthetechie:
This is a good first issue for anyone looking to get into the iOS code. At least the third item re: updating the main guide with links to the SPI hosted docs.

Some pointers on getting started with DocC:

* Official docs on making a landing page: https://www.swift.org/documentation/docc/adding-structure-to-your-documentation-pages
* An extremely minimal project with a DocC landing page: https://github.com/stadiamaps/mapkit-caching-tile-overlay/tree/main/Sources/CachingMapKitTileOverlay. Ideally we'd figure out how to make this mirror the iOS setup docs from [the guide](https://github.com/stadiamaps/ferrostar/blob/main/guide/src/ios-getting-started.md). Both use Markdown so this should be scriptable.

PR #928, opened 2026-06-16
docs: add DocC landing page for FerrostarCore

NAKKA-POOJITHA:
This PR addresses part of #347 by introducing a dedicated DocC documentation landing page for the `FerrostarCore` framework target. 

### Changes
- Added `FerrostarCore.docc` directory structure under `apple/Sources/FerrostarCore/`.
- Created `FerrostarCore.md` providing a high-level architectural overview, core concepts, and links to the user guide using proper DocC symbol linking.


ianthetechie:
Thanks; this is a good start!

From the original issue:

> Ideally we'd figure out how to make this mirror the iOS setup docs from [the guide](https://github.com/stadiamaps/ferrostar/blob/main/guide/src/ios-getting-started.md). Both use Markdown so this should be scriptable.

I think that we do still need to do this. Though in retrospect maybe my original idea of using a script wasn't the best, since mdbook is more vanilla markdown formatting, and DooC has its own special way of doing links. So, I think we should probably rather try to _move_ the iOS docs out of mdbook and over to here, and adapt the formatting / page structure as needed to Apple/DooC conventions.

NAKKA-POOJITHA:
I've moved the four iOS documentation files (ios-getting-started.md, swiftui-customization.md, ios-dynamic-island.md, and ios-carplay.md) directly out of the guide directory and into the FerrostarCore.docc bundle folder. I also set up FerrostarCore.md to act as the primary DocC landing page mapping to them using proper symbol/page links.

ianthetechie:
Thanks! It looks like you didn't make any changes to the Guide 😅

There should still be an iOS entry in the menu, but it can be either a direct link (if mdbook supports that; I don't actually know offhand) or a simple placeholder page that says you can find the documentation on the Swift Package Index. For now, the link can go here: https://swiftpackageindex.com/stadiamaps/ferrostar. (There is a docs link but it's broken since we don't have this yet haha).

PR #931, opened 2026-06-23
Android: Enable synchronous updates for NavigationPuckOverlay GeoJSON source

klemensz:
## Summary

Enable synchronous GeoJSON updates for the Android `NavigationPuckOverlay` source.

The navigation puck source contains a single point feature with the current position and bearing. This is a small, frequently updated, latency-sensitive source, so it is a good fit for MapLibre Compose's `GeoJsonOptions(synchronousUpdate = true)` (see https://github.com/maplibre/maplibre-compose/pull/761 and https://github.com/maplibre/maplibre-native/pull/3968).

## Motivation

The navigation puck should visually track position and bearing updates with as little latency as possible during active navigation. Applying this update synchronously avoids the extra async source update delay for this very small in-memory GeoJSON source.

## Notes

This is intentionally limited to the navigation puck source. Larger GeoJSON sources such as routes, shape collections, or clustered data should continue using the default asynchronous update path to avoid hurting frame performance.

## Testing

- Tested demo app with simulated location
  - I tested with both values (true and false) and also checked the profile - could't observe increased CPU usage


ianthetechie:
Thanks!

PR #942, opened 2026-07-12
Dismiss Live Activity on Terminate

hactar:
When you cancel a navigation by hitting the X button in ferrostar, the live activity can "linger", continuing to show instructions. Using "immediate" this is prevented - which makes sense as showing a "turn left" or similar for a few minutes makes little sense.

IMG_5907


Archdoog (approved):

ianthetechie (approved):
Thanks!

PR #943, opened 2026-07-14
Fix session recording format drift

ianthetechie:
A user reported an issue with replaying a session recording from mobile on the web. It turns out that the JavaScript side was dumping in camelCase but native was doing snake_case. And as a complicating factor, native used a different timestamp format. I'm not sure how we missed that one...

This PR fixes that, and then some:
* Standardizes on camelCase (gotta pick something, and this ended up being a larger _snapshot_ diff but smaller _code_ diff).
* Accepts both casings as a temporary measure. I plan to drop this as part of the 1.0 release.
* Adds our first real web tests!
* Surfaces errors rather than panicking (this was just something we missed in the post-GSoC cleanup). The APIs now return results, and the web debugger displays the actual error clearly (it was just an incomprehensible wasm crash prior).

Pasted Graphic


Archdoog (approved):
Let's go over the snapshot changes over the call. Generally speaking though, looks fine 👍

louim:
I also hit this while testing the replay and had to temporarily work around it by rejigging the replay json to match the camelCase the web replay tool expected. It's nice to see this fixed!

ianthetechie:
Sorry for the delay on this. I was trying to get this even cleaner since the serialization is a bit inconsistent in casing at the moment, but that turned out to be a lot more work than I bargained for 😓 Going to just merge this as-is for now to unblock everyone.

PR #944, opened 2026-07-22
Add name to Valhalla waypoint properties

freszu:
## Motivation

The Valhalla locations API supports a `name` property ([API reference](https://valhalla.github.io/valhalla/api/turn-by-turn/api-reference/#locations)) — a location/business name that may be used in route narration, e.g. *"You have arrived at \<business name\>."* `ValhallaWaypointProperties` doesn't currently expose it, so there's no way to name waypoints when building requests through Ferrostar. 

## Human gist 😉 
This was ai asisted - but i realized some problems along the way:

I need to expose name property in valhlalla to be able to achieve instructions like:
"You have arrived at your final destination"
Example from demo:

<details>
<summary> DIFF to enable in demo </summary>

```
diff --git a/android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoNavigationViewModel.kt b/android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoNavigationViewModel.kt
index f32405b1..6789b479 100644
--- a/android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoNavigationViewModel.kt
+++ b/android/demo-app/src/main/java/com/stadiamaps/ferrostar/DemoNavigationViewModel.kt
@@ -25,8 +25,10 @@ import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.launch
 import uniffi.ferrostar.GeographicCoordinate
 import uniffi.ferrostar.UserLocation
+import uniffi.ferrostar.ValhallaWaypointProperties
 import uniffi.ferrostar.Waypoint
 import uniffi.ferrostar.WaypointKind
+import uniffi.ferrostar.createWaypointWithValhallaProperties
 
 data class DestinationSelection(
     val coordinate: GeographicCoordinate,
@@ -196,7 +198,11 @@ class DemoNavigationViewModel(
           ferrostarCore.getRoutes(
               lastLocation,
               listOf(
-                  Waypoint(coordinate = destination, kind = WaypointKind.BREAK),
+                  createWaypointWithValhallaProperties(
+                      coordinate = destination,
+                      kind = WaypointKind.BREAK,
+                      properties = ValhallaWaypointProperties(name = "Final destination")
+                  ),
               ),
           )
 

```

</details>

Result:

https://github.com/user-attachments/assets/eed72442-490e-439c-b0e3-00ec4f5e15ea

Problems:
1. due to how ValhallaWaypointProperties work and how valhalla responds back there was initially a clash of name and any reroute would end up overwriting waypoint name - with OSRM name - which is street name - **proposed solution** - to avoid this clash -  #[serde(rename = "waypoint_name")] was introduced
2. due to documented behavior of valhalla not responding back with passed name (see properties docs) & 1 - **i kinda follow this behavior and accept the fact this param can get lost** - also making sure it doesn't get replaced with OSRM name from 1. In production app this is solved through `RouteDeviationHandler` which makes sure to inject expected waypoints - this workarround in our app was actually introduced earlier for different reason - ferrostar currently looses information about e.g. stop position (on the left / on the right) due to how it uses waypoints from route which are already snapped.

I tried documenting these tricky problems

## Changes

- Adds `name: Option<String>` to `ValhallaWaypointProperties`, serialized into the `locations` entry of the generated request as `name`.
- **`ValhallaWaypointProperties` loses `Copy`** (a `String` field can't be `Copy`) — technically a breaking change for Rust API consumers relying on implicit copies; `Clone` remains. The Kotlin/Swift/TS bindings are unaffected.
- The field is stored as `waypoint_name` inside the internal waypoint `properties` blob (via `#[serde(rename)]`; the request still sends `name`). This sidesteps a subtle collision: waypoints rebuilt from a parsed route response carry `OsrmWaypointProperties` in the same blob, where `name` is the *snapped street name*. When a deviation handler feeds `TripState`'s remaining waypoints back into a new route request (the default reroute path in the platform cores), that street name would deserialize into the Valhalla `name` and be sent to the server, polluting narration. With the rename, reroutes simply drop the property — consistent with how all other Valhalla waypoint properties already behave (they aren't echoed back in OSRM format either). A regression test covers this.

🤖 Assisted by [Claude Code](https://claude.com/claude-code)


common/ferrostar/src/routing_adapters/valhalla.rs
```diff
@@ -113,6 +113,22 @@ pub struct ValhallaWaypointProperties {
     /// into a Valhalla `break_through`, and a [`WaypointKind::Via`] into a `through`.
     #[cfg_attr(feature = "uniffi", uniffi(default))]
     pub allow_uturns: Option<bool>,
+    /// Location or business name.
+    ///
+    /// The name may be used in the route narration directions,
+    /// such as "You have arrived at &lt;business name&gt;."
+    ///
+    /// WARNING! this param is not echoed back in the response - so its value might get lost on
+    /// reroute
```
ianthetechie:
```suggestion
    /// WARNING: this param is not echoed back in the response, so this can disappear when rerouting!
```

oof good catch. I'd say this is probably an unintentional bug in Valhalla. The `name` is preserved in Valhalla JSON format responses, but not OSRM-format ones.

ianthetechie (approved):
Thanks for this PR! I'll get the semver warning resolved and merge.

PR #945, opened 2026-07-23
Android: keep direction_* drawables from resource shrinking

freszu:
Note! it looks like it starts happening when switching to AGP 9
before there was a heuristic checking for that - with AGP 9 and default settings this looks to be gone

[ManeuverIcon](https://github.com/stadiamaps/ferrostar/blob/main/android/ui-shared/src/main/java/com/stadiamaps/ferrostar/ui/shared/icons/ManeuverIcon.kt) looks up the `direction_*` drawables by name via `Resources.getIdentifier`, so the AGP resource shrinker sees no static reference to them and strips all 95 icons from any consumer app built with `isShrinkResources = true`. Because `ManeuverImage` deliberately ignores resolution failures, the symptom is maneuver icons silently missing in release builds only.

This ships a `tools:keep` rule inside the `ui-shared` AAR. Keep files in library resources are merged into the consuming app and honored by the resource shrinker, so consumers get the fix transparently.

The file is named `ferrostar_ui_shared_keep.xml` rather than the conventional `keep.xml` to avoid a resource-merge collision with a consumer's (or another library's) own `res/raw/keep.xml`; the shrinker scans all raw XML resources for `tools:keep` attributes regardless of file name.

Verified in a consumer app (AGP 9.x, `isMinifyEnabled = true`, `isShrinkResources = true`): before, `build/outputs/mapping/<variant>/resources.txt` listed every `drawable:direction_*` as removed; with the rule merged, all 95 report `reachable from keep xml file` and the icons render.


freszu:
format lint will fail - its broken on main - in #944 i've fixed it

won't be touching this in that PR - its pure android change

Archdoog (approved):

PR #958, opened 2026-09-03
Add web distance system with aligned km/mi decimals

eminmapme:


eminmapme:
@ianthetechie This is a continuation of [#661](https://github.com/stadiamaps/ferrostar/pull/661). Same web distance-system work, rebased onto current main, with conflicts resolved and km/mi decimal cutoffs aligned to the formatter so imperial no longer shows "0 mi" / a long "1 mi". #661 can be closed in favor of this PR.

ianthetechie:
Thanks! I'm going to switch the default just so that it's more familiar to most of the world, but this is great! I just tried a few scenarios and it seems to work well.

PR #960, opened 2026-09-09
feat(apple): make SpokenInstructionObserver's audio session management opt-out

meme-te:
## Problem

`SpokenInstructionObserver` always takes over the shared `AVAudioSession` around each spoken
instruction. That is the right default for apps using the built-in `AVSpeechSynthesizer`, but it
permanently strands other apps' audio for apps that inject a custom `SpeechSynthesizer` playing
through their own session.

The mechanism:

* `requestAudioFocus()` sets `.duckOthers` + `.interruptSpokenAudioAndMixWithOthers` and
  `mode = .voicePrompt`, then activates the session.
* `releaseAudioFocus()` clears `hasAudioFocus` only **after** `setActive(false)` succeeds.
* `setActive(false)` fails while the session still has active audio I/O.

An app that keeps a microphone tap open (wake-word standby) or plays its own guidance audio
therefore never releases focus: `hasAudioFocus` stays `true`, every later `requestAudioFocus()`
early-returns on `guard !hasAudioFocus`, and other apps stay ducked for the rest of the session.

We measured this on a device with a Bluetooth headset: background music dropped from `0.80` to
`0.30` system output volume about one second after navigation started, and never recovered.

Recovering from the host app does not work either. Re-applying `setCategory` on an already-active
session interrupts other apps' playback, so "take the session back" trades a stuck duck for
repeated audio dropouts (we tried it, and it was worse).

There is also no way to opt out today: `FerrostarCore` takes `spokenInstructionObserver` as a
non-optional concrete `SpokenInstructionObserver`, and the class is `public` but not `open`, so it
can be neither replaced nor subclassed from another module.

## Change

Adds `managesAudioSession` to `SpokenInstructionObserver`, defaulting to `true` so **existing
behavior is unchanged**. Apps that own their audio session can pass `false`:

```swift
SpokenInstructionObserver(synthesizer: mySynthesizer, isMuted: false, managesAudioSession: false)
```

One file, guarding the three `audioManager` calls. No change to instruction de-duplication or any
other logic.

## Testing

Built and run on device (iPhone, Bluetooth A2DP headset) with `managesAudioSession: false`:
background music keeps playing across app launch and repeated microphone on/off cycles, and the
app's own spoken guidance still plays through its session.


apple/Sources/FerrostarCore/Speech/SpokenInstructionObserver.swift
```diff
@@ -14,17 +14,43 @@ public class SpokenInstructionObserver {
...
+    /// that inject a custom ``SpeechSynthesizer`` which plays audio through the app's own session:
+    ///
+    /// * `requestAudioFocus()` sets `.duckOthers` and `.voicePrompt` on the shared session.
+    /// * `releaseAudioFocus()` only clears `hasAudioFocus` **after** `setActive(false)` succeeds,
+    ///   and `setActive(false)` fails while the session still has active audio I/O.
+    ///
+    /// An app that keeps a microphone tap open (e.g. for wake-word standby) or plays its own
+    /// audio therefore never releases focus, so other apps stay ducked for the rest of the
+    /// session. Recovering by re-applying `setCategory` is not a workable fix either: changing
+    /// the category of an already-active session interrupts other apps' playback.
```
ianthetechie:
```suggestion
    /// Whether this observer should take over the app's `AVAudioSession` while speaking.
    ///
    /// When true, automatically manages audio focus (ducking other apps)
    /// before speaking, and releases after each instruction.
    ///
    /// Setting it to `false` means the application will manage this itself.
    /// This matters for some apps that inject a custom ``SpeechSynthesizer``
    /// which plays audio through the app's own session.
```

Cleaning this up to be more concise. LLM internal hallucinations about fixes that don't work aren't really relevant for framework users or maintainers.

apple/Sources/FerrostarCore/Speech/SpokenInstructionObserver.swift
```diff
@@ -14,17 +14,43 @@ public class SpokenInstructionObserver {
...
     ///
     /// - Parameters:
     ///   - synthesizer: The speech synthesizer.
     ///   - isMuted: Whether the speech synthesizer is currently muted. Assume false if unknown.
+    /// - Parameters:
+    ///   - synthesizer: The speech synthesizer.
+    ///   - isMuted: Whether the speech synthesizer is currently muted. Assume false if unknown.
+    ///   - managesAudioSession: Whether this observer may configure the shared `AVAudioSession`
+    ///     while speaking. Defaults to `true` (existing behavior). Pass `false` if the host app
+    ///     owns its audio session; see ``managesAudioSession``.
```
ianthetechie:
```suggestion
    ///   - isMuted: Whether the speech synthesizer is currently muted. (Normally this will be false,
    ///     unless you're providing your own "hot" synth.)
    ///   - managesAudioSession: Whether this observer should manage the shared `AVAudioSession`
    ///     while speaking.
    ///     Set to `false` if the host app will manage the audio session lifecycle and focus itself.
```

Cleaned up to give the framework user everything they need to know in a concise docstring. Removed the reference to a private member, as this probably won't even render in DooC anyways.

ianthetechie (commented):
Thanks for the PR! In the future, please make sure to read AGENTs.md for our LLM contribution guidelines. The meat of this PR is great, but we'd appreciate a bit more care. Claude etc. are _terrible_ at writing documentation.

PR #962, opened 2026-09-15
Cleanup & align route start / resume when no location available

freszu:
## What & why
this started with noticing that: UserLocationUnknown` error turned out to be entirely dead:
Android's exception class was never thrown after the fallback was introduced,
and all three platforms still documented ```@throws UserLocationUnknown```

The Android and React Native cores already fall back to the first coordinate of the route
when no location update has been observed yet,
so a session can begin before the first fix arrives.
iOS instead threw `FerrostarCoreError.userLocationUnknown`. - aligned to react & android behavior

## Changes

**iOS** — `startNavigation` and `resumeNavigation` now fall back to the route origin,
building the same synthetic location Android does
(accuracy `0`, no course or speed, current timestamp).

`resumeNavigation` resolves the location *after* `buildResumedSession()`,
since the cached route isn't known before then.
Nothing on `self` is mutated before the guard, so a throw still leaves the core untouched.

`FerrostarCoreError.userLocationUnknown` is replaced by `.emptyRouteGeometry`.
A route with no geometry offers no origin to fall back to,
so it's rejected rather than silently starting a session we can't track.

^^this one is behavior i am unsure off - android react might fail silently - but empty geometry is something user should verify beforehand so i consider provided behavior acceptable

Kotlin's `geometry.first()` throws on an empty list, so Android traps in that case;
Swift's `Optional` forces the case to be handled,
and `AGENTS.md` rules out force-unwrapping, hence the thrown error instead.

**Android** — deletes the unused `UserLocationUnknown` exception and the `@Throws` annotation.

**TS** - initially was also droping stale comment here - but this [was just fixed in main](https://github.com/stadiamaps/ferrostar/commit/c4d25ca40e5a60b82b3c199ae4757d9dd7476c73)

## Tests

Three new cases in `FerrostarCoreTests`:
the fallback when no fix has arrived,
a caller-supplied location taking precedence over it,
and the empty-geometry rejection leaving no partial state behind.


apple/Sources/FerrostarCore/FerrostarCore.swift
```diff
@@ -326,20 +328,19 @@ public protocol FerrostarCoreDelegate: AnyObject {
     /// triggering a recalculation.
     /// If this parameter is `nil`, the last location will be obtained from the configured location provider
     /// automatically.
-    /// If no location is available, this method will throw an exception.
+    /// If no location is available at all, the session starts from the first coordinate of the route
```
ianthetechie:
```suggestion
    /// If no location is available, the session starts from the first coordinate of the route
```

apple/Sources/FerrostarCore/FerrostarCore.swift
```diff
@@ -366,19 +367,20 @@ public protocol FerrostarCoreDelegate: AnyObject {
     /// **Important! This feature is experimental and may exhibit unexpected behavior. Please
     /// report any issues you encounter to help us improve it.**
     ///
-    /// - Parameter userLocation: The user's current location.
+    /// - Parameter userLocation: The user's current location. If this parameter is `nil`, the last location will be
+    /// obtained from the configured location provider automatically. If no location is available at all, the session
```
ianthetechie:
```suggestion
    /// obtained from the configured location provider automatically. If no location is available, the session
```

ianthetechie (commented):
Thanks! Just a few minor things to tighten up the documentation wording.

Issue #951, opened 2026-08-25
Recalculation on arrival/route completion can leak a trip and navigation session/cache


Archdoog:
The issue occurs when you get a recalculation right as you arrive. If you app doesn't immediately call stop navigation, the recalculation task can complete and re-inject a new trip continuing the existing session.

**What Happens**

1. You reach trip state complete/arrival.
2. You continue driving past before you end trip by calling `stopNavigation()`
3. Your UI probably exits the navigation view.
4. Ferrostar re-injects the trip on successful fetch of the recalculation.
5. Your app unexpectedly continues monitoring location.
6. You force kill your app to stop location.
7. Your app also unexpectedly still has a nav session to resume. Even though you arrived.

Note - this probably doesn't happen if your app calls stop immediately. However, my app (and probably others) allow users to loiter (e.g. to find parking) on the nav screen until they tap "Complete trip" which finally closes it out.

**Expected (Option 1)**

1. You reach trip state complete/arrival.
2. Ferrostar immediately blocks any behavior related to trip state flow. E.g. cancel recalculation and prevent further.
3. Ferrostar terminates the navigation session (no unexpected resume).
4. Ferrostar leaves location updates running (for loitering)
5. User taps "Complete Trip" in my app's case and then finally stops the location engine.

Archdoog:
Note, it's worth checking Android even though this was noted on iOS.

ianthetechie:
This was _probably_ fixed by #963 but leaving it open for further testing from @Archdoog

PR #963, opened 2026-09-15
Live Activity Life Cycle

hactar:
We have been getting user reports that sometimes live activities are not successfully removed when ending navigation. I have been unable to reproduce this reliably. But I did spot a few fishy things that I fixed:
- Queued navigation updates and stale reroute results could revive a session after it was stopped
- Apple recommends manually checking and clearing stale sessions on app launch to clear live activities that survive crashes or forced terminations

I also added some documentation for a new function one should call from the app and added some regression tests.

Hopefully this reduces the chances of stale live activities - and additionally allows users to clear a weird state by relaunching their app.


hactar:
@codex review

Archdoog:
This might be adjacent or caused by something I've noticed some of the time: https://github.com/stadiamaps/ferrostar/issues/951.

ianthetechie (approved):
Awesome; thanks for this, Patrick! We've known about several of these issues for a while but just never got around to cleaning them up 😅

PR #965, opened 2026-09-22
fix: android foreground service

### LINKED ISSUE #413: Android - ForegroundServiceManager not clearing notification on end (and maybe stop)
After starting a navigation session. The foreground notification will remain on your phone's notification list after the navigation session has ended/completed.

Archdoog:
Ensure the foreground service doesn't reappear for orphaned navigation sessions.

Closes #413


android/core/src/main/java/com/stadiamaps/ferrostar/core/service/ForegroundServiceManager.kt
```diff
@@ -43,7 +43,9 @@ class FerrostarForegroundServiceManager<T : ForegroundNotificationBuilder>(
   private val context: Context
     get() = weakContext.get() ?: throw IllegalStateException("Context is null")
 
-  private var isStarted = false
+  // True from startService() until stopService(). Set here, not in onServiceConnected, so a stop
+  // that arrives before the bind connects is honored.
```
ianthetechie:
```suggestion
  // True from startService() until stopService().
```

ianthetechie (approved):

